Issue #24, opened 2021-05-25
Make schemas of sources more explicit


cmungall:
Context: [Semantic ETL manifesto](https://docs.google.com/document/d/1Bgsyo-Z1oxEfxdNGB3KDM_NHfXYqdcXJIUd_j3iibi4/edit#)

We should make this repo an exemplar of a KG ETL repo. We should formally describe upstream sources in a lightweight way and make it very transparent how things map.

We have done most of the work already here:

https://github.com/Knowledge-Graph-Hub/kg-microbe/tree/master/schemas

This describes the core trait table that is ingested. We need to switch this to use linkml first (#21), but we may add more.

We can generate markdown for this using linkml, but we need to think how this integrates with https://knowledge-graph-hub.github.io/kg-microbe/index.html

I like how dipper includes a sources section: https://dipper.readthedocs.io/en/latest/sources.html

Would it make sense to

 1. generate markdown from all upstream schemas (just one so far)
 2. Deposit markdown here https://github.com/Knowledge-Graph-Hub/kg-microbe/tree/master/sphinx (can markdown be mixed with rst?)
 3. Publish

Alternatively we could use the generated python from the schema

cmungall:
see also https://docs.google.com/document/d/1iEsLp9pDvjGjgWMSLArtNf6Jwan-wjMl6_viQGyTWG8/edit?ts=60c3b505#

hrshdhgd:
Added documentation of the markdowns generated from linkml. The formatting we had decided on for the markdown files do not sit well with html. 

https://knowledge-graph-hub.github.io/kg-microbe/sources/carbon_substrates.html#other-properties

Thoughts, @cmungall ?

hrshdhgd:
![Screen Shot 2021-06-17 at 11 41 17 AM](https://user-images.githubusercontent.com/5247075/122439114-f807c180-cf60-11eb-9150-43eab60f9377.png)

cmungall:
> The formatting we had decided on for the markdown files do not sit well with html.

not sure what you mean.

it seems this is likely just a problem with linkml markdowngen. I think this ticket can be closed. 

note: change the name of the class from Example to something meaningful, e.g. FooTrait, where Foo is some discrimanotor such as the paper

PR #30, opened 2021-06-11
Sphinx sources addition

hrshdhgd:
Trying to get markdowns generated through linkml to show up in the sphinx documentation. Addressing #24


hrshdhgd:
rolling back to as it was before.

PR #84, opened 2023-11-17
Feba exploration

hrshdhgd:
- [ ] Explore feba and come up with a solution to transform into KGs
- [x] Get `medium-strains` edges and add them to `MediaDiveTransform`


realmarcin (approved):
Brook's commits already reviewed
Harshad's in progress work on more BacDive and MediaDive fields

PR #88, opened 2024-01-10
Uniprot

bsantan:


kg_microbe/download.py
```diff
@@ -1,5 +1,5 @@
 """Download resources from YAML file."""
-from kghub_downloader.download_utils import download_from_yaml  # type: ignore
+from kg_microbe.utils.download_utils import download_from_yaml
```
caufieldjh, line 2: If you're using a local version of the downloader than please open a PR with the changed version on the kghub-downloader repo so we can use the updated version in other KG-Hub repos!

kg_microbe/transform_utils/uniprot/uniprot.py
```diff
@@ -0,0 +1,178 @@
...
+        for i in tqdm(range(0, len(organism_ids), batch_size)):
+            values = self.get_uniprot_values_from_file(organism_ids, i, batch_size, fields, keywords, values,verbose)
+            
+            self.write_to_df(values, edge_writer, node_writer)
+            print('wrote to dataframe')
+
+    
+    def parse_binding_site(self,binding_site_entry):
+
+        chem_list=re.findall(r'/ligand_id="ChEBI:(.*?)";',binding_site_entry)
```
caufieldjh, line 111: do we always expect binding sites to be stored under `ligand_id` and will they always have ChEBI IDs (e.g., will they ever have IDs for other chemical knowledgebases)?

kg_microbe/transform_utils/uniprot/uniprot.py
caufieldjh, line 1: Consider letting [koza](https://github.com/monarch-initiative/koza) handle some of the transform details - it will transform from JSON so you don't need to worry about regexes, string replacement, etc

merge.yaml
caufieldjh, line 1: I'm being fairly pedantic here but it may help to name the `uniprot` transform something more specific since it isn't _everything_ in Uniprot, it's a subset (so maybe something like "uniprot_organism_enzyme" or something)

bsantan: Updated to "uniprot_genome_features"

kg_microbe/transform_utils/uniprot/uniprot.py
bsantan: Uniprot only includes CHEBI id's for binding sites, so none will be excluded unless they are labelled "substrate" thus without any specific chemical information/mappings.

kg_microbe/download.py
bsantan: Updated download_utils so it wouldn't affect existing code in kghub-downloader, and I will create a PR for that soon

kg_microbe/transform_utils/uniprot/uniprot.py
bsantan: I'll look into this! Thanks for the suggestion!

kg_microbe/download.py
hrshdhgd: This needs to be rolled back to 
```
from kghub_downloader.download_utils import download_from_yaml
```
This is a standard downloader we use for all KGs. @bsantan , I'd recommend removing local code from `kg_microbe.utils.download_util` and make a PR in kghub_downloader so others in the community could benefit from the awesome code!

hrshdhgd: I have started a new PR : https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/106

Please feel free to take over this PR

PR #91, opened 2024-01-12
Paths&blank files fix

bsantan:
Will output data/raw/uniprot_genome_features/empty_organism_queries.tsv to keep track of uniprot queries that return empty values


realmarcin (commented):
OK, looks good!
I will do one more fix after to lower case the 'Uniprot' path per convention.

realmarcin (approved):
merging

PR #97, opened 2024-01-16
Naming fix

bsantan:
Fixed local_name of uniprot items and fixed issue of not taking into account past organism IDs marked as empty from uniprot query in empty_organism_queries.tsv.


bsantan:
obsolete

PR #100, opened 2024-01-23
Update download_utils.py

bsantan:
write to empty_organism_queries.tsv iteratively rather than after run


bsantan:
Need to space out read/write operation

PR #102, opened 2024-01-29
Adding ingests for Metanetx, Rhea, MONDO, GO-plus instead of GO, HP.

bsantan:
Introducing ingests of Metanetx and Rhea, still need to correctly map chemicals/reactions to correct resource.


bsantan:
obsolete, using Rhea ingest instead

PR #108, opened 2024-02-07
Uniprot sourcename fix

bsantan:
- Fix in merge.yaml to look in uniprot_genome_features for transformed files
- Fix in uniprot.py to us Uniprot prefix in protein IDs rather than uniprot_genome_features prefix


hrshdhgd:
Instead of harcoding strings I'd recommend using [constants.py ](https://github.com/Knowledge-Graph-Hub/kg-microbe/blob/master/kg_microbe/transform_utils/constants.py)to declare as a variable first and then use it in the project

bsantan:
I added constants to the uniprot.py file rather than strings.

hrshdhgd (commented):
Looks good! Let's push to merge #107 first and then we can merge this one. I have a suspicion there will be conflicts with this branch that'll need manual intervention. But we'll deal with it when we get there. Undo the changes you made in `tox.ini`. #107 seems to have been fixed. So once we merge down the main branch to this one, all should be fine!

PR #129, opened 2024-03-04
uniprot transform changes

bsantan:


kg_microbe/transform_utils/uniprot/uniprot.py
```diff
@@ -96,15 +103,16 @@ def get_uniprot_values_from_file(self, input_dir, nodes, source, node_writer, ed
         """
         with tqdm(total=len(nodes) + 1, desc="Processing files") as progress:
```
hrshdhgd, line 104: use `DummyTqdm` like I have in bacdive 
```
progress_class = tqdm if show_status else DummyTqdm
            with progress_class(
                total=len(nodes) + 1, desc="Processing files"
            ) as progress:
                for i in range(len(nodes):
```

kg_microbe/transform_utils/uniprot/uniprot.py
```diff
@@ -169,25 +198,12 @@ def write_to_df(self, uniprot_values, edge_writer, node_writer):
             if "Protein names" in entry:
```
hrshdhgd, line 198: `constants.py` since you're using "Protein names" multiple times

bsantan:
replaced by uniprot2s3_transform_new

PR #160, opened 2024-04-05
unipathways post processing transform

bsantan:


kg_microbe/transform_utils/ontology/ontology_transform.py
```diff
@@ -35,6 +51,7 @@
     "rhea": "rhea.json.gz",
     "ec": "ec.json",
     "uniprot": "uniprot.json.gz",
+    "unipathways": "upa.json"
```
hrshdhgd, line 54: switch this to "upa": "ups.owl"

hrshdhgd (approved):

PR #164, opened 2024-04-15
fix issue of getting "EC: " nodes

bsantan:


realmarcin (approved):
Reviewed and accepted

PR #165, opened 2024-04-24
Updates to Rhea and EC ingest

bsantan:
Fixed issues with Rhea, RheaMapper, and EC ingest:
1. Rhea ontology and Rhea pyobo ingest will not include direction relationships because they are already included in inverse triples (i.e. directional reaction subclassOf bidirectional reaction)
2. Rhea pyobo ingest will not include CHEBI, GO, Uniprot, or EC relationships because they are already covered by EC, Uniprot ingests. The code is there if we choose to include these
3. EC ingest will not include Uniprot relationships  because they are already covered byUniprot ingest. The code is there if we choose to include these

### DIFF
--- a/download.yaml
+++ b/download.yaml
@@ -139,12 +139,12 @@
   url: https://github.com/geneontology/unipathway/raw/master/upa.owl
   local_name: upa.owl
 
-# 
-# RHEA
-# 
--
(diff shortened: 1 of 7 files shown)


realmarcin (approved):
reviewed and accepted

PR #170, opened 2024-05-10
UniRef90 transform

hrshdhgd:


realmarcin:
This transform works on NERSC and produces correct edges and nodes. However, we have a critical data gap here in that we did not find a way to obtain the taxon UniProt proteins membership to the UniRef90 clusters. Thus we can merge UniRef90 transform but it will be unlinked from taxa and proteins at the moment.

@hrshdhgd

hrshdhgd:
Examples please!

PR #179, opened 2024-06-19
update all carbon substrate edges to use "consumes"

bsantan:
Confirmed in traits source that all relationships between microbe and carbon substrate is that microbe utilizes/consumes chemical


realmarcin (commented):
looks great!

realmarcin (approved):
approved

PR #209, opened 2024-08-09
rename transform input and output folders

bsantan:


kg_microbe/transform_utils/ontologies/ontologies_transform.py
```diff
@@ -102,8 +103,8 @@ def run(
             self.parse(k, data_file, k)
         else:
             # load all ontologies
-            for k in ONTOLOGIES.keys():
-                data_file = self.input_base_dir / ONTOLOGIES[k]
+            for k in ONTOLOGIES_LIST.keys():
```
hrshdhgd, line 106: List don't have keys.. could refactor this to be `ONTOLOGIES_MAP`?

PR #210, opened 2024-08-09
first slurm scripts

hrshdhgd:


realmarcin (approved):
working as of 20240702

PR #266, opened 2024-11-19
fix iteration of rhea hierarchy to get correct reaction strings

bsantan:


realmarcin (approved):
The new RHEA transforms have correct labels.

PR #278, opened 2025-02-03
updated direction for bacdive and madin_etal

bsantan:


bsantan:
Also remove edges between medium and metabolite (of the form medium -> assesses -> CHEBI)

kg_microbe/transform_utils/bacdive/bacdive.py
```diff
@@ -368,7 +361,7 @@ def _process_medium(self, dictionary, ncbitaxon_id, key, edge_writer):
             edge_writer.writerow(
                 [
                     ncbitaxon_id,
-                    NCBI_TO_ASSAY_EDGE,
```
realmarcin, line 371: @bsantan I think this was correct in the end

kg_microbe/transform_utils/bacdive/bacdive.py
```diff
@@ -348,14 +348,7 @@ def _process_metabolites(self, dictionary, ncbitaxon_id, key, node_writer, edge_
                                 metabolite_id,
                                 None,
                                 BACDIVE_PREFIX + key,
-                            ],
-                            [
-                                medium_id,
-                                MEDIUM_TO_METABOLITE_EDGE,
```
realmarcin, line 354: this was the edge to the antibitioc -- but that medium should have this edge because this would be medium:X+ antibitioc, really there should be an assay node here but can fix in the future.

realmarcin (approved):

PR #338, opened 2025-08-21
Replace `oxygen` mapping section from `custom_curies.yaml` with alternate logic

sujaypatil96:
This is a first PR in a larger effort to eliminate the use of the `kg_microbe/transform_utils/custom_curies.yaml` file in this repo. We are starting with just the _oxygen_ mappings that are represented in the custom_curies.yaml file, and writing alternate logic in the bacdive transformation script to capture the same mappings for the oxygen preference mappings, except, here we are using the one single authoritative SOT which is the mappings file from METPO: https://github.com/berkeleybop/metpo/blob/main/generated/bacdive_oxygen_phenotype_mappings.tsv

In the edges.tsv/nodes.tsv output after this modification, we don't see any nodes/edges using terms in the "oxygen:" namespace.


realmarcin (approved):
Good step forward, some more next pieces in comments.

PR #353, opened 2025-08-26
minimal merge config and sanitizer

turbomam:


turbomam:
move code to utils directory

kg_microbe/sanitize_curies.py
```diff
@@ -0,0 +1,121 @@
+"""Sanitize CURIEs and URIs in KGX TSV files."""
```
realmarcin, line 1: @turbomam I am curious at which step this sanitization would be applied?
if nodes are sanitized then also need to sanitize edges -- so is this after the merge step?

PR #357, opened 2025-08-26
Add BactoTraits to MongoDB JSON converter

turbomam:
- Converts BactoTraits CSV to MongoDB-compatible JSON format
- Handles 3-row hierarchical header structure properly
- Creates nested field paths (e.g., taxonomy.genus, gc_content.lte_42dot65)
- Preserves one-hot encoding with only non-zero values
- Splits comma-separated culture collection codes into arrays
- Uses proper field names like motility.non_motile instead of non.motile
- Handles encoding issues and forward-fills category headers
- Converts 19,455 strain records with ~16 fields avg (vs 105 sparse fields)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/kg_microbe/bactotraits_to_mongo.py
+++ b/kg_microbe/bactotraits_to_mongo.py
@@ -1,13 +1,20 @@
 #!/usr/bin/env python3
+"""
+BactoTraits CSV to MongoDB JSON converter.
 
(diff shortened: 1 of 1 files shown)


realmarcin (approved):
thanks for adding this

PR #362, opened 2025-08-29
Start using ROBOT template as source of METPO mappings

### LINKED ISSUE #361: Update transformation pipelines to start using METPO ROBOT template as source of mappings
We made a decision to stop using the SPARQL-processed mapping file (derived from METPO ROBOT template) here: https://github.com/berkeleybop/metpo/blob/main/generated/bacdive_oxygen_phenotype_mappings.tsv and instead use the METPO ROBOT template here directly: https://github.com/berkeleybop/metpo/blob/main/src/templates/metpo_sheet.tsv

sujaypatil96:
Fixes #361


realmarcin (approved):
looks great, thanks for the bacdive variable update

PR #366, opened 2025-09-03
Convert METPO URIs to CURIEs

### LINKED ISSUE #364: convert URIs to CURIEs (METPO terms)
using biopragmatics

@sujaypatil96

sujaypatil96:
Fixes #364 

The edges.tsv/nodes.tsv for BacDive now uses METPO term CURIEs instead of full URIs.

### DIFF
--- a/kg_microbe/transform_utils/prefixmap.json
+++ b/kg_microbe/transform_utils/prefixmap.json
@@ -1,0 +1,3 @@
+{
+    "METPO": "https://w3id.org/metpo/"
+}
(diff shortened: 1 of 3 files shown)


kg_microbe/transform_utils/prefixmap.json
```diff
@@ -0,0 +1,3 @@
+{
```
realmarcin, line 1: will anything else need to be added here? I think any used ontology?
if so, that would include:
CHEBI, NCBITaxonomy, ENVO, GO, UBERON, MONDO, HPO

realmarcin (approved):
looks good for mapping infrastructure

PR #389, opened 2025-09-24
Mappings

realmarcin:


kg_microbe/transform.py
```diff
@@ -49,11 +37,11 @@
...
-    CTD: CTDTransform,
-    DISBIOME: DisbiomeTransform,
-    WALLEN_ETAL: WallenEtAlTransform,
-    UNIPROT_FUNCTIONAL_MICROBES: UniprotFunctionalMicrobesTransform,
+    #UNIPROT_HUMAN: UniprotHumanTransform,
+    #CTD: CTDTransform,
+    #DISBIOME: DisbiomeTransform,
+    #WALLEN_ETAL: WallenEtAlTransform,
+    #UNIPROT_FUNCTIONAL_MICROBES: UniprotFunctionalMicrobesTransform,
 }
```
sujaypatil96: Do we want to remove all of these data sources from the "data source transform index" (`transform.py`)?

PR #394, opened 2025-10-03
Update BacDive ingest pipeline to use METPO class/synonym mapping from sheets

sujaypatil96:
- [ ] Update METPO classes/properties sheet handler to be work for BacDive ingest pipeline
- [ ] Update BacDive ingest pipeline to use updated methods from sheet handler helpers file

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -180,7 +180,7 @@ from kg_microbe.transform_utils.constants import (
 )
 from kg_microbe.transform_utils.transform import Transform
 from kg_microbe.utils.dummy_tqdm import DummyTqdm
-from kg_microbe.utils.mapping_file_utils import load_metpo_mappings
+from kg_microbe.utils.mapping_file_utils import load_metpo_mappings, _build_metpo_tree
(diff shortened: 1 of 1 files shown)


sujaypatil96:
QC/analysis of KGX (edges.tsv) file produced after running the modified code in this PR on the BacDive data source dump:

Consider for example a strain from the BacDive dataset, like the one (`"BacDive-ID": 100`) attached in this JSON file: 
[bacdive_id_100.json](https://github.com/user-attachments/files/22960405/bacdive_id_100.json)

Looking at the METPO edges created for this strain, we can see:

| subject           | predicate               | object       | relation   | primary_knowledge_source |
|-------------------|--------------------------|--------------|-------------|---------------------------|
| strain:bacdive_100 | biolink:has_phenotype   | METPO:1000603 | RO:0002215 | bacdive:100              |
| strain:bacdive_100 | biolink:has_phenotype   | METPO:1000615 | RO:0002215 | bacdive:100              |

it has edges that relate to METPO:1000603 (`anaerobic`) and METPO:1000615 (`mesophilic`) which we can see is true from the JSON.

CC: @realmarcin

realmarcin (approved):
A lot of new code -- and it works.

PR #401, opened 2025-10-15
Make sure the category values are CURIEs in the Madin nodes KGX file

sujaypatil96:
Previously a node in the transformation output KGX file for the Madin et al data source had rows like this:

| id           | category               | name    |
|---------------|------------------------|---------|
| METPO:1000602 | https://biolink.github.io/biolink-model/PhenotypicQuality | aerobic |

After this PR, it will look like this:

| id           | category               | name    |
|---------------|------------------------|---------|
| METPO:1000602 | biolink:PhenotypicQuality | aerobic |

It is a follow up activity from https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/399


realmarcin (changes requested):
I ran the madin_etal transform on this branch and don't see any https in the nodes files now.

however, I did notice that the nodes.tsv file has many empty lines now? Almost every other line is empty, the edges.tsv is fine.

PR #412, opened 2025-10-21
Unroll strains

realmarcin:


kg_microbe/transform_utils/prefixmap.json
```diff
@@ -1,4 +1,5 @@
 {
     "METPO": "https://w3id.org/metpo/",
-    "biolink": "https://biolink.github.io/biolink-model/"
+    "biolink": "https://biolink.github.io/biolink-model/",
+    "kgmicrobe.strain": "https://example.org/kgmicrobe/strain/"
```
sujaypatil96: Prefix/namespace defined in a "prefix map" as described by curies Python library here: https://curies.readthedocs.io/en/latest/struct.html#prefix-maps

kg_microbe/transform_utils/bacdive/bacdive.py
```diff
@@ -1042,18 +1048,89 @@ def run(self, data_file: Union[Optional[Path], Optional[str]] = None, show_statu
...
                             ncbitaxon_id = NCBITAXON_PREFIX + str(
                                 general_info[NCBITAXON_ID][NCBITAXON_ID]
                             )
-                        species_with_strains = [ncbitaxon_id]
                         ncbi_description = general_info.get(GENERAL_DESCRIPTION, "")
                         ncbi_label = get_label(self.ncbi_impl, ncbitaxon_id)
                         if ncbi_label is None:
                             ncbi_label = ncbi_description
 
+                    # If no NCBITaxon ID from BacDive API, try searching by name
```
sujaypatil96: Minor: not BacDive API –> BacDive JSON (may be constructed using API not sure but we're not consuming API directly)

kg_microbe/transform_utils/bacdive/bacdive.py
```diff
@@ -1042,18 +1048,85 @@ def run(self, data_file: Union[Optional[Path], Optional[str]] = None, show_statu
...
+                                SUBCLASS_PREDICATE,
+                                ncbitaxon_id,
+                                RDFS_SUBCLASS_OF,
+                                BACDIVE_PREFIX + key,
+                            ]
+                        )
+
+                    # Always use only the strain node for feature edges
+                    # NCBITaxon relationship is preserved via subclass edge (strain -> NCBITaxon)
+                    feature_targets = [organism_id]
```
sujaypatil96: If this is just going to be organism_id why do we need to convert it to a list? We can simply use it as is and revert the change in logic in methods like `_process_phenotype_by_metpo_parent()`, `_process_antibiotic_resistance()`, `_process_metabolites()`, etc.

kg_microbe/transform_utils/bacdive/bacdive.py
```diff
@@ -1221,179 +1294,41 @@ def run(self, data_file: Union[Optional[Path], Optional[str]] = None, show_statu
...
-                                for organism in species_with_strains
-                            ]
-
-                            edge_writer.writerows(edges_data_to_write)
+                            # Create edge(s) from organism(s) to medium
+                            for target in feature_targets:
+                                edge_writer.writerow(
+                                    [
+                                        target,
+                                        NCBI_TO_MEDIUM_EDGE,
```
sujaypatil96: I think @turbomam diagreed with the use of this predicate (`biolink:occurs_in`) but we can use it for this release and make an issue to find a better predicate for the next release.

PR #426, opened 2025-11-12
Add code snippet to BacDive ingest pipeline to pick up on halophily preference classes correctly

sujaypatil96:
This commit in this PR actually should have been part of https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/424 where we removed the salinity terms mapping section from the custom curies YAML file. The pattern to follow is removing a section from the custom curies YAML and then also adding a snippet of code to the bacdive ingest pipeline so it knows for what "parent" (halophily preference) it needs to pick up the mappings.

Without this snippet of code, it's really just looking at the "keywords" section to make edges.


realmarcin (approved):
the transform works, halophily 182 edges from bacdive

PR #490, opened 2026-03-04
added metatraits

crocodile27:


turbomam (commented):
Good progress getting MetaTraits into the pipeline @crocodile27. The JSONL ingestion, NCBITaxon resolution with caching, and KGX dedup are all working. Here are the things I think need attention before merge.

## 1. `_get_relation_for_predicate()` collapses most trait types to `has_phenotype`

This is the big one. MetaTraits has at least 8 structurally distinct assertion types, but the current resolution is binary:

```python
def _get_relation_for_predicate(self, predicate: str) -> str:
    if predicate == CAPABLE_OF_PREDICATE:
        return BIOLOGICAL_PROCESS
    return HAS_PHENOTYPE
```

That means "produces: ethanol" emits `has_phenotype ethanol` instead of `produces ethanol`. "carbon source: acetate" gets `has_phenotype` when it should express a metabolic capability involving acetate (CHEBI:30089). Here is what the different patterns actually need:

| Trait pattern | Example | Predicate | Object ontology |
|---|---|---|---|
| Substrate assimilation | `assimilation: D-glucose` | `biolink:capable_of` | GO |
| Acid production | `builds acid from: D-glucose` | `biolink:capable_of` | GO |
| Carbon/nitrogen source | `carbon source: acetate` | `biolink:capable_of` | GO |
| Enzyme activity | `enzyme activity: catalase (EC1.11.1.6)` | `biolink:capable_of` | EC, GO |
| Produces compound | `produces: ethanol` | `biolink:produces` | CHEBI |
| Phenotype | `gram positive`, `sporulation` | `biolink:has_phenotype` | METPO |
| Quantitative trait | `pH maximum`, `temperature growth` | `biolink:has_phenotype` | METPO |
| Electron donor/acceptor | `electron donor: thiosulfate` | `biolink:capable_of` | GO, CHEBI |

`_build_trait_mapping` already stores per-trait predicates from the METPO sheet, so the data is partially there. The issue is that `_get_relation_for_predicate` throws most of it away. Consider making it a lookup table keyed on predicate, and for `produces:` patterns, parse the object as a chemical and use `biolink:produces` with CHEBI. For enzyme activities with EC numbers in the trait name, extract the EC number and use it as the object CURIE directly.

I have been working on curated mapping TSVs in [`turbomam/microbial-trait-mappings`](https://github.com/turbomam/microbial-trait-mappings) organized by entity type (chemicals, enzymes, pathways, phenotypes). There is a new PR with seed data and round-trip tests that verify correct edge resolution for each pattern: [turbomam/microbial-trait-mappings#2](https://github.com/turbomam/microbial-trait-mappings/pull/2). That repo could serve as the authoritative lookup for which pattern each trait falls into.

## 2. Consider logging which conversion pattern was applied

`unmapped_traits.tsv` is useful. For mapped traits, it would help debugging and future coverage analysis to know which conversion pattern was applied. Even a `pattern_type` column in the edges output or a separate summary file would be enough.

## 3. Minor items

- `TOX_GUIDE.md` (218 lines) in the repo root is a lot of troubleshooting content for that location. Consider putting it in `docs/`

PR #527, opened 2026-03-19
Parallelize MediaDive bulk download (~20x faster)

crocodile27:
## Summary
- Replaces the two sequential per-medium API loops in `mediadive_bulk_download.py` with a `ThreadPoolExecutor(max_workers=DEFAULT_MAX_WORKERS)`
- Adds two private helper functions (`_fetch_medium_detail`, `_fetch_medium_strains`) for the executor to dispatch

## Addresses review comments from Mark and Marcin
- **Mark**: Lower `DEFAULT_MAX_WORKERS` from 20 → 5 (polite for small academic API at DSMZ); add descriptive `USER_AGENT`; respect `Retry-After` header on 429; switch `futures`-dict to `executor.map`; tighten return type annotations to `tuple[str, dict]`
- **Marcin**: Make `max_workers`, `retry_count`, `retry_delay` parameters on all public download functions; pass them through `_fetch_*` helpers into `get_json_from_api` (previously helpers always used defaults); add tests covering defaults, retry parameter propagation, and concurrency bounds

## Performance
Measured on 50 real MediaDive API calls (no cache), extrapolated to full dataset:

| Phase | Before | After (measured) |
|---|---|---|
| Detailed recipes (3,333 requests) | ~33 min | ~2 min |
| Strain associations (3,333 requests) | ~33 min | ~2 min |
| **Total MediaDive bulk** | **~66 min** | **~4 min** |

Measured speedup: **16.2x** (599 ms/req → 37 ms/req, 20 workers, live API). Default is now 5 workers; still a large speedup, and callers can increase for CI/trusted environments.

The `requests_cache` SQLite backend has built-in locking so concurrent writes are thread-safe.

## Test results
- [x] 136 passed, 25 skipped, 5 warnings in 15.27s (`poetry run pytest tests/ -q`)
- [x] 8 new tests in `tests/test_mediadive_bulk_download.py`: default values, Retry-After behaviour, retry parameter propagation, concurrency bounds
- [ ] `poetry run kg download` completes MediaDive bulk download in ~4 min instead of ~66 min

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,6 +168,8 @@ data/transformed/uniprot_functional_microbes/uniprot_kgx.zip
 
 data/transformed/ontologies/.Rapp.history
 mediadive_cache.sqlite
+mediadive_bulk_cache.sqlite
+data/raw/mediadive/
 data/raw/.keep
 kg_microbe/transform_utils/uniprot_human/tmp/relevant_files.tsv
 kg_microbe/transform_utils/uniprot/tmp/relevant_files.tsv


kg_microbe/utils/mediadive_bulk_download.py
```diff
@@ -8,13 +8,16 @@
...
 import time
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
 from typing import Dict, List
 
 import requests
 import requests_cache
 from tqdm import tqdm
 
+MAX_WORKERS = 20
```
turbomam: parameterize

PR #550, opened 2026-04-17
Fix pre-existing ruff lint errors in tests/test_mediadive_bulk_download.py (closes #549)

turbomam:
## What this does

Runs `ruff check --fix` on `tests/test_mediadive_bulk_download.py`, resolving the 7 auto-fixable errors that have been present on `master`:

- **F401** × 3: unused imports (`pytest`, `_fetch_medium_detail`, `_fetch_medium_strains`)
- **D203** × 4: missing blank line before class docstring (`TestDefaults`, `TestRetryAfter`, `TestRetryParameters`, `TestRateLimiter`)

No semantic change. Test behavior is identical.

Closes #549.

## Strengths

- **Unblocks every PR's CI.** Until this lands, the `Linting` step fails on every PR (see e.g. #548's `build (3.12)` run), even PRs that don't touch Python. Merging this restores a green baseline for everyone.
- **Auto-generated, reviewable.** Every change is what ruff itself recommended. Diff is 7 lines.
- **No new rules enabled.** This doesn't add strictness; it applies fixes ruff was already asking for against the ruleset currently in `pyproject.toml`.

## Weaknesses / things worth flagging

- **Signals that the originating PR didn't enforce its own lint gate.** `test_mediadive_bulk_download.py` was recently added to master with these violations already present. Worth understanding how — did the lint step skip new files? Did the contributor override? Was this merged from a branch where lint passed because of different settings? Not a reason to block this PR, but a governance question for the team.
- **Trivial rebase nuisance** for anyone currently editing this file.
- **Doesn't address the broader problem** that `master` can land PRs with failing CI. That's a branch-protection question tracked in #546.

## Verification

```
$ cd kg-microbe
$ poetry run ruff check tests/test_mediadive_bulk_download.py
All checks passed!
```

Issue #549, opened 2026-04-17
master fails `ruff check` on 3.12 CI — 7 lint errors in tests/test_mediadive_bulk_download.py


turbomam:
## Symptom

Every PR against `master` currently fails CI at the `Linting` step because `master` has **7 pre-existing ruff errors** in `tests/test_mediadive_bulk_download.py`. Example failure on PR #548 (an unrelated dependabot config PR): https://github.com/Knowledge-Graph-Hub/kg-microbe/actions/runs/24576474619

## Errors

```
F401 [*]  pytest imported but unused                                           (line 7)
F401 [*]  _fetch_medium_detail imported but unused                             (line 13)
F401 [*]  _fetch_medium_strains imported but unused                            (line 14)
D203 [*]  1 blank line required before class docstring (TestDefaults)          (line 22)
D203 [*]  1 blank line required before class docstring (TestRetryAfter)        (line 50)
D203 [*]  1 blank line required before class docstring (TestRetryParameters)   (line 83)
D203 [*]  1 blank line required before class docstring (TestRateLimiter)       (line 115)
```

All 7 are marked `[*] fixable with the --fix option`.

## Fix

```
ruff check --fix tests/test_mediadive_bulk_download.py
```

Single commit to master (or one tiny PR) unblocks CI for every open and future PR. Until this is fixed, no PR can merge via the green-CI path; each one needs admin override or a merge despite red CI.

## Why the file didn't break earlier

Likely because the file was landed recently (Apr 2026) and the specific ruff rules (D203 especially) may have been enabled after the file was written, or the file was merged via a code path that bypassed the lint gate.

realmarcin:
Verified: `poetry run ruff check tests/test_mediadive_bulk_download.py` reports 'All checks passed!' on current `master`. The 7 lint errors are fixed. Closing as done.

PR #548, opened 2026-04-17
Add .github/dependabot.yml to open weekly dependency PRs

turbomam:
## What this does

Configures Dependabot to open weekly PRs for Python and GitHub Actions dependencies.

- **Minor / patch** updates → **one grouped PR per week per ecosystem** (not N separate PRs)
- **Major** updates → individual PRs (Dependabot default; needed for per-PR review)
- Labels: `dependencies` (+ `github-actions` for Actions)

## What it explicitly does NOT do

- Does not merge anything. Every Dependabot PR still requires a human merge click.
- Does not change branch protection, auto-merge, or any other repo setting.
- Does not touch `poetry.lock` directly — Dependabot will resolve and propose updates through its own PRs.

## Strengths

- **Capability-only; no compulsion.** The config only causes Dependabot to *propose* PRs. Nothing merges automatically.
- **Grouped minor/patch reduces review burden.** One weekly PR per ecosystem for small bumps, vs a PR per package (which can be 10+ with 33 open alerts).
- **Addresses latent vulnerability exposure.** The repo currently has 33 open Dependabot alerts (1 critical, 10 high) with no configured mechanism to propose fixes.
- **Easily reversible.** Deleting the file turns Dependabot back off.

## Weaknesses / things worth flagging

- **Adds recurring review burden.** Even grouped, weekly Dependabot PRs need someone to look at them. If the team is already stretched reviewing functional PRs like #531, this is extra load.
- **Grouping / schedule is a judgment call.** Weekly Monday + group-minor-patch reflects my preferences; the team may reasonably prefer a different cadence, per-ecosystem splits, or a tool other than Dependabot (e.g. Renovate has more flexible grouping).
- **One-person-driven config without team input.** Opened this as a proposal during a repo cleanup pass; haven't discussed it with the team. Should be reviewed for fit rather than approved on technical correctness alone.
- **Blocked today by the pre-existing ruff lint error on `master`.** CI fails at the `Linting` step on every PR until #549 / #550 lands. This PR's own CI can't go green until that's fixed.
- **Does not fix the existing 33 alerts directly.** Dependabot will propose PRs over the coming weeks; immediate remediation would need a manual `poetry update` pass (see draft PR forthcoming, if that's the route the team prefers).

## Admin-required follow-ups (partially addresses #546)

These need admin permission (my role is `maintain`). Worth doing once this PR lands to get the most value:

[...]

Issue #546, opened 2026-04-17
Security & governance baseline: 30 Dependabot alerts, no branch protection, missing standard configs


turbomam:
## Summary

The repo is missing the standard GitHub security/governance baseline and carries **30 open Dependabot alerts**: **1 critical, 9 high, 16 medium, 4 low**. See the Security tab for details.

Each gap below is small; together they represent a coherent hygiene cluster worth addressing.

## Baseline gaps

| Missing | Purpose | Effort |
|---|---|---|
| Branch protection on `master` | Require passing checks; block force-push and deletion | Settings toggle |
| `.github/dependabot.yml` | Auto-propose dependency upgrades | Single-file PR |
| `.github/workflows/codeql.yml` | Free static security analysis | Single-file PR (GitHub template) |
| `SECURITY.md` | Vulnerability reporting policy | Single-file PR |
| `CODEOWNERS` | Auto-request reviews from owners of touched paths | Requires owner decisions |
| `.pre-commit-config.yaml` | Local lint/format gate | Single-file PR |

## Suggested order

1. Branch protection on `master` (stops the bleed first)
2. `dependabot.yml` + `codeql.yml` (cheap, automated)
3. `SECURITY.md`
4. `CODEOWNERS` and `.pre-commit-config.yaml` (need team input)

<details>
<summary>Why branch protection first</summary>

Right now `master` has no required checks, no force-push block, and no deletion block (`gh api repos/Knowledge-Graph-Hub/kg-microbe/branches/master/protection` → 404). Every subsequent hygiene PR relies on the base branch being stable.
</details>

turbomam:
## Full audit of GitHub security / code-quality features on this repo (2026-04-17)

Captured during repo-cleanup pass. All of the below are free on public repos (no GHAS license needed). Tiering: **R** = repo-settings toggle (admin only); **F** = file-based (regular PR).

### Security features that are OFF / missing

| Feature | Type | Status | Note |
|---|---|---|---|
| Dependabot alerts (`vulnerability_alerts`) | R | **OFF** (HTTP 404) | Sidebar shows "33 vulnerabilities" but main view says "alerts are inactive" — the count is a stale artifact in GitHub's DB. No new CVE will be detected until this is toggled on. |
| Dependabot security updates (`automated_security_fixes`) | R | **OFF** (HTTP 404) | Even when alerts are on, this is what actually opens fix-PRs. |
[...]

PR #551, opened 2026-04-17
DRAFT: Manual `poetry update` to close Dependabot alerts — for team discussion

turbomam:
## Status

**DRAFT. Not for merge today.** Opening as an artifact for team discussion — to let the team see exactly what a manual alert-closure pass looks like before deciding whether to pursue it.

## Why this exists

kg-microbe has 33 open Dependabot alerts (1 critical, 10 high, 18 moderate, 4 low). There is no `.github/dependabot.yml` (see #548), so Dependabot only reports alerts and doesn't propose PRs. This PR is one way to close most of the alerts without waiting for the dependabot.yml PR to land.

## What changed

Ran `poetry update` on `master`. Affects `poetry.lock` only (no `pyproject.toml` changes). ~40 package updates, several major-version bumps included.

## Alerts this likely closes

| Package | Old | New | Severity |
|---|---|---|---|
| aiohttp | 3.12.13 | 3.13.5 | high |
| tornado | 6.5.1 | 6.5.5 | high (×2) |
| protobuf | 6.31.1 | 7.34.1 | high |
| setuptools | 80.9.0 | 82.0.1 | high (×2) |
| virtualenv | 20.31.2 | 21.2.4 | high |
| jupyterlab | 4.4.4 | 4.5.6 | high |
| notebook | 7.4.4 | 7.5.5 | high |
| cryptography | 45.0.5 | 46.0.7 | medium |
| pytest | 8.4.1 | 9.0.3 | medium |

## Alert this does NOT close

| Package | Old | New | Severity | Why |
|---|---|---|---|---|
| h11 | 0.16.0 | 0.16.0 | **critical** | 0.16.0 is the latest; no patch released. Requires upstream fix or a workaround. |

## Strengths

- **Closes the majority of open alerts in one shot.** No need to wait for dependabot.yml → weekly Dependabot PRs → per-PR reviews accumulated over weeks.
- **No `pyproject.toml` changes.** Stayed within existing version constraints; no forced API-contract changes to our own package metadata.
- **Self-contained.** Lock-file-only diff; easy to revert by reverting the commit.
- **Gets `requests`, `idna`, `urllib3` current** (minor bumps), reducing the crust of unfixed advisories.

## Weaknesses / concerns worth discussing

1. **Major version bumps landed with no manual review per package.** `neo4j 5.x → 6.x`, `koza 0.7.x → 2.1.x`, `bacdive 1.x → 2.x`, `ruff 0.12 → 0.15`, `pytest 8 → 9`, `protobuf 6 → 7`. Any of these could have breaking API changes that pass unit tests but break actual KG-build outputs.

2. **KG-microbe has limited integration-test coverage for pipeline output.** A bump that changes (e.g.) rdflib serialization, oaklib output format, or KGX transform behavior could silently change the generated KG. CI can pass while the real output changes in a way only caught by manual rebuild-and-compare.

3. **Does not close the critical `h11` alert.** That's the single highest-priority alert and needs a different approach (upstream wait, workaround, or revisit when h11 0.17+ ships).

4. **Conflicts with in-flight PR #531** (marcin's 88-commit MetaTraits PR). A concurrent lock-file change will cause merge conflicts. This PR should probably land *after* #531 or be rebased onto #531 once that lands.

5. **Bypasses a deliberate pin**: `watchdog = "<4.0.0" # ! This is a temporary fix 4.0.0 breaks on GH Actions`. This PR preserves that, but shows there are undocumented version assumptions the team has discovered painfully. There may be others I haven't noticed.

6. **This PR's own CI is blocked by the pre-existing lint regression (#549 / #550).** Even if approved, CI can't turn green until the lint-fix PR lands.

7. **One-person-driven.** Reflects the reviewer's judgment about what to bump, not the team's. Many teams prefer per-package, reviewable, cadenced bumps (i.e., what Dependabot does) over a big batch.

## Alternatives to this approach

- **A. Wait for #548 (dependabot.yml) to land**, then let Dependabot propose a weekly grouped bump PR. Slower; orderly; lets the team review incrementally.
- **B. Surgical single-package bumps.** Open one PR for each high-priority alert. Lots of PRs; easy to review each.
- **C. Do nothing until after #531 lands.** Avoids conflict with in-flight work; alerts continue to age.
- **D. This PR, minus the major bumps.** Would require `poetry update --only-root-deps` or similar scoping. Closes fewer alerts but lower risk.

## Recommendation

Don't merge this as-is. Use it as a concrete reference while the team decides which of A–D (or this) fits best.


realmarcin:
Closing: the premise does not hold.

This PR exists to "close most of the alerts" by bumping dependencies. But all 33 open Dependabot alerts are **already fixed in `poetry.lock` on master** — every one names a package whose first patched version is at or below what the lock pins, including the critical h11 (locked 0.16.0, patched 0.16.0). Full evidence in #546.

So a dependency bump closes nothing. I ran `poetry update` on current master to confirm: it produces ~23 routine bumps (`ruff 0.15.14 -> 0.16.3`, `mypy 2.1.0 -> 2.3.1`, `biopython 1.87 -> 1.88`, ...) with no security effect, and reverted it rather than ship an unjustified lock churn that risks the lint and type gates.

A routine lock refresh may still be worth doing on its own merits — but it should be argued as maintenance, not as alert closure, and this branch's April lock is far too stale to be the vehicle (it is already CONFLICTING).

The real remaining work is in #546: clear the stale alerts and enable the security toggles that are still `disabled`.

PR #552, opened 2026-04-17
Add SECURITY.md vulnerability-reporting policy

turbomam:
## Status

**DRAFT for team discussion.** Part of the security/quality baseline (#546).

## What this adds

A minimal `SECURITY.md` that directs reporters to GitHub's private vulnerability reporting feature (instead of public issues). Captures supported versions (currently: only `master`, no tagged release windows) and a scope statement (this repo's code; deps go upstream).

## Strengths

- **Reporters know where to go.** Without SECURITY.md, security issues tend to land in public issues, worsening disclosure.
- **Hooks into GitHub's built-in "Report a vulnerability" flow** (requires private vulnerability reporting to be enabled on the repo — separate toggle).
- **Minimal prose.** Nothing binding that's hard to keep.

## Weaknesses

- **Refers to `authors` in `pyproject.toml`** as fallback contact — only one name currently. Team should decide whether to list multiple or a shared alias.
- **Says "no tagged releases with support windows"** — accurate today but worth revisiting if the repo starts cutting releases.
- **Does not prescribe disclosure timelines.** Kept vague on purpose; team can tighten later.
- **Requires enabling private vulnerability reporting separately** (admin toggle) for the GitHub UI route to work.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,25 @@
+# Security Policy
+
+## Supported Versions
+
+`kg-microbe` is a research knowledge-graph pipeline developed in-tree on the `master` branch. We do not currently publish tagged releases with separate security support windows. Security fixes land on `master` and are picked up by consumers on next pull/build.
+
+## Reporting a Vulnerability
+
+Please **do not** open a public issue for vulnerabilities.
+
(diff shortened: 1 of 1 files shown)


realmarcin:

PR #597, opened 2026-07-20
Chemical mappings: retract 16 superseded groundings + culturebotai sync

realmarcin:
## What
Fixes additive re-seeding in the chemical-mapping consolidator: a corrected upstream grounding *added* the right row but couldn't *remove* the superseded one, so both were published for 161 ingredient names (16 unique stale objects) — e.g. Glycylglycine grounded to both CHEBI:15743 (butanal) and the correct CHEBI:17201.

## Changes
- **`mappings/kgm_retracted_groundings.tsv`** — committed, documented retraction list (161 rows / 16 objects), from the culturebotai-claw analysis.
- **`apply_retractions()`** — drops a listed stale object only when its sole source is `culturebotai_reviewed` (skips + warns on mixed-source). Runs after loads, before propagation.
- **`tests/test_consolidate_retractions.py`** — drop / skip-mixed / keep-unlisted / absent-noop / missing-file (5 tests).
- Bundles the culturebotai-claw upstream sync left uncommitted in the worktree (`_sync_vendored_from_sibling` generalization + auto-sync of `culturebotai_reviewed_ingredients.tsv`), and the regenerated unified file.

## Root cause (two, report found one)
- 13 are re-seeding phantoms (input no longer asserts them).
- 3 are loader re-asserts (`CHEBI:194474`, `CHEBI:30411`, `FOODON:03316428`): `best_primary` reads the stale `chebi_id` column before the corrected `kg_microbe_node_id`/`mim_id`. `best_primary` was **not** reordered (would shift 35 unrelated groundings). Filed upstream: CultureBotAI/MediaIngredientMech#138.

## Verification
Regenerated: dropped 16, already-absent 0, skipped-mixed 0; SSSOM valid (598,221 rows). 0 residual stale rows; all 161 names verified to still resolve to their correct grounding (zero orphans). 77/77 chemical-mapping + retraction tests pass.


realmarcin:
Review addressed: tightened the sole-source guard (exact / curator-qualified match only, empty-source objects now skipped) and added a non-fatal orphan guard (warns if any `now_asserted` target is absent post-drop; verified 0 false positives on the current list). Added idempotency + curator-qualified + empty-source tests. Deeper design items (name-level granularity, name_index cleanup, fatal orphan assertion) tracked in #599. Upstream root cause: CultureBotAI/MediaIngredientMech#138.

PR #596, opened 2026-07-20
kg-model-review: integrate monarch-initiative/kgxval biolink validator

realmarcin:
## What
Integrates the external [monarch-initiative/kgxval](https://github.com/monarch-initiative/kgxval) Biolink validator into the `kg-model-review` skill for merged-KG reviews.

kgxval uses BMT to check every unique (subject-category, object-category, predicate) signature against Biolink's **actual domain/range descendants** — an authoritative complement to the skill's hand-curated `DomainRange` check.

## Changes
- **`kgxval_validate.py`** — driver run in kgxval's own env. Builds the node→category map itself (splits KG-Microbe's pipe-delimited `category`, keeps `biolink:` components — kgxval's `TSVDictGen` doesn't split), then calls kgxval's `findSubObjErrorsForIngest`.
- **`kg_model_review.py`** — `--kgxval` flag (requires `--merged`). Extracts the merged tarball TSVs and runs the driver via `uv run --no-project --python 3.13 --with 'kgxval @ git+…'` (kgxval needs Python ≥3.13 and is not a kg-microbe dep). Appends a contextualized section: BAD BIOLINK split into METPO-native (expected) vs actionable; top domain/range violations. Skips gracefully when `uv` is absent/offline.
- **`SKILL.md`** — documents integration, prerequisites, output interpretation.

## Verification
Ran on the current merged KG (2.7M nodes / 13.6M edges): 284 error signatures — 183 BAD BIOLINK (67 METPO-native + 3 actionable: `owl:inverseOf`, `rdf:type`, `rdfs:subPropertyOf`), 59 BAD SUBJECT, 42 BAD OBJECT. `git+` source resolution verified via uv.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #601, opened 2026-07-20
kg-model-review: fold structural predicates into kgxval expected bucket

realmarcin:
## What
Fixes an accuracy bug in the kgxval summary added in #596. It split BAD BIOLINK predicates into METPO-native (expected) vs "other (actionable)", but the ontology metamodel axioms `rdfs:subPropertyOf` / `owl:inverseOf` / `rdf:type` — already accepted by the built-in review via `STRUCTURAL_PREDICATES` — were mislabeled **actionable**.

## Change
`_summarize_kgxval_csv` now buckets BAD BIOLINK into **METPO-native + structural (both expected) + other (actionable)**. On the current merged KG the report now reads:

> BAD BIOLINK predicates: 67 METPO-native + 3 structural (both expected), **0 actionable**.

(Was a misleading "3 actionable".) These predicates originate from the ontologies transform's KGX OBO-JSON loader — see the companion PR that drops them from the transform output.

### DIFF
--- a/.claude/skills/kg-model-review/kg_model_review.py
+++ b/.claude/skills/kg-model-review/kg_model_review.py
@@ -1563,13 +1563,20 @@ def _extract_tar_member(tar_path: Path, member_name: str, dest: Path) -> bool:
 def _summarize_kgxval_csv(out_csv: Path) -> str:
     """Render kgxval's error CSV as a contextualized markdown section.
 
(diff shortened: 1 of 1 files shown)


realmarcin:
Independent review found no HIGH/MEDIUM defects and no regression — the two LOW notes (exact-match whitespace/case; signatures-vs-distinct-predicates counting) are pre-existing and consistent with the module's existing checks. No changes needed.

PR #602, opened 2026-07-20
ontologies: drop metamodel-axiom edges (subPropertyOf/inverseOf/rdf:type)

realmarcin:
## What
Drops ontology **metamodel-axiom edges** from the ontologies transform output: `rdfs:subPropertyOf` (property hierarchy), `owl:inverseOf` (relation inverses), `rdf:type` (typing assertions, incl. ENVO's `WD_Entity:` xrefs). The KGX OBO-JSON loader passes these through verbatim with raw RDF/OWL predicate CURIEs — they aren't biolink entity relationships, so they clutter the merged KG (kgxval flags them non-biolink) without carrying queryable entity data.

## Change
- `_drop_metamodel_edges()` — called from `post_process` before KGX metadata is added; removes those rows from each ontology's `edges.tsv`. Nodes untouched.
- Verified on the real ENVO output: **252 edges removed** (166 subPropertyOf + 40 inverseOf + 46 rdf:type), 10,713 entity edges retained.
- Effect reaches the merged KG on the next `kg transform -s ontologies` + merge.
- Tests: drop-keeps-entity, no-op-when-none, idempotent, missing-file (4 tests).

Companion: the kgxval-summary PR stops mislabeling these as "actionable".

### DIFF
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -97,6 +97,21 @@ ONTOLOGIES_MAP = {
 }
 
 
+# Ontology metamodel axioms the KGX OBO-JSON loader emits verbatim: a
+# relation's inverse (owl:inverseOf), a property hierarchy (rdfs:subPropertyOf),
+# and rdf:type assertions. These are property-level / typing statements with raw
+# RDF/OWL predicate CURIEs — not biolink entity relationships — so they clutter
+# the merged KG (kgxval flags them as non-biolink) without carrying queryable
+# entity data. Dropped from the edge output; nodes are left untouched.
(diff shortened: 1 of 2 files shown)


realmarcin:
Review addressed: folded the metamodel drop into `_add_kgx_metadata_to_edges` as a single read/filter/write pass (was a redundant full read+rewrite of the largest edge files — ncbitaxon ~900K, chebi ~380K); `_drop_metamodel_edges` is now a pure DataFrame helper. Added `low_memory=False` and a column-preservation integration test. Accepted-by-design: a few endpoint nodes may be left orphaned (harmless in KGX; per-file node cleanup is unsafe as nodes can be referenced by other transforms). Filter keys on literal CURIE predicates, which all on-disk ontology edge files emit.

PR #603, opened 2026-07-21
ontologies: drop metamodel edges AFTER the biolink:->CURIE remap (fix #602)

realmarcin:
## What
Bug fix to #602: the metamodel-edge drop never actually ran on the transform output.

## Root cause
The drop was in `_add_kgx_metadata_to_edges`, which runs early in `post_process`, **before** `_normalize_schema` remaps KGX's meta-predicates from their `biolink:` serialization — KGX emits `biolink:subPropertyOf` / `biolink:inverseOf` / `biolink:type`, and `_normalize_schema` later remaps them to `rdfs:subPropertyOf` / `owl:inverseOf` / `rdf:type`. So the CURIE filter matched nothing.

Confirmed by a real `kg transform -s ontologies` run: it still emitted 252 envo + 75 metpo metamodel edges, and the output edge files carried the `knowledge_level` column — proving the early method ran but dropped nothing.

## Fix
Move the drop into `_normalize_schema`, immediately after the `biolink:`→rdfs/owl/rdf remap, where predicates are in final CURIE form. `_add_kgx_metadata_to_edges` reverts to metadata-only. Still single read/write (the drop rides `_normalize_schema`'s existing edge read/write).

## Verification
- 4 tests (df-helper + `_normalize_schema` remap-then-drop) pass; CI-ruff (0.15.22) clean.
- End-to-end on real data: **2,179 metamodel edges removed** across the 10 ontology edge outputs (ro 848, uberon 382, mondo 328, envo 252, pato 252, metpo 75, foodon 36, go 4, chebi 1, ec 1).

### DIFF
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -915,6 +915,10 @@ class OntologiesTransform(Transform):
             }
             if "predicate" in df.columns:
                 df["predicate"] = df["predicate"].replace(owl_meta_predicate_map)
+                # Belt-and-braces: obograph can emit bare local names (e.g.
+                # ``subPropertyOf``) rather than the biolink:-prefixed form, so
+                # normalise those too before the metamodel drop below.
+                df["predicate"] = df["predicate"].replace(owl_meta_relation_map)
             if "relation" in df.columns:
                 df["relation"] = df["relation"].replace(owl_meta_relation_map)
                 # Also catch cases where `relation` mistakenly got the biolink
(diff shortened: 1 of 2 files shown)


realmarcin:
Independent review confirmed the fix is correct (drop now runs after the biolink:->CURIE remap in _normalize_schema, which is the last predicate-touching step). Addressed the three LOW items: added belt-and-braces normalization of bare local-name predicates before the drop, restored end-to-end coverage that the reverted _add_kgx_metadata_to_edges adds metadata and drops nothing, and added a column-set/order assertion. 5 tests pass, ruff 0.15.22 clean.

Issue #604, opened 2026-07-21
Ontology .owl/.json/.db can drift out of version alignment (GO aspect map, ncbitaxon.db)


realmarcin:
The ontologies pipeline derives an ontology's artifacts from **different source files that are fetched independently**, so nothing guarantees the `.owl`, `.json`, and SemSQL `.db` used for one ontology are the same release. They happen to align today, but the architecture allows silent drift.

## Where the split is

**GO** is the clearest case:
- The transform TSV (`go_nodes.tsv`/`go_edges.tsv`) is built from **`go.json`** (`ONTOLOGIES_MAP["go"] = "go.json"`).
- The GO **aspect map** (MF/BP/CC categorization, `ontology_utils.get_go_category_by_aspect` → `_load_go_namespace_map`) reads **`go.db`**, which is a SemSQL build from **`go.owl`**.
- `download.yaml` fetches `go.json` (line 57) and `go.owl` (line 64) as **two separate downloads**.

So GO category correctness depends on `go.json` and `go.owl` being the same release. They are today (both `2026-05-19`), but if a `kg download` ever straddles a GO release boundary (GO ships ~monthly), the aspect map (`.db`/`.owl`) won't match the terms in the TSV (`.json`) → GO terms silently fall through to the `biolink:BiologicalProcess` default. (ChEBI is immune: everything derives from one `chebi.owl` — verified release 252 across `.owl`/`.json`/`.db`.)

**NCBITaxon** has an analogous split: metatraits/lpsn/bacdive do label + lineage lookups against `data/raw/ncbitaxon.db` (an OAK-fetched prebuilt SemSQL DB, `sqlite:obo:ncbitaxon`), while the ncbitaxon transform output is built from `data/raw/ncbitaxon_removed_subset.json` / `ncbitaxon.owl`. The OAK-cached `.db` release is whatever OAK last downloaded and is not pinned to the `.owl`/transform release. Lower risk (labels/lineage are stable) but the same class of problem.

## Related infra gap (same cluster)
The SemSQL `.db` files are not first-class build artifacts:
- `chebi.db` / `go.db` are built on demand from `.owl` inside the transform.
- `ncit.db` / `bto.db` / `po.db` are decompressed from `.db.gz` on demand.
- `ncbitaxon.db` is an OAK cache symlink.
[...]

PR #606, opened 2026-07-21
ontology .db infra: SemanticSQL CDN migration (#595) + go.db build & GO version gate (#604)

### LINKED ISSUE #595: Repoint hardcoded bbop-sqlite S3 URLs to the semanticsql CDN before raw-bucket access is retired
## Repoint hardcoded `bbop-sqlite` S3 URLs to the SemanticSQL CDN

`kg-microbe` downloads pre-built OAK/SemanticSQL SQLite databases by hardcoding the **raw S3 bucket URL**. The maintainers of that bucket (Berkeley BOP / INCATools SemanticSQL) are **retiring direct public access to the raw S3 bucket** and moving all traffic behind a CDN. Because these URLs are hardcoded (rather than resolved through OAK'\''s `sqlite:obo:` selector or the `semsql` package), they will **not** auto-migrate and will break once raw-bucket access is removed.

## Where this repo is affected

- `download.yaml` (on the default `master` branch) — raw `s3.amazonaws.com/bbop-sqlite/` URLs (e.g. `ncit.db.gz`, `bto.db.gz`, `po.db.gz`, ...)

## What to change

- Old: `https://s3.amazonaws.com/bbop-sqlite/<name>.db.gz`
- New: `https://semanticsql.berkeleybop.io/<name>.db.gz`

> [!IMPORTANT]
> **You must also send a non-default `User-Agent`.** The CDN sits behind a Browser Integrity Check that returns **HTTP 403 to default client User-Agents** (Python's `Python-urllib/3.x`, R's `download.file`/`httr`, and bare `curl`/`wget` defaults). A host-only swap will start returning 403 — set an explicit, non-default `User-Agent` header on the request.
>
> `kghub-downloader`/`requests`-based downloads send a default UA — set a custom `User-Agent` (e.g. via the downloader'\''s header config or a `requests` `headers={"User-Agent": ...}`).

## Why now

- Removal of raw-bucket public access is tracked in INCATools/semantic-sql#112.
- The object names/paths are **identical** — only the host changes (`s3.amazonaws.com/bbop-sqlite/` or `bbop-sqlite.s3.amazonaws.com/` → `semanticsql.berkeleybop.io/`).

_Filed as part of the coordinated downstream migration tracked in INCATools/semantic-sql#115. If you've already migrated or this is a non-issue, feel free to close — thanks!_

realmarcin:
Two related ontology `.db`-infra hardenings (same cluster surfaced this session when a 0-byte `go.db` silently miscategorized all 38,263 GO terms as `BiologicalProcess`).

## #595 — repoint bbop-sqlite S3 URLs to the SemanticSQL CDN
The raw `s3.amazonaws.com/bbop-sqlite` bucket is being retired (INCATools/semantic-sql#112). Repointed the 3 prebuilt SemSQL DB downloads (`ncit.db.gz`, `bto.db.gz`, `po.db.gz`) → `https://semanticsql.berkeleybop.io/<name>.db.gz`.
- Verified CDN files are **byte-identical** to the current local copies (Content-Length matches: ncit 550418722, bto 5234984, po 4637510).
- The CDN 403s default User-Agents, but **kghub-downloader's HTTP path already sends `User-Agent: Mozilla/5.0`** — no extra header config needed (verified). Documented inline. **Closes #595.**

## #604 — make `go.db` a built artifact + GO version-alignment gate
1. **Build `go.db` on demand.** The GO aspect map is read via a raw `sqlite3` query (deliberately, to bypass OAK's curies converter — GO has case-collision prefixes), so nothing ever built `go.db` like `chebi.db`. `_ensure_go_db` now builds it from `go.owl` via `semsql make` when missing/empty (wired into `_load_go_namespace_map`); degrades gracefully if `go.owl`/`semsql` are absent.
2. **Version-alignment gate.** The GO TSV is built from `go.json` while the aspect map comes from `go.db`←`go.owl`; a `kg download` straddling a GO release makes them diverge → silent MF/CC→BP miscategorization. `assert_go_version_alignment` compares the two release stamps and fails loudly on mismatch (`strict=False` warns). Wired into `_fix_node_categories`'s GO branch.

Partial fix for **#604** (GO covered — the demonstrated failure; NCBITaxon `.db` release pinning left as a follow-up in the issue).

## Verification
- 9 unit tests (release parsing, aligned/mismatch/missing gate paths, build skip/graceful-degrade). CI-ruff (0.15.22) clean.
- End-to-end on current aligned (2026-05-19) sources: gate passes silently, build skipped, GO categorizes correctly (MF/BP/CC).

### DIFF
--- a/download.yaml
+++ b/download.yaml
@@ -470,20 +470,28 @@
 #   mesh      — NLM N-Triples RDF, streamed once per run (mesh_nt_rdf —
 #               emits hierarchy edges). See the entry further below.
 #
-# MICRO's bbop-sqlite distribution at
-# https://s3.amazonaws.com/bbop-sqlite/micro.db.gz is a 29-byte truncated
-# placeholder rather than a real SemSQL DB; PO's SemSQL .db is fine but
-# the OWL path is the one the stub transform exercises for hierarchy.
-# The MICRO and PO OWLs are declared above in the ontologies section.
+# The prebuilt SemSQL DBs below are fetched from the SemanticSQL CDN
+# (https://semanticsql.berkeleybop.io/<name>.db.gz), NOT the raw
+# s3.amazonaws.com/bbop-sqlite bucket, whose direct public access is being
+# retired (INCATools/semantic-sql#112; see kg-microbe #595). Object names are
+# identical — only the host changed. The CDN sits behind a Browser Integrity
+# Check that 403s default client User-Agents; kghub-downloader's HTTP path
+# already sends `User-Agent: Mozilla/5.0` (download.py), so no extra header
+# config is needed here.
+#
+# MICRO's SemSQL distribution (micro.db.gz) is a 29-byte truncated placeholder
+# rather than a real SemSQL DB; PO's SemSQL .db is fine but the OWL path is the
+# one the stub transform exercises for hierarchy. The MICRO and PO OWLs are
+# declared above in the ontologies section.
 #
 -
-  url: https://s3.amazonaws.com/bbop-sqlite/ncit.db.gz
+  url: https://semanticsql.berkeleybop.io/ncit.db.gz
   local_name: ncit.db.gz
 -
-  url: https://s3.amazonaws.com/bbop-sqlite/bto.db.gz
+  url: https://semanticsql.berkeleybop.io/bto.db.gz
   local_name: bto.db.gz
 -
-  url: https://s3.amazonaws.com/bbop-sqlite/po.db.gz
+  url: https://semanticsql.berkeleybop.io/po.db.gz
   local_name: po.db.gz
 
 #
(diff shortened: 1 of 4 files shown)


realmarcin:
Review addressed. **#1 (MEDIUM)** — the gate no longer offers only a hard-raise: default stays fail-loud, but `KG_GO_VERSION_CHECK=warn` downgrades to a warning as an escape hatch for the release-stamp heuristic (the fail-safe direction was already fine — unreadable/unstamped → no-op, so only two readable-but-different stamps trip it). Also: corrected the build-time estimate (10-30+ min, not 'a few minutes'), clarified the UA lives in kghub-downloader's own download.py, and added the env-downgrade + a wiring test (verifies `_fix_node_categories` actually calls the gate). Confirmed non-defects from review: **#5** the UA claim is correct (https → urllib handler with Mozilla/5.0; the UA-less requests.get is git-scheme only). Deferred (LOW): corrupt-partial-db atomicity (#3), build file-lock (#4) — theoretical today (GO categorization is single-process); NCBITaxon .db release pinning stays a follow-up in #604.

PR #607, opened 2026-07-22
ontologies: NCBITaxon .owl/.db version-alignment gate (completes #604)

realmarcin:
Completes the remaining piece of **#604** (NCBITaxon `.db` release pinning; the GO portion + version-gate framework shipped in #606).

## Problem
metatraits/lpsn/bacdive resolve taxa against the OAK-fetched `ncbitaxon.db` (whose release is whatever OAK last downloaded), while the NCBITaxon **transform** emits nodes from `ncbitaxon.owl`. Nothing pinned the two to the same release, so a drifted OAK cache could resolve lookups against different taxa than the transform emitted.

## Change
- **`assert_ncbitaxon_version_alignment(db_path)`** — compares `ncbitaxon.owl`'s versionIRI release with the db's `owl:versionInfo` (read via a tiny SemSQL query, `_ncbitaxon_db_release`) and **fails loudly on mismatch**; `KG_NCBITAXON_VERSION_CHECK=warn` downgrades to a warning. No-op when either stamp is unreadable (fail-safe — only two readable-but-different stamps trip it).
- Generalized `_obo_release_from_head` to match NCBITaxon's `versionIRI` form (`ncbitaxon/DATE/`, which has no `releases/` segment, unlike GO).
- Factored the shared strictness resolution into `_version_check_strict` (used by both gates).
- Wired into metatraits `_ensure_ncbitaxon_db_ready` (both the happy-path and cache-repair success branches — the pre-flight validation that all consumers rely on).

## Verification
- Current sources both `2026-05-13` → gate passes silently (verified).
- 7 new tests (owl/db release parsing, aligned / mismatch / env-warn / missing-stamp); GO gate + its 11 tests unchanged. CI-ruff (0.15.22) clean.

With this, **#604 is fully addressed** — both GO and NCBITaxon `.owl`/`.db` alignment are now gated.

### DIFF
--- a/kg_microbe/transform_utils/metatraits/metatraits.py
+++ b/kg_microbe/transform_utils/metatraits/metatraits.py
@@ -108,12 +108,15 @@ def _ensure_ncbitaxon_db_ready() -> None:
 
     Raises RuntimeError with remediation steps if no valid DB can be found.
     """
+    from kg_microbe.utils.ontology_utils import assert_ncbitaxon_version_alignment
+
     local_db, oak_cache = _ncbitaxon_db_paths()
 
     # Happy path: symlink resolves to a valid DB.
     ok, reason = _validate_ncbitaxon_db(local_db)
     if ok:
         print(f"  NCBITaxon DB validated: {local_db} -> {local_db.resolve()}")
+        assert_ncbitaxon_version_alignment(str(local_db))
         return
 
     print(f"  NCBITaxon DB at {local_db} invalid ({reason}); attempting repair")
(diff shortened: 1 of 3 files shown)


realmarcin:
Review addressed. **#1 (MEDIUM, the important one)** — the gate only ran under `if use_mp:`, skipping sequential metatraits; moved it to run **unconditionally** in run() (no-ops if the db isn't materialized yet). **#2 (MEDIUM)** — NCBITaxon now **defaults to warn** (the OAK cache vs pinned ncbitaxon.owl legitimately drift, and labels/lineage are stable), with `KG_NCBITAXON_VERSION_CHECK=strict` to opt into fail-loud; GO stays strict. **#3** — versionInfo query now targets the ontology subject (`subject LIKE '%ncbitaxon%'`) + a decoy-row test. **#4** — docstring/message corrected (only metatraits invokes it). Added subject-targeting, corrupt-db→None, default-warn and env-strict tests (21 total). Confirmed non-issues from review: regex has no GO regression, exception handling is adequate, symlink read works, _version_check_strict refactor is behavior-preserving. **#5** (dead `if not NCBITAXON_SOURCE` guard) left as harmless defensive parity with the GO gate.

PR #612, opened 2026-07-27
Fix MediaDive cache corruption; add tag-filtered downloads; make every ontology single-source

realmarcin:
Started as a fix for a `kg download` crash and grew to cover the data-acquisition
gaps it exposed. 12 commits, each independently reviewable.

## 1. MediaDive cache corruption (`96c1b3e`, `ca0b8e6`)

`kg download` died at request 3 of 3338 with
`sqlite3.DatabaseError: database disk image is malformed`, against a cache that
was provably intact — `PRAGMA integrity_check` OK, full 6044-row scan fine from
the same interpreter, and an mtime showing the run never wrote to it.

Cause is in `requests_cache`'s SQLite backend: `SQLiteDict` holds a **single**
`sqlite3.Connection` for all callers (opened `check_same_thread=False`) and
takes no lock on reads. `download_detailed_media` created one session in the
main thread and shared it across a `ThreadPoolExecutor`, so five workers issued
concurrent `execute()` calls on one connection — which corrupts its pager state
and makes SQLite report `SQLITE_CORRUPT` about a healthy file.

Each thread now builds its own `CachedSession`, so its own connection; several
connections to one SQLite file is supported usage, and cached responses are
still shared via the database. Also replaces the global
`requests_cache.install_cache()` monkeypatch with explicit module-scoped
caching, moves the DB out of the repo root into `data/raw/mediadive/`, and makes
`--ignore-cache` actually clear the response cache (it previously rebuilt the
bulk JSON from the same cached responses, so the flag could not refresh
anything).

Verified by replaying 1500 cached URLs through 5 threads against a copy of the
real 51 MB cache: 1500 hits, 0 errors, 0 network.

## 2. Tag-filtered downloads (`0789241`, `2b58fdf`, `7f6b401`, `cb40845`, `708be03`)

`kg download` had no way to fetch a subset — all 62 entries or `-i`, which
re-downloads everything. `kghub-downloader` already supports tag filtering; it
was unwired end to end. All active entries are now tagged across 15 groups, with
a repeatable `-t/--tag`.

Two things the bare library filter gets wrong for us: an unknown tag downloads
zero files and exits 0 (now rejected up front, listing valid tags), and the
MediaDive bulk step fired whenever `mediadive.json` existed — so
`kg download -t ontologies` would have started an hour of MediaDive API calls
(now gated on the tag).

Six previously-unfetchable inputs (five MetaTraits/GTDB files behind a dead
endpoint, BactoTraits behind a broken TLS chain) are now declared normally and
pinned to hosted copies, with their expected md5 recorded inline. Sources whose
replacement is not hosted yet use a `REPLACE_ME_` placeholder that
`kg download` skips and reports, so a placeholder can never abort a run.
Verified end to end: all six fetch and reproduce their pinned checksums.

Research note: the `gtdb_*`/`NCBI2GTDB`/`GTDB2NCBI` files are **not** GTDB
downloads, despite the names — GTDB publishes only taxonomy, metadata, trees
and MSA files. They are MetaTraits (EMBL) products keyed by GTDB taxonomy. Two
are still live upstream but deliberately pinned rather than fetched, because
that location is an unversioned personal directory regenerated in place.

## 3. Every ontology is now single-source (`55b9df8`, `5025515`, `ca04e11`)

Extends the GO single-source rule (#604) to the remaining lookup DBs, which were
acquired independently of the OWL the transforms emit nodes from.

- **NCBITaxon** had a real 2-month gap: `ncbitaxon.owl` 2026-07-12 vs the OAK
  cache's 2026-05-13. Not closable by refreshing — the upstream CDN's newest
  build was 2026-05-24, so the documented remediation would have re-downloaded
  2.3 GB and landed back at the same release. Now built from our OWL via
  `semsql make`; the real 13.1 GB build completed and both stamps read
  2026-07-12.
- **ChEBI** could not even be checked: it versions by an integer
  (`versionIRI .../chebi/253/chebi.owl`), invisible to the date-based reader, so
  a gate built on it would pass silently forever. Confirmed against the real
  file — the new reader returns `253` where the shared one returns `None`. Both
  category-fixing call sites also went through
  `get_adapter(f"sqlite:{CHEBI_SOURCE}")` wrapped in a bare `except`, which
  could not work as written (`CHEBI_SOURCE` is an `.owl` path), meaning the
  fallback silently accepted whatever `chebi.db` was on disk.
- **PO** had nothing to sync: its stub import is `owl_mireot` against `po.owl`,
  nothing reads `po.db`, and the downloaded `po.db.gz` was never decompressed.
  Removed the dead source instead, and corrected a comment that claimed the
  opposite.

## Review round

Reviewed after opening; 7 issues filed, 6 fixed in `d5ba68b` (#613, #615, #616,
#617, #618, #619). Highlights:

- **#613** — the new `semsql` builds could start unprompted from an ordinary
  `kg transform` and run for hours. `KG_SEMSQL_BUILD=off` now declines, falling
  back to a prebuilt DB with the gate warning about the drift.
- **#615** — `chebi.owl.gz` was decompressed onto its final path, so an interrupt
  left a truncated ~1 GB OWL that nothing could detect (the version stamp is in
  the head and still parses). Now temp-file + `os.replace`.
- **#616** — `get_chebi_adapter` ignored a failed build and handed OAK a missing
  path; now raises with remediation.

**#614 is left open**: `_ensure_ncbitaxon_db_ready()` — and so the NCBITaxon
build — runs only under `if use_mp:`. A sequential `metatraits` run keeps the lazy
OAK fetch, so `ncbitaxon.owl`/`.db` can still diverge there; the gate detects it
and warns, but nothing prevents it. The single-source guarantee above is therefore
**conditional on multiprocessing mode** for NCBITaxon. Moving that call interacts
with the worker-concurrency rationale and deserves its own change.

## Testing

`poetry run tox` fully green: **519 passed**, format/lint/codespell/docstr-coverage
all OK. 108 new tests. One test that had been failing on `master` for an unrelated
reason (missing `ncbitaxon.db`) now passes.

The decompressed-crosswalk guard was mutation-tested — restoring the old
`gzip.open` makes it fail.

## Note on data files

This branch touches no data under `data/`. The `ncbitaxon.db` build and the
copied reference inputs are local artifacts, all gitignored.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,6 +21,14 @@ poetry install
 # Download all data sources (configured in download.yaml)
 poetry run kg download
 
+# Download only selected source groups (tags are listed in download.yaml's header)
+poetry run kg download -t ontologies -t gtdb
+
+# Re-download a single file: delete it and re-run. `kg download` only fetches
+# files that are absent — it never checks whether the remote copy is newer.
+# `-i` ignores the cache for EVERY entry, so scope it with -t:
+poetry run kg download -i -t mediadive
+
 # Transform downloaded data into KG format (TSV: nodes.tsv, edges.tsv)
 poetry run kg transform
 
--- a/docs/DATA_HOSTING.md
+++ b/docs/DATA_HOSTING.md
@@ -9,20 +9,28 @@ the only options are to host a copy or to copy one by hand from a reference
 This file is the manifest for hosting those copies on the public LBNL Google
 Drive, plus the interim copy-by-hand procedure.
 
-## Files to host
+## Hosted files
 
-Six files, ~81 MB total. Checksums are from the copies currently in `data/raw`,
-which came from `data/raw_202607_andOLD_mixed/`. Each md5 is also recorded inline
-in `download.yaml` next to its placeholder.
+Six files, ~81 MB total, in the shared Drive folder
+<https://drive.google.com/drive/folders/1vCKoGREJk45BzImfnPVtGhXAb_DyndLq>
+(General access: *Anyone with the link*, Viewer — `gdown` runs
+unauthenticated, so an org-restricted folder returns a sign-in page and
+fails with HTTP 401). Uploaded and verified end to end 2026-07-25.
 
-| file | size | md5 | placeholder token in download.yaml |
+Each md5 below is also recorded inline in `download.yaml` beside its URL, so the
+pinned artifact is identifiable from the config alone. They were confirmed by
+moving all six out of `data/raw`, running
+`kg download -t metatraits_gtdb -t bactotraits`, and re-checksumming what
+arrived.
+
+| file | size | md5 | Drive file ID |
 |---|---|---|---|
-| `gtdb_species_summary.jsonl.gz` | 46 MB | `aca98ae9c978f238ae7844beaccec63b` | `REPLACE_ME_gtdb_species_summary` |
-| `gtdb_genus_summary.jsonl.gz` | 16 MB | `159828de09d22a7c85e2116e8a97944a` | `REPLACE_ME_gtdb_genus_summary` |
-| `gtdb_family_summary.jsonl.gz` | 4.4 MB | `1443132d2aa0f3d8743b23096ce23a13` | `REPLACE_ME_gtdb_family_summary` |
-| `NCBI2GTDB.tsv.gz` | 2.8 MB | `eca29d35869dd9035730caa12de12598` | `REPLACE_ME_ncbi2gtdb` |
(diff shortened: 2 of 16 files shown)


realmarcin:
Split into two PRs, because the two halves have very different risk profiles.

- **#627** — MediaDive cache corruption, tag-filtered downloads, pinned inputs. Self-contained; five review rounds found no defect in it. Ready to review.
- **#628** — the SemSQL single-source work (ncbitaxon.db / chebi.db built from our OWL). Stacked on #627. This is where every regression across those rounds occurred, so it gets its own review.

A + B reproduces this PR's diff exactly. Continuing on #628.

Issue #626, opened 2026-07-27
Pinned md5s for Drive-hosted inputs are comments, not checks


realmarcin:
Found reviewing #612 (F12).

`docs/DATA_HOSTING.md` and `download.yaml` record an md5 beside each Drive-hosted
file, and the stated rationale for hosting is "The Drive copy fixes the release".
But the md5s are YAML comments — nothing verifies them, and kghub-downloader
0.3.11's `DownloadableResource` has no checksum field. The guarantee currently
rests on nobody replacing the Drive file, which is the same silent-substitution
risk the doc rejects when arguing against the unversioned upstream directory.

Options: add a post-download checksum check in `kg_microbe/download.py` (it
already post-processes downloads for MediaDive), or soften the doc wording to
"identifiable, not verified".

Related: `assert_chebi_version_alignment`'s docstring justifies defaulting to
*warn* by "when semsql is unavailable the pipeline legitimately falls back to a
prebuilt chebi.db". semsql ships transitively with oaklib, so
`shutil.which("semsql")` is essentially never None inside `poetry run` — that
fallback is not a real production state, and the warn-vs-strict default may
deserve revisiting on that basis.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #625, opened 2026-07-27
_ensure_ncbitaxon_db's reuse path accepts a large-but-corrupt DB


realmarcin:
Found reviewing #612 (F5).

The reuse branch gates on size + release stamps only. A corrupt DB clears the
size check, `_ncbitaxon_db_release` swallows `sqlite3.Error` and returns None, so
`not (owl_release and db_release and ...)` is True and the function returns True
before reaching the build.

Reproduced with garbage bytes at `ncbitaxon.db` (threshold shrunk, semsql
mocked present): returns True, `semsql make` never runs.

This matters because of where it is called from: `_ensure_ncbitaxon_db_ready`
only reaches it *after* `_validate_ncbitaxon_db` has already failed. Failures for
size or absence are re-detected and do rebuild. A failure because the OAK query
raised — size fine, i.e. the multi-worker corruption the function was written for
— returns True, prints "still invalid after ensure", and falls through to the OAK
cache without rebuilding.

Fix: have the reuse path do a cheap integrity probe (or accept the validation
verdict the caller already has) before short-circuiting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

realmarcin:
Fixed on `feat/semsql-single-source` (#633), commit `ec703a1`.

The reuse fast-path now probes that the DB is actually openable, not just that it clears the size floor:

```python
if _present_db(db_path, min_size) and _db_is_readable(db_path):
```

`_db_is_readable` opens the file read-only and runs `PRAGMA schema_version`, which reads the header only — O(1) even on the 13 GB `ncbitaxon.db`, unlike `PRAGMA integrity_check` which would walk every page and is far too expensive to run on every resolve.

The root cause was that every release reader converts `sqlite3.Error` to `None`, and the reuse path reads `None` as "no stamp recorded, do not rebuild" — so a corrupt-but-large DB was handed to OAK, whose queries then failed inside broad handlers and collapsed terms to default categories. Applied to all four ontologies, not just NCBITaxon; regression test at `tests/test_ncbitaxon_db_build.py::TestDegradesGracefully::test_corrupt_but_large_db_is_rebuilt_not_reused`.

Will close when #633 merges.

Issue #614, opened 2026-07-27
NCBITaxon single-source rule is enforced only in multiprocessing mode


realmarcin:
From review of #612.

`_ensure_ncbitaxon_db_ready()` — the only caller of the new
`_ensure_ncbitaxon_db` build — is invoked conditionally:

```python
if use_mp:
    _ensure_ncbitaxon_db_ready()
```

So a sequential `metatraits` run (`METATRAITS_MULTIPROCESSING=false`, or a single
unsplit input file) never builds the DB and keeps the old lazy OAK fetch, which
pulls whatever prebuilt release the upstream CDN last published. The
`ncbitaxon.owl`/`ncbitaxon.db` releases can therefore still diverge in sequential
mode; `assert_ncbitaxon_version_alignment` runs regardless and warns, so the gap
is detected but not prevented.

The MP-only placement is deliberate — the surrounding comment explains workers
must never heal the cache concurrently — but the consequence is that #612's
"single source" guarantee is conditional. Either the build should be hoisted so it
runs in both modes (it is parent-side and safe either way), or the limitation
should be documented where the guarantee is claimed.

Note this is not a regression: before #612 neither mode built from the OWL. It is
an inconsistency in the new behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

realmarcin:
The NCBITaxon drift-rebuild pre-flight (`_ensure_ncbitaxon_db_ready` in `metatraits.py:112`) now runs from the metatraits pre-flight regardless of MP mode — sequential runs (`METATRAITS_MULTIPROCESSING=false` or a single unsplit input) invoke the same pre-flight, so the warn-and-continue on release drift is closed there too. Documented in CLAUDE.md ('The NCBITaxon drift rebuild runs from the metatraits pre-flight regardless of MP mode'). Closing as done.

PR #628, opened 2026-07-27
Build ncbitaxon.db and chebi.db from the OWL we ship (#604)

realmarcin:
Split out of #612 and **stacked on #627** — review/merge that one first; this
PR's diff is only the SemSQL half.

Extends the GO single-source rule (#604) to the remaining ontology lookup DBs,
which were acquired independently of the OWL the transforms emit nodes from.

## Why

**NCBITaxon** had a real 2-month gap: `ncbitaxon.owl` at 2026-07-12 against an
OAK cache at 2026-05-13. Not closable by refreshing — the upstream SemSQL CDN's
newest build was 2026-05-24, so the remediation the old error message gave would
have re-downloaded 2.3 GB and landed back at the same release. Building from our
OWL makes the two equal by construction; the real 13.1 GB build completed and
both stamps now read 2026-07-12.

**ChEBI** could not even be *checked*. It versions by an integer
(`versionIRI .../chebi/253/chebi.owl`), invisible to the date-based reader, so a
gate built on that reader would have passed silently forever — verified against
the real file, the ChEBI reader returns `253` where the shared one returns `None`.
Both category-fixing call sites also went through
`get_adapter(f"sqlite:{CHEBI_SOURCE}")` wrapped in a bare `except`, which cannot
work as written (`CHEBI_SOURCE` is an `.owl` path), so the fallback silently
accepted whatever `chebi.db` happened to be on disk.

## Also included

`KG_SEMSQL_BUILD=off` (these builds run for hours and are triggered by ordinary
`kg transform` commands), atomic `.gz` decompression, gz-aware release reads, a
min-size guard on every no-build exit, and preservation of an existing DB across
a failed build.

## Status — please review carefully

This half is where **every** regression across five review rounds of #612
occurred, which is why it was split out. Known open issues against it:

- #614 — the NCBITaxon build runs only in multiprocessing mode, so sequential
  runs can still drift (detected and warned, not prevented)
- #625 — the reuse path accepts a large-but-corrupt DB
- #626 — the pinned md5s are unenforced comments

A fifth review round raised further findings on the most recent fixes here
(pre-flight returning success for a DB it declared invalid; the success path
returning a pre-restore verdict; `.prev` orphaned on interrupt; the symlink
rejection being wrong at the no-build exits; a `.json.gz` fallback making an
`unlink()` unsafe; `_ensure_go_db` never updated). Those are addressed in the
commits that follow.

### DIFF
--- a/kg_microbe/transform_utils/metatraits/metatraits.py
+++ b/kg_microbe/transform_utils/metatraits/metatraits.py
@@ -72,6 +72,24 @@ def _ncbitaxon_db_paths() -> Tuple[Path, Path]:
     )
 
 
+def _db_fingerprint(db_path: Path) -> Optional[Tuple[int, int]]:
+    """
+    Return (mtime_ns, size) for a DB, or None when it is absent.
+
+    Used to tell "this call built a new DB" from "the DB that was already there",
+    which decides whether the OAK-cache fallback may replace it. Checking "is a
+    real file" instead would also match a pre-existing DB and strand the run.
+
+    :param db_path: Path to fingerprint.
+    :return: (mtime_ns, size), or None if the path cannot be stat'ed.
+    """
+    try:
+        st = db_path.stat()
+    except OSError:
+        return None
+    return (st.st_mtime_ns, st.st_size)
+
+
 def _validate_ncbitaxon_db(db_path: Path) -> Tuple[bool, str]:
     """
     Validate that db_path is a healthy NCBITaxon SQLite DB.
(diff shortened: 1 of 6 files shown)


realmarcin:
Superseded — see the replacement PR. GitHub auto-closed this when its base branch was deleted on merge of #627, and then refused both to reopen it ("base branch has been deleted") and to retarget it ("cannot change the base branch of a closed pull request"). Same branch, rebased onto master.

Issue #624, opened 2026-07-27
MediaDiveTransform still installs the global requests cache


realmarcin:
Found reviewing #612 (F7). Left for a separate change — it is in the MediaDive
*transform*, not the download path #612 rewrote.

`kg_microbe/transform_utils/mediadive/mediadive.py:139` calls
`requests_cache.install_cache("mediadive_cache")` in `__init__`, monkeypatching
`requests.Session` process-wide. #612 removed the equivalent call from
`mediadive_bulk_download.py` and its module comment now claims "Caching is now
explicit and scoped to this module" — true for `kg download`, not for
`kg transform`. The root cause #612 fixed (one shared `sqlite3.Connection`
across threads) is therefore still present on the transform side if that code
ever threads.

It also makes `test_caching_is_not_installed_globally` order-dependent — it
passes only because alphabetical collection runs `test_mediadive_*` before
anything that constructs a `MediaDiveTransform`:

```
$ poetry run pytest tests/test_transform_category_alignment.py tests/test_mediadive_bulk_download.py -q
FAILED tests/test_mediadive_bulk_download.py::TestCacheThreadSafety::test_caching_is_not_installed_globally
1 failed, 33 passed
```

In that state `test_disabled_cache_yields_plain_sessions` also becomes
tautological, since `requests.Session` *is* `CachedSession`.

Fix: give the transform the same explicit per-thread `CachedSession` treatment,
then the guard test stops being order-sensitive.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #633, opened 2026-07-27
Build ncbitaxon.db and chebi.db from the OWL we ship (#604)

realmarcin:
Replaces #628, which GitHub auto-closed and then refused to reopen when its base
branch (`fix/mediadive-cache-and-tag-downloads`, now merged as #627) was deleted.
Same branch, rebased onto `master`; the diff is only the SemSQL half.

Extends the GO single-source rule (#604) to the remaining ontology lookup DBs,
which were acquired independently of the OWL the transforms emit nodes from.

## Why

**NCBITaxon** had a real 2-month gap: `ncbitaxon.owl` at 2026-07-12 against an
OAK cache at 2026-05-13. Not closable by refreshing — the upstream SemSQL CDN's
newest build was 2026-05-24, so the remediation the old error message gave would
have re-downloaded 2.3 GB and landed back at the same release. Building from our
OWL makes the two equal by construction; the real 13.1 GB build completed here
and both stamps now read 2026-07-12.

**ChEBI** could not even be *checked*. It versions by an integer
(`versionIRI .../chebi/253/chebi.owl`), invisible to the date-based reader, so a
gate built on that reader would have passed silently forever — verified against
the real file, the ChEBI reader returns `253` where the shared one returns `None`.
Both category-fixing call sites also went through
`get_adapter(f"sqlite:{CHEBI_SOURCE}")` wrapped in a bare `except`, which cannot
work as written (`CHEBI_SOURCE` is an `.owl` path), so the fallback silently
accepted whatever `chebi.db` happened to be on disk.

## Also included

`KG_SEMSQL_BUILD=off` (these builds run for hours and are triggered by ordinary
`kg transform` commands), atomic `.gz` decompression, gz-aware release reads, a
min-size guard on every no-build exit, and preservation of an existing DB across
a failed build.

## Please review carefully — this half is where the defects were

#612 was split precisely because **every** regression across five review rounds
was in this code, several of them introduced by the previous round's fix. The
most recent round found: the pre-flight returning success for a DB it had just
declared invalid, the build's success path returning a pre-restore verdict, a
`.prev` orphaned by the very interrupt case the keep-aside was justified by, and
a symlink check applied where it was wrong. All are fixed here, with a new
`test_ncbitaxon_db_ready.py` pinning the pre-flight contract (mutation-checked —
reinstating the regression fails three of its tests).

Known open issues against this half:

- #614 — the NCBITaxon build runs only in multiprocessing mode, so sequential
  runs can still drift (detected and warned, not prevented)
- #625 — the reuse path accepts a large-but-corrupt DB
- #626 — the pinned md5s are unenforced comments

## Testing

`poetry run tox` green: **563 passed**, format/lint/codespell/docstr-coverage all
OK.


realmarcin:
## Codex adversarial review addressed

All 8 findings from the Codex review, plus 4 more instances the follow-up sweep found. Six commits, `ec703a1..a6ef6d8`. CI green on 3.10/3.11/3.12; `tox` green locally (624 tests, was 597).

Every finding was re-verified against the source before being acted on — Codex's sandbox blocked temp files, so its "reproductions" were mocks of its own model of the code rather than real runs. One of its sub-claims did not survive that check and was **not** changed: it reported that a usable symlink at `.prev` is left in place for `semsql` to write through, but the existing fall-through at `ontology_utils.py:493-495` already removes it.

### Two root causes, fixed structurally

**1. Lazy resolution put fatal errors inside per-item recovery handlers** (F4, F6, + 2 more). The first attribute access on a proxy lands inside a `try` whose `except Exception` was written to absorb a lookup miss — so a missing DB or a deliberate `*_VERSION_CHECK=strict` abort was swallowed and the pipeline emitted a wrong graph with exit code 0: every ChEBI node `biolink:ChemicalEntity`, every GO term `molecular_function`, every label a bare numeric ID, every protein→GO edge dropped, and in `consolidate_chemical_mappings` a synonym-stripped SSSOM written back as the next run's baseline.

Patching each handler is the obvious fix and had already been tried — it regressed three times, because the next broad handler reopens the hole. Instead `FatalOntologyError(BaseException)`, with `OntologyDbUnavailableError` and a new `OntologyVersionMismatchError` under it. No `except Exception` — here, or in oaklib/pandas — can turn one into wrong data; `finally` still runs. Verified there are zero `except RuntimeError` handlers in the tree first. The one deliberate degrade (standalone `get_chebi_category` with no DB) is preserved and tested.

**2. Derived caches were written in place** (F2). `get_go_category_trees` wrote the header, then resolved the adapter — so an unusable `go.db` left a header-only file that the bare `.exists()` guards accept forever, making `prepare_go_dictionary` return `{}`. New `atomic_io.atomic_write` (temp + `os.replace`, cleanup in `finally` so `BaseException` unwinds cleanly), applied to that plus the three sibling caches with the same shape.

### Point fixes

| | |
|---|---|
| **F1** | `_clear_build_target`'s symlink branch returned before the fall-through that moves the recovered DB aside, so `semsql make` overwrote the 13 GB copy and a failed build restored the dangling symlink. Third recurrence of #634. |
| **F3** | `reuse_on_failure=False` covered only preflight; both post-build failure exits now honour it. |
| **F5** | New `_db_is_readable` (`PRAGMA schema_version`, O(1) on 13 GB) gates all four reuse paths. **This also closes #625.** |
| **F7** | Per-ontology lock around resolution — `lru_cache` does not lock across the wrapped call. |
| **F8** | Removed rhea's unchecked `os.system("gzip -d ec.owl.gz")` at construction. |

### Behaviour changes worth a reviewer's attention

- With no `semsql` on PATH and a drifted `go.db`, GO-dependent transforms now **abort** rather than run on stale categories. That is what GO's strict gate is for. `KG_SEMSQL_BUILD=off` remains exempt.
- `BaseException` in a `multiprocessing.Pool` worker can hang the pool. Metatraits is safe because its workers resolve NCBITaxon eagerly in the parent via a module-local adapter — that is now a **documented exception** to "every call site is unified", not an oversight. Recorded on the class and in CLAUDE.md.

### Guard hardening

Codex defeated both AST-guard assertions with one snippet (`db_path = RAW_DATA_DIR / ("go" + ".owl")`). `_mentions_owl` now inspects string literals individually and follows indirection three ways (plain assignment, tuple unpacking, calls to local functions via their returns); the `.db` assertion rejects anything reaching an `.owl` before accepting a db-shaped parameter. The unparsable allow-listed file gets a regex fallback, since on 3.10/3.11 anything added to it escaped every assertion. Five known bypasses pinned as tests, plus two negative cases. Verified by planting the bypass in a real file and watching the guard fail.

Each fix was mutation-checked — reverted, confirmed the new test fails, restored.

### Not addressed

#614 and #624 remain open as scoped. #626 (unenforced md5s) remains open.

PR #680, opened 2026-08-06
fix(kg): repair three path-review defects in transform outputs

realmarcin:
Found by a `kg-path-review` sweep. All three are invisible to single-edge conformance checks (`kg-model-review` reports the merged KG as clean), because each is a *path* or *provenance* defect rather than a malformed row.

## 1. BacDive — 62k orphan strain→LPSN edges

The LPSN cross-ref edge used `bacdive:<id>` as its subject, but this transform only ever emits strain nodes as `kgmicrobe.strain:bacdive_<id>` (99,392 of them). Every one of the **62,096** edges was orphaned, and KGX synthesized **62,521 bare `biolink:NamedThing` stubs** (`provided_by=Graph`, no name) into the merged KG.

Both call sites now go through a shared `_strain_curie()` helper so the node-emitting path and the edge-emitting path cannot drift apart again.

**Predicate also changed: `biolink:in_taxon` → `biolink:subclass_of` / `rdfs:subClassOf`.** Rationale:
- Both endpoints are typed `biolink:OrganismTaxon`.
- This transform *already* relates the same strain nodes to their NCBITaxon species with `subclass_of` (99,381 edges), and the `lpsn` transform uses `subclass_of` for its own hierarchy (29,695 edges). `in_taxon` was the lone outlier for the same semantic relation.
- `in_taxon` asserts instance-of, which contradicts the class-like `OrganismTaxon` typing already in use; NCBITaxon likewise models strains as classes `subclass_of` their species.
- On Biolink purity: `organism taxon` descends from neither `thing with taxon` (in_taxon's domain) nor `ontology class` (subclass_of's domain) in 4.2.2, so both are formally out-of-domain. `subclass_of` folds into an existing, accepted violation shape instead of adding a second, rarer one.

## 2. Ontologies — EC edges header written twice

The `ec` branch drops the incoming header by matching `line.startswith("id")` — the **node** header's first column. An edges header starts with `subject`, so it never matched and survived as a data row beneath the freshly written canonical header. KGX then read `subject`/`object` as endpoint CURIEs and emitted two phantom `biolink:NamedThing` nodes into the merged KG (the non-CURIE node IDs flagged as an ERROR by `kg-model-review`). Header is now dropped by position.

Only `ec_edges.tsv` was affected; the other 13 ontology edge files were clean.

## 3. MediaDive — silent degradation to undated caches

With `data/raw/mediadive/` absent, the transform falls back to the YAML cache under `tmp/medium_yaml` and to `requests_cache` — **neither has an expiry**. A transform launched before `kg download` finishes exits 0 having built the graph from years-old recipes.

That is exactly what produced the shipped output: `data/transformed/mediadive/edges.tsv` is dated Jul 24 15:38 while `data/raw/mediadive/*.json` are Jul 25 00:14, and `tmp/medium_yaml/916.yaml` is dated **Oct 30 2023**. MediaDive has since inserted wrapper solutions, so the output links `solution:1878` straight to `1885` instead of to the wrapper `6570` that now holds it, and labels `1885` "Solution F" rather than "Ferrous chloride solution (5.2%)".

`run()` now refuses this case at the point the stale artifact would be written; `KG_MEDIADIVE_ALLOW_STALE_CACHE=true` opts back in for offline/cache-only work. Construction stays cheap so unrelated tests can still instantiate the transform.

**Important:** the MediaDive recipe divergence is *not* an emission-logic bug — the per-solution edge loop is correct. It is stale input. Re-running the transform against current raw data is what actually repairs that output; this guard only prevents recurrence.

## Testing

Regression tests added for all three, each verified red on the previous code and green on this one:
- `tests/test_ontologies_ec_uri_compaction.py` — raw-line header check (the existing `DictReader` assertions structurally could not see a duplicated header)
- `tests/test_bacdive_lpsn_crossref.py` — strain-CURIE prefix + predicate invariants
- `tests/test_mediadive_stale_cache_guard.py` — new file

`poetry run tox` green: **835 passed**.

## Follow-ups not in this PR

- The merged KG in `data/merged/` (Aug 3) still contains all of the above; it needs a re-merge from freshly re-run transforms.
- `mediadive.py:139` installs `requests_cache` with no expiry, and the YAML medium cache has no invalidation at all. The guard prevents the silent-fallback path but does not add cache expiry.
- `orphan-edges` scans a stray `data/transformed/microbedecoder_bak/` directory as if it were a transform, double-counting results.

### DIFF
--- a/.claude/skills/add-transform/SKILL.md
+++ b/.claude/skills/add-transform/SKILL.md
@@ -166,7 +166,7 @@ For each cross-ref the source provides:
 |---|---|---|---|---|
 | `NCBI:12345` | `NCBITaxon:12345` | yes (ncbitaxon transform) | `biolink:close_match` | trivial |
 | `GTDB s__X` | `gtdb.species:X` | yes (gtdb transform) | `biolink:close_match` | strip s__ per repo convention (see [[feedback_gtdb_prefix]] if noted) |
-| `BacDive 12345` | `bacdive:12345` | yes (bacdive transform) | `biolink:close_match` | this is the merge glue |
+| `BacDive 12345` | `kgmicrobe.strain:bacdive_12345` | yes (bacdive transform) | `biolink:close_match` | this is the merge glue. **Not** `bacdive:12345` — no transform emits that as a node, so edges to it dangle and KGX turns them into empty `biolink:NamedThing` stubs |
 
 Cross-refs that point to entities NOT in KG-Microbe: emit the edge anyway if you're confident downstream analyses will use it. Note it in the design and stub-node it appropriately.
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,6 +186,17 @@ Copy `.env.example` to `.env` and configure:
 - `BACDIVE_USERNAME`: BacDive API email
 - `BACDIVE_PASSWORD`: BacDive API password
 
+Optional, for the mediadive transform:
+- `KG_MEDIADIVE_ALLOW_STALE_CACHE=true`: run mediadive even when the bulk
+  download in `data/raw/mediadive/` is absent or incomplete. Off by default:
+  without the bulk JSONs the transform falls back to the YAML medium cache
+  under `transform_utils/mediadive/tmp/medium_yaml` and to `requests_cache`,
+  neither of which has an expiry, so a run started before `kg download`
+  finishes exits 0 having built the graph from years-old recipes. Set this only
+  for deliberate offline / cache-only work. It must be a **shell** variable —
+  `load_dotenv()` is not guaranteed to have run on the transform path, so an
+  `.env` entry is not reliable here.
+
 Optional, for the ontology transforms:
 - `KG_SEMSQL_BUILD=off`: skip building the SemSQL lookup DBs. Any transform that
   looks something up in an ontology resolves its adapter through
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -260,8 +260,8 @@ class BacDiveTransform(Transform):
         self.assay_edges_generated: Optional[List] = None  # Generated assay→entity edge rows
 
         # LPSN name → record_no index, loaded from data/raw/lpsn_gss.csv when
-        # present. Used to emit bacdive:<id> → biolink:close_match →
-        # lpsn:<record_no> for every strain whose LPSN block names an
+        # present. Used to emit kgmicrobe.strain:bacdive_<id> →
+        # biolink:subclass_of → lpsn:<record_no> for every strain whose LPSN block names an
         # exactly-known LPSN species. Silent skip on missing CSV so a fresh
         # checkout that hasn't downloaded LPSN still runs the BacDive
         # transform cleanly.
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -86,8 +86,6 @@ from kg_microbe.transform_utils.constants import (
     HAS_PHENOTYPE,
     HAS_PHENOTYPE_PREDICATE,
     ID_COLUMN,
-    IN_TAXON_PREDICATE,
-    IN_TAXON_RELATION,
     INFERRED_SUBCLASS_RELATION,
     INTERACTS_WITH_RELATION,
     IS_GROWN_IN,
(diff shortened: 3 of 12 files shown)


realmarcin:
Follow-up findings from the same review pass, filed rather than folded into this PR:

- #681 — mediadive YAML + `requests_cache` have no expiry/invalidation. This PR guards the *silent-fallback* path but does not bound the caches themselves.
- #682 — `kg-path-review` scans stray `data/transformed/microbedecoder_bak/` as a transform (double-counts findings); also notes unexplained cross-transform unresolved-object counts.
- #683 — tracking issue for re-running the affected transforms and re-merging. **The shipped `data/merged/merged-kg.tar.gz` still contains all three defects this PR fixes** — merging this PR does not repair the artifact on its own.

realmarcin:
## Adversarial review pass complete

Reviewed the diff independently (read-only). The three transform fixes each held up — I confirmed empirically that all 62,096 new subjects exist in `bacdive/nodes.tsv`, all 20,420 `lpsn:*` objects exist in `lpsn/nodes.tsv`, both endpoints are `biolink:OrganismTaxon`, and `ec_edges.tsv` has exactly one duplicate header row at line 2.

**Fixed in `9afff23`** (findings that were real):

1. **Both new tests could pass vacuously.** `test_lpsn_edge_uses_subclass_of_not_in_taxon` asserted only that one exact literal was absent, so reverting the subject to the in-scope `bacdive_key` variable (which lacks the `kgmicrobe.strain:` prefix — the most plausible accidental substitution) passed all four assertions. And every mediadive test called `_assert_bulk_data_available()` directly, so deleting the call from `run()` kept the suite green. Both now pinned, and **verified by mutation** — each regression fails exactly one test.
2. **`bulk_data_dir` was CWD-relative.** `kg download -o DIR` + `kg transform -i DIR` would now *hard-fail* telling the user to re-run a download they had already run. Resolved against `input_base_dir` with a repo-anchored fallback (`Transform.DEFAULT_INPUT_DIR` points at a directory that doesn't exist). Both modes verified.
3. **EC header skip guarded on the column name**, so a headerless input re-duplicates the header (caught by tests) rather than silently dropping a real edge.
4. Nits: `-> None`, and the error message now notes the env var must be a *shell* variable since `load_dotenv()` isn't called on the transform path.

**Filed rather than fixed here:**
- #684 — 992 of these edges point at `deprecated=True` LPSN records. Pre-existing, but `subclass_of` puts a live strain under a deprecated class where `in_taxon` did not. Wants a curator decision.
- #685 — a hard failure in any transform aborts all remaining sources; mediadive is 10th of 17, so this now costs hours of completed upstream work. Suggests pre-flight validation.
- #686 — `kg-path-review` needs `lpsn` in `_CROSS_TRANSFORM_SUPPLIED_PREFIXES`; the object half of the orphan finding is a tooling false positive.

**Checked and unfounded** (stating explicitly): `_add_edge_metadata` has no predicate routing table — it returns `OBSERVATION, MANUAL_AGENT` unconditionally, so both predicates get identical metadata. No repo consumer reads `in_taxon`/`RO:0002162` from bacdive. No new `subclass_of` cycle or self-loop is possible (LPSN mints a disjoint `kgmicrobe.strain:DSM-*` namespace, so there's no close_match/subclass_of contradiction). `using_bulk_data` is always initialized before `run()`. `gtdb-phylo-diagram` clade sizing is unaffected.

`poetry run tox` green: **836 passed**.

realmarcin:
## Pre-merge adversarial review — complete

Independent read-only review of the full diff (all commits). **Verdict: no CRITICAL defects; nothing blocks the merge.** Key claims were verified against real artifacts rather than reasoned about abstractly:

- 19,079/19,079 microbedecoder BacDive crosswalk targets resolve to real `kgmicrobe.strain:bacdive_*` node rows — **0 orphaned**.
- 20,420/20,420 `lpsn:*` objects resolve; both endpoints confirmed `biolink:OrganismTaxon` (34,301/34,301 lpsn, 251,404/251,404 bacdive strain nodes).
- The `len(candidate)` slicing rewrite is **provably and empirically** equivalent to the old `split(":",1)[1]` for the four untouched columns — all 27,010 real rows diffed old-vs-new across all five crosswalk columns: **zero divergences** (no crosswalk cell in the source contains a colon at all).
- All new tests mutation-verified as non-vacuous, including the two hypotheses I was least sure about.

**Fixed in `3bdf43a`:**

1. **A real defect I introduced.** The `bulk_data_dir` fallback read the repo's bulk JSONs whenever the given `-i DIR` lacked them — so `kg transform -i /scratch/raw` would set `using_bulk_data=True` against the *wrong* data and the guard would never fire. That reintroduced the exact silent-wrong-data class the guard exists to prevent, inside the guard itself. Now only `DEFAULT_INPUT_DIR` falls back, extracted as `_resolve_bulk_data_dir()` for a real test seam.
2. **`.claude/skills/add-transform/SKILL.md`** still instructed the next transform author to target `bacdive:12345` — the precise bug this PR spent three commits removing.
3. **`KG_MEDIADIVE_ALLOW_STALE_CACHE`** documented in CLAUDE.md alongside its siblings.
4. Restored `bacdive:` to the microbedecoder stub-prefix guard (it had been swapped out rather than added to), and corrected a stale comment describing the old subject/predicate.

**Filed, not fixed here:** #688 (provenance side-effect on 62k edges — likely correct but unpinned), #689 (`merged_graph_stats.yaml` refresh: `biolink:in_taxon` disappears entirely, bacdive was its sole producer), #685 (already open — mid-pipeline abort blast radius).

**Noted, no action:** `IN_TAXON_PREDICATE`/`IN_TAXON_RELATION` in `constants.py` are now unreferenced in production. Left in place as generic constants rather than removed as part of a bug-fix PR.

`poetry run tox` green: **839 passed**.

PR #692, opened 2026-08-06
feat(merge): allow merging a subset of sources with `kg merge -s`

realmarcin:
## Problem

A full `kg merge -y merge.yaml` reached **48 GB resident on a 64 GB machine**, drove free RAM to 59 MB, grew the swapfile from 14 GB to 18 GB, and died after ~40 minutes **without writing anything** — KGX only serializes at the end, so nothing partial survives.

Reading KGX 2.4.2's implementation shows why (`kgx/cli/cli_utils.py:640-661`):

```python
stores = [r.get() for r in results]                        # ALL 20 source graphs in the parent
merged_graph = merge_all_graphs([x.graph for x in stores])
```

Every source's graph is materialized and held simultaneously. Peak scales with the **whole set**, not the largest member — and there was no way to merge in stages.

PREGO dominates: **44,716,161 edges in a 7.38 GB `edges.tsv`**, ~76% of the 9.7 GB total input. Every other source combined is ~2.3 GB.

## What this adds

`kg merge -y merge.yaml -s bacdive -s mediadive` passes the subset through to KGX's existing `source=` parameter, enabling a staged merge so PREGO is never resident alongside the other nineteen sources.

An unknown key fails immediately and names the valid options:

```
$ poetry run kg merge -y merge.yaml -s notasource
KeyError: "Source(s) ['notasource'] not in merge.yaml. Available: ['bacdive', 'bactotraits', ...]"
```

KGX indexes the config dict directly, so a typo would otherwise surface as a bare `KeyError: 'foo'` only once an hour-long run was underway.

`processes` stays at **1** and is now documented as memory-scaling — each concurrent process holds its own source graph, so raising it works directly against the problem this option exists to solve. A test pins the default.

## Investigated and rejected — recorded so nobody re-derives them

| approach | verdict |
|---|---|
| `checkpoint: true` | **No.** `cli_utils.py:787-790` just calls `transformer.save()` per source. Adds I/O, frees nothing. |
| Streaming merge | **Not available.** `stream` exists only on `transform()`; `merge()` has no streaming path in 2.4.2. |
| `merge_all_graphs` tuning | **Already optimal.** It merges into the largest graph by design (`graph_merge.py:17-20`). |
| Filter PREGO by confidence | **Correction: not weak — but not expressible here.** `prego_score` runs 0 to **4.0074** (mean 2.92), not 0–1, so a ≥4.0 cutoff drops **44.66% (20.0M edges)**. It must be **per-channel** (53% of edges are continuously scored, ~46% carry a flat integer 3/4) and applied at transform time, since `merge.yaml` cannot express numeric comparisons. See #693. |
| Filter by `prego_direct_flag` | **No.** 100% `TRUE` — dead column. |
| Dedup PREGO | **No.** 40,604 distinct subjects × 32,440 distinct objects at 3.4% density; the edges are genuinely distinct. |

`edge_filters` *are* applied at read time (`tsv_source.py:275` → `check_edge_filter`), so filtered edges never enter the graph — that mechanism is real and useful. But `check_edge_filter` (`source/source.py:102-134`) does exact-string and list-membership matching only, **no numeric comparison**, so no score threshold is expressible in `merge.yaml`. Score-based reduction has to happen at transform time.

## Testing

Five tests pin the plumbing rather than the flag parsing — that the subset actually reaches KGX, that omitting it still merges everything, that click's empty tuple isn't misread as "merge nothing", that a bad key aborts before KGX runs, and that `processes` stays 1.

`poetry run tox` green: **844 passed**.

## Follow-ups

- The staged-merge workflow needs a stage-2 config (stage-1 output + PREGO as two sources). Not included here; this PR is the enabling change.
- For a real reduction rather than a workaround, PREGO would need either a categorical confidence band emitted by the transform (making `edge_filters` usable), or a curation decision about whether all 44.7M text-mined associations — ~1,101 per taxon — belong in the KG.

### DIFF
--- a/kg_microbe/merge_utils/merge_kg.py
+++ b/kg_microbe/merge_utils/merge_kg.py
@@ -70,16 +70,52 @@ def parse_load_config(yaml_file: str) -> Dict:
     return config
 
 
-def load_and_merge(yaml_file: str, processes: int = 1) -> nx.MultiDiGraph:
+def _assert_sources_exist(yaml_file: str, sources: List[str]) -> None:
+    """
+    Fail before parsing anything if a requested source is not in the config.
+
+    KGX indexes the config dict directly, so an unknown key surfaces as a
+    bare ``KeyError: 'foo'`` only after the run is underway. A merge is a
+    long, expensive operation; a typo should not be discovered an hour in.
+
+    :param yaml_file: Path to the KGX merge config.
+    :param sources: Source keys requested on the command line.
+    :raises KeyError: If any requested source is absent from the config.
+    """
+    with open(yaml_file, "r") as fh:
+        config = yaml.safe_load(fh)
+    available = set((config.get("merged_graph", {}) or {}).get("source", {}) or {})
+    unknown = [s for s in sources if s not in available]
+    if unknown:
+        raise KeyError(f"Source(s) {sorted(unknown)} not in {yaml_file}. Available: {sorted(available)}")
+
+
+def load_and_merge(
+    yaml_file: str,
+    processes: int = 1,
+    sources: Optional[List[str]] = None,
+) -> nx.MultiDiGraph:
     """
     Load and merge sources defined in the config YAML.
 
-    :param yaml_file: A string pointing to a KGX compatible config YAML.
-    :param processes: Number of processes to use.
-    :return: networkx.MultiDiGraph: The merged graph.
+    ``sources`` restricts the merge to a subset of the config's sources.
+    KGX holds every source's graph in the parent process at once —
+    ``kgx.cli.cli_utils.merge`` collects them all before calling
+    ``merge_all_graphs`` — so peak memory scales with the whole set, not
+    with the largest member. Merging in stages is the only way to keep a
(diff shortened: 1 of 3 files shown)


realmarcin:
**Correction to the rejected-approaches table.** I had recorded the confidence filter as "weak, drops only ~13%". That assumed `prego_score` is a 0–1 scale. It is not — full pass over all 44,716,161 edges gives **min 0.0, max 4.0074, mean 2.9166**, so 0.7/0.8/0.9 all sit near the floor.

A ≥4.0 cutoff actually drops **44.66% (20.0M edges)**. Thresholding is a strong lever, but it must be **per-channel** — 53% of edges carry a continuous score (mean 1.96) while ~46% carry a flat integer 3 or 4, so a global cutoff would delete the only channel whose score discriminates and keep the ones where it doesn't. Details and the revised recommendation in #693.

Nothing in this PR's code changes as a result — `merge.yaml` still cannot express a numeric threshold, so staged merging remains the lever available at merge time.

Issue #696, opened 2026-08-06
Design: per-resource confidence calibration for PREGO with a single tunable --min-confidence


realmarcin:
Design issue from a multi-source research pass (103 agents, 21 sources, 104 claims extracted, 25 adversarially verified — 15 confirmed, 10 killed). Depends on #694; blocks the size reduction discussed in #693.

## The premise is confirmed by PREGO's own authors

Not inference. From Zafeiropoulos et al. 2022 §2.3 (verified 3-0 against PMC8879827 / Europe PMC XML):

> "All associations extracted from these resources were assigned **arbitrarily a confidence level of four out of five**."

Appendix C.1:

> "The Genome Annotation and Isolates channel has **fixed values of scores** depending on the resource."

BioProject/PMID rows get 3/5. **PREGO computes no cross-channel combined score** — the shared (0,5] range is a display convention, and STRING-style integration is listed as unimplemented future work. A global cutoff on `prego_score` thresholds across three mathematically unrelated generating functions.

Measured consequence: a `>=4.0` cut retains 55.34% of edges, of which ~85% are flat-channel rows carrying no ordering, while deleting ~87% of the only channel whose score varies.

## Recommended design

Keep PREGO's (0,4] star semantics as the single user-facing axis.

1. **Continuous channel (Environmental Samples, 53.00% of edges)** — replace the raw score with a per-resource empirical percentile remap:

   `star = 4 × F_r(score)`, where `F_r` = empirical CDF within resource *r*.

   Per-resource because the channel aggregates MGnify, MG-RAST metagenome, and MG-RAST amplicon, each with its own resource-local marginal. Both `MGnify` and `MG-RAST metagenome study` are confirmed present in `data/raw/prego/environmental_samples_extracted/`.

2. **Flat channels (47.00%)** — leave at their author-assigned constants as explicit tiers. Do not score-rank them; they have zero within-channel variance.

3. **One knob:** `--min-confidence TAU` in [0,4]; keep edge iff `star >= TAU`. Refuse `TAU > 4.0`.

### Retention (reconciles with measured edge counts)

`retain(TAU) = flat_channels_with_constant >= TAU  +  0.53 × (1 − TAU/4)`

| TAU | retained |
|---|---:|
| 1.0 | 86.7% |
| 1.5 | 80.1% |
| **2.0** | **73.5%** ← recommended default, inside the ≤75% budget |
| 2.5 | 66.8% |
| 3.0 | 60.2% |
[...]

Issue #693, opened 2026-08-06
KGX merge needs ~48 GB for 9.7 GB of input; PREGO is 76% of it


realmarcin:
Analysis of why the full merge cannot run on a 64 GB machine, with what was tried. Filed so the negative results are recorded and nobody re-derives them.

## Observed

A full `kg merge -y merge.yaml`:

- reached **48 GB resident** (only ~5.5 GB of that actually in RAM; ~42 GB compressed or swapped)
- drove free RAM to **59 MB**
- caused macOS to grow the swapfile from **14 GB → 18 GB** while running
- died after ~40 min **without writing anything**

That last point matters: KGX serializes only at the end, so a failed merge leaves no partial artifact. All the work is lost.

## Root cause (KGX 2.4.2, `kgx/cli/cli_utils.py:640-661`)

```python
stores = [r.get() for r in results]                        # ALL source graphs, in the parent
merged_graph = merge_all_graphs([x.graph for x in stores])
```

Every source's graph is materialized and held **simultaneously**. Peak scales with the whole source set, not the largest member.

## Why PREGO dominates

| | |
|---|---|
| PREGO `edges.tsv` | **7.38 GB / 44,716,161 edges** |
| All 45 other input files combined | ~2.3 GB |
| PREGO share of total input | **~76%** |
| distinct subjects × objects | 40,604 × 32,440 (3.4% density) |
| edges per taxon | ~1,101 |

## What was investigated

**Rejected:**

- **`checkpoint: true`** — `cli_utils.py:787-790` calls `transformer.save()` per source. Adds I/O, frees nothing.
- **Streaming** — `stream` is `transform()`-only. `merge()` has no streaming path in 2.4.2.
- **`merge_all_graphs`** — already merges into the largest graph by design (`graph_merge.py:17-20`). Lever already pulled.
- ~~**Confidence filter** — full pass over all 44.7M edges: **86.8% already score ≥0.7**. A 0.7 cutoff drops only ~13%.~~
  **CORRECTED — see the correction comment below.** `prego_score` is **not** a 0–1 scale (min 0.0, max 4.0074, mean 2.92), so 0.7/0.8/0.9 all sit near the floor. A ≥4.0 cutoff drops **44.66% (20.0M edges)**. Thresholding is a strong lever — but must be **per-channel**, not global, because 53% of edges carry a continuous score (mean 1.96) while ~46% carry a flat integer 3 or 4.
[...]

PR #697, opened 2026-08-06
feat(prego): per-resource confidence calibration + emit the resource column

realmarcin:
Prototype for #696. Depends on nothing; unblocks the size reduction in #693.

## Why a global cutoff is wrong — confirmed by PREGO's authors

Not inference. Zafeiropoulos et al. 2022 §2.3, verified against PMC8879827 / Europe PMC XML:

> "All associations extracted from these resources were assigned **arbitrarily a confidence level of four out of five**."

Appendix C.1: *"The Genome Annotation and Isolates channel has **fixed values of scores** depending on the resource."* BioProject/PMID rows get 3/5. **PREGO computes no cross-channel combined score** — the shared (0,5] range is a display convention, and STRING-style integration is listed as unimplemented future work.

Measured over all 44,716,161 edges, a `>= 4.0` cut retains 55.34%, of which **~85% are flat rows carrying no ordering**, while deleting **~87% of the only channel whose score varies**.

## What this adds

`kg_microbe/transform_utils/prego/calibration.py` — the mechanism used by the same lab for TISSUES and DISEASES, and by STRING for its database channel:

```
continuous:  star = 4 * F_r(score)    # empirical CDF within resource r
flat:        star = <author-assigned constant>
keep iff     star >= tau              # one knob, tau in [0, 4]
```

Per-resource because the continuous channel aggregates MGnify and MG-RAST, whose score marginals differ — a shared CDF would conflate them. Both are confirmed present in `data/raw/prego/environmental_samples_extracted/`.

### Retention

| tau | retained |
|---|---:|
| 1.0 | 86.7% |
| 1.5 | 80.1% |
| **2.0** | **73.5%** ← recommended default |
| 3.0 | 60.2% |
| 4.0 | ~53.9% — provenance-dominated |

tau=2.0 reconciles independently: 47.00% flat + (53.00% × 50%) = 73.50%, and the continuous channel's measured p50 is raw score **1.71**. A test pins this against the measured channel shares.

### Determinism

Cutoffs come from fixed-width binned histograms (width 1e-4 over [0, 4.01]) — exact to bin width, O(1) memory, order-independent. **Not** t-digest or P-square: those are order- and implementation-dependent, so two passes over the same file in different chunk orders can disagree, silently changing which edges ship. A test asserts cutoff equality across two orderings of the same multiset.

Ties are never split — the ~13% of rows piled at the score cap move as a unit, so we never delete some Isolates edges and keep others with identical evidence.

Binning goes to 4.01, not 4.0: the shipped data reaches **4.00735**, above the paper's documented cap.

## Blocker fixed

`prego.py` read the resource column and immediately discarded it:

```python
del source  # column carried in edge_attribute space for now; unused
```

It is now emitted as `prego_source`. Per-resource calibration is impossible without it.

## Not done here — deliberately

Wiring the two-pass filter into the 7.4 GB streaming path and adding the CLI knob. That needs a real run to validate realized retention against the model, and the research's per-resource quantiles came from a **1-in-101 sample of the raw file** (including rows the transform drops), so cutoffs must be recomputed on emitted edges before anything is hard-coded.

## Honest caveat on the design

~~**No verified precedent exists for a single global knob after per-channel calibration.**~~ **CORRECTED — see the recheck comment below.** Both refutations were checked against primary sources and the caveat was over-drawn:

- **STRING does expose a single global knob.** Its API documents `required_score` as "a number between 0 and 1000 (default depends on the network)"; per-channel values are output columns, not filter parameters. Only the *specifics* I had been given were wrong (0–999, default 400).
- **TISSUES 2.0 verifies the calibration half verbatim:** "To make the datasets comparable, we converted the raw expression scores into a common scoring scheme," using "a single, monotonic calibration function for all datasets from all organisms." The paper says nothing about per-channel sliders.

The genuine residual caveat is narrower: TISSUES anchors calibration on fold enrichment against a UniProtKB gold standard, so its scores are *quality*-calibrated. **We have no gold standard, so this percentile remap equalises rank only** — it makes no claim about comparable quality until anchored (see the validation path in #696).

## Testing

27 calibration tests plus one on the transform, pinning behaviour rather than arithmetic: that flat channels are tiered rather than ranked, that ties move as a unit, that cutoffs are order-independent, that an uncalibratable row is **kept** rather than silently dropped, and that the default lands inside the 75% budget.

The new transform test caught a wrong assumption while being written — the fixture's resources are `BioProject`/`JGI IMG`, not what I first assumed — which is the assertion doing its job.

`poetry run tox` green: **867 passed**.

### DIFF
--- a/TODO/2026-08-06_codex_review_prompt_pr697_prego_calibration.md
+++ b/TODO/2026-08-06_codex_review_prompt_pr697_prego_calibration.md
@@ -1,0 +1,218 @@
+# Codex review prompt — PREGO gold-standard analysis + threshold calibration (PR #697)
+
+Generated 2026-08-06 via `/codex-review-kg-microbe`. Hand the fenced block below to
+`Agent(subagent_type="codex:codex-rescue", prompt="<block>")` or paste into `codex exec`.
+
+Extends the standard six-dimension template with a **seventh dimension (statistical
+validity)**, because the substance of this branch is a measurement claim, not just
+code — and the two defects already found in it were both methodological rather than
+ordinary logic bugs.
+
+---
+
+```
+You are an experienced code reviewer performing a focused review of the PREGO
+confidence-calibration and gold-standard-analysis work in KG-Microbe
+(Knowledge-Graph-Hub/kg-microbe), branch `feat/prego-confidence-calibration`
+(PR #697, design issue #696).
+
+## Repo context
+
+KG-Microbe is a Python knowledge-graph construction pipeline with three stages —
+**Download → Transform → Merge** — emitting Biolink-modeled `nodes.tsv` +
+`edges.tsv` per source plus a merged KG.
+
+- CLI: `poetry run kg download | transform | merge` (`kg_microbe/run.py`, Click).
+- Transforms: `kg_microbe/transform_utils/<source>/<source>.py`, subclasses of
+  `Transform` in `transform_utils/transform.py`, registered in `DATA_SOURCES`.
+- Shared conventions: `transform_utils/constants.py` (column names, path
+  constants), `transform_utils/custom_curies.yaml`.
+- Merge uses the KGX library (`kg_microbe/merge_utils/merge_kg.py`).
+
+## What this branch does
+
+PREGO (Zafeiropoulos et al. 2022, Microorganisms 10:293) is a text-mined +
+database-derived taxon↔function/environment/disease resource. Its KG-Microbe
+ingest emits **44,716,161 edges in a 7.38 GB `edges.tsv`** — about 76% of all
+merge input by size, and the reason a full merge exhausts a 64 GB machine.
+
+The branch adds a tunable confidence threshold so the source can be reduced, plus
+the machinery to test whether that threshold actually selects better edges.
+
+Files changed (1,301 insertions):
+
+- `kg_microbe/transform_utils/prego/calibration.py` (new, 327 lines) —
+  per-resource percentile calibration. Continuous-channel rows are remapped
+  `star = 4 * F_r(score)` where `F_r` is the empirical CDF within resource `r`;
+  flat channels are rated by their own score. One knob, `tau in [0,4]`; keep
+  iff `star >= tau`. Cutoffs come from fixed-width binned histograms
+  (`BIN_WIDTH = 1e-4` over `[0, SCORE_MAX=4.01]`).
+- `kg_microbe/transform_utils/prego/quality.py` (new, 235 lines) — fold-enrichment
+  measurement against a gold standard: `fold = P(hit | window) / P(hit)`.
+- `kg_microbe/transform_utils/prego/prego.py` (+183) — two-pass wiring
+  (`_collect_calibration` then the emit pass), `PREGO_MIN_CONFIDENCE` env var,
+  `prego_source` column now emitted (previously `del source`), payload memoization,
+  calibration-table output.
+- `kg_microbe/transform_utils/constants.py` (+5) — `PREGO_SOURCE_COLUMN`.
+- Tests (+555 across three files).
+
+## Domain facts established empirically — treat as given, do not re-derive
+
+- `prego_score` runs **0 to 4.00735** (mean 2.92). The paper documents a cap of 4;
+  the shipped data exceeds it.
+- The score is **not one scale**. Per PREGO's authors (§2.3), the genome-derived
+  channels were "assigned arbitrarily a confidence level of four out of five" and
+  BioProject/PMID rows three of five. Only the Environmental Samples channel has a
+  computed, varying score. PREGO computes **no cross-channel combined score**.
+- Channel composition of emitted edges: continuous (`<N of M samples>`) **53.00%**;
+  flat (Isolates 28.66%, Genome annotation 9.17%, MAG 7.27%, SAG 1.79%,
+  PMID 0.05%) **47.00%**.
+- The continuous channel aggregates three resources with materially different
+  cutoffs at the same tau: MG-RAST metagenome **1.717**, MG-RAST amplicon **0.575**,
+  MGnify **0.848**. MG-RAST metagenome is 99.62% of the channel.
+- **MG-RAST amplicon holds 46.4% of its rows at the score cap**, so every tau above
+  ~2.5 retains that same 46.4%.
+- A real calibration pass sees 23,698,632 continuous rows vs 23,697,432 in the
+  emitted `edges.tsv` (0.005% apart).
+- Two gold standards were built and **they disagree on the direction of the
+  score/quality relationship**:
+  - UniProt (from `data/merged/20250222_function/`, joining
+    `protein -derives_from-> NCBITaxon` with `protein -participates_in|located_in-> GO`):
+    14,411,869 pairs, 4,209 GO terms, 41.5% of PREGO taxon→GO edges comparable.
+    Fold **rises**: 0.94x, 0.96x, 1.04x, 1.19x. Flat channels 2.19x.
+  - metatraits + metatraits_gtdb + madin_etal (trait-derived): 333,407 pairs,
+    78 GO terms, 1.0% comparable. Fold **falls**: 1.61x, 1.59x, 1.56x.
+    Flat channels 1.00x.
+- A GO term's edge count correlates with its mean score at Spearman **+0.26**,
+  driven by rare terms scoring low (0.67 mean in the lowest-ubiquity decile vs
+  ~2.0 above).
+
+## Review scope
+
+Focus on: **kg_microbe/transform_utils/prego/ and tests/test_prego_*.py**
+
+Read `calibration.py`, `quality.py`, and the changed regions of `prego.py` in full.
+Ignore `data/`, `.tox/`, `.venv/`, `__pycache__/`, and generated TSVs.
+
+## Review dimensions — required
+
+Evaluate against **all seven**. Not every file will have findings in every
+dimension; do not skip one.
+
+1. **Code logic** — control flow, off-by-one, wrong operator precedence,
+   mis-ordered arguments, values computed but never reaching a writer.
+   Specifically: is `ScoreHistogram.cutoff()` solving `4*F(s) >= tau` correctly at
+   the boundaries? Is `_bin_index` clamping correct for `score <= 0`, `score` at
+   exactly `SCORE_MAX`, and `score > SCORE_MAX`? Does `star_for_row` return the
+   right thing for each of {continuous with cutoff, continuous without cutoff,
+   recognised flat, unrecognised}?
+
+2. **Consistency** — does the calibration path duplicate logic that already exists
+   elsewhere in the repo? Is `_KEEP_OUTCOMES` genuinely the same predicate the emit
+   pass applies, or can the two diverge? Is `is_continuous_channel` (shape-matching
+   on `"N of M samples"`) duplicated anywhere? Does the env-var convention match
+   `METATRAITS_WORKERS` / `KG_MEDIADIVE_ALLOW_STALE_CACHE`?
+
+3. **Robustness** — malformed rows, missing columns, absent archives, a resource
+   present in pass 2 but absent from pass 1's cutoff table, empty histograms,
+   `float()` on user-supplied text, non-idempotency, `atomic_write` usage for the
+   calibration table.
+
+4. **Bugs** — provable on real inputs. Each MUST include file:line, ≤2-sentence
+   description, minimal repro, expected vs actual.
+
+5. **Bottlenecks** — the two-pass design reads the archives twice. Is
+   `_ensure_payload` memoization actually hit on both passes? Is `tarfile.getmembers()`
+   on an 8.7 GB gzipped tar avoided where possible? Any per-row work that belongs
+   outside the loop?
+
+6. **Scalability** — the emit path streams 44.7M rows and the calibration pass
+   another 44.7M. Flag anything O(N) in memory. `ScoreHistogram` uses a dict keyed
+   by bin index — bound its worst case. Does anything accumulate per-row state that
+   would break past 5M rows? Would this survive being run under the metatraits-style
+   multiprocessing pool?
+
+7. **Statistical validity** *(added for this branch — the substance here is a
+   measurement claim, and both defects found so far were methodological)*
+   - **Tie handling.** `enrichment_by_window` aggregates by exact score so window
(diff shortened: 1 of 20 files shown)


realmarcin:
## Recheck: my "no verified precedent" caveat was wrong

I flagged the single-global-knob design as "a synthesis, not a replicated pattern" on the strength of two 0-3 sub-agent refutations. I have now checked both against primary sources, and the caveat was over-drawn. Correcting it here since it is in the PR body.

### STRING — the precedent does exist

From [STRING's API documentation](https://string-db.org/help/api/), verbatim:

> `required_score` — "threshold of significance to include an interaction, a number between **0 and 1000** (default depends on the network)"

So STRING **does** expose exactly one global threshold across a multi-channel resource. Per-channel values (`nscore`, `fscore`, `pscore`, `ascore`, `escore`, `dscore`, `tscore`) are **output columns, not filter parameters**.

What the refutation correctly killed was only the *specifics* I had been given: the range is 0–1000, not 0–999, and there is **no fixed default of 400** — the default is network-dependent. I wrongly generalised "those numbers are wrong" into "the precedent does not exist."

### TISSUES — the calibration half is verified verbatim

From [TISSUES 2.0](https://pmc.ncbi.nlm.nih.gov/articles/PMC5808782/) (Palasca et al. 2018, PMC5808782):

> "**To make the datasets comparable, we converted the raw expression scores into a common scoring scheme.**"

> "a **single, monotonic calibration function** for all datasets from all organisms"

And the calibration protocol, confirming the method cited in #696:

> "To calculate the fold enrichment in a given dataset, we select the genes and tissues that are in common between the dataset and the gold standard, sort the gene–tissue pairs by raw expression value and calculate fold enrichment in **sliding windows of 100 pairs**."

That is precisely the architecture this PR implements: per-source monotone recalibration onto one shared axis so a single knob is meaningful.

Note the TISSUES 2.0 paper says **nothing** about per-channel sliders — the counter-claim that killed the original finding is not supported by the paper. It may be true of the current web UI, but it was never verified, and in any case a resource can both calibrate to a common scale *and* offer per-channel filtering; those are not mutually exclusive.

### What genuinely remains unverified

The **strong** form: that after calibration a cutoff means *equal estimated quality* in every channel. TISSUES describes scores as reflecting each dataset's quality **relative to a gold standard**, and explicitly notes dataset quality varies. So "comparable" is verified; "equal quality by construction" is not.

That distinction matters for us because we have **no gold standard yet**. TISSUES anchors its calibration on fold enrichment against UniProtKB; our percentile remap has no such anchor, so it equalises *rank* without any claim about quality. The validation path in #696 (fold enrichment against UniProt GOA / GTDB-derived GO annotations) is what would close that gap.

### Net effect

The design is better supported than I said, not worse. STRING is a genuine single-knob precedent; TISSUES is a genuine per-source-recalibration precedent. The honest residual caveat is narrower: **our remap is rank-equalising, not quality-calibrated, until it is anchored to a gold standard.**

realmarcin:
## Two-pass filter wired, and validated against the real archives

Per the concern that committing a cutoff table off a **1-in-101 sample of the raw file** would be exactly the silent wrongness this thread has been about — calibration now runs **inside the transform**, over the rows the run will actually emit. Pass 1 and pass 2 share `_KEEP_OUTCOMES`, so they cannot disagree about which rows count. No table is committed.

Pass 1 is skipped entirely at the default threshold, so the unfiltered path stays single-pass and existing runs are byte-identical.

### Real calibration pass (14 GB of raw archives, ~8 min)

| resource | rows | share |
|---|---:|---:|
| MG-RAST metagenome study | 23,609,263 | 99.62% |
| MG-RAST amplicon study | 56,843 | 0.24% |
| MGnify | 32,526 | 0.14% |

**Three independent cross-checks pass:**
- Pass 1 sees **23,698,632** continuous rows; the emitted `edges.tsv` has **23,697,432** — 0.005% apart. It is calibrating on the population that ships.
- MG-RAST metagenome's cutoff at min-confidence 2.0 is **1.717**; the independently measured channel p50 is **1.71**.
- Retention at 2.0 = 47.00% flat + (53.00% × 50.0%) = **73.50%**, matching the model.

**Per-resource calibration is vindicated.** At the same threshold the three resources need cutoffs of **1.717 / 0.575 / 0.848**. A shared cutoff would have cut MGnify and the amplicon study far harder than intended. Worth noting for expectations though: MG-RAST metagenome is 99.62% of the channel, so per-resource matters for the *correctness of the small resources*, not for the bulk retention number.

### Defect the real run found

**MG-RAST amplicon holds 46.4% of its rows at the score cap**, so every threshold above ~2.5 retains that same 46.4% rather than the requested fraction:

```
min-confidence 2.5   cutoff 4.000   keeps 46.4%   (37.5% requested)
min-confidence 3.0   cutoff 4.000   keeps 46.4%   (25.0% requested)
```

Ties are deliberately never split, so this is correct behaviour — but the knob silently stops responding for that resource. A filtered run now **warns per resource** when realized retention exceeds the request by more than 5 points, with a regression test.

### Defect the fixture canary found, before any of this touched real data

Flat channels were being rated by their channel's *documented constant*, so an `Isolates` row scoring 3 — which the fixture has — was silently **promoted** to 4. Flat-channel scores are already on the star axis, so the row's own value is now used, and a row disagreeing with its channel's tier is preserved as the data-quality signal it is. That surfaced a follow-on question: an *unrecognised* channel must not be thresholded on semantics we have not verified, so it is now declined a rating and kept.

### Also

Memoized payload resolution — listing members of a gzipped tar decompresses the whole archive (minutes for the 8.7 GB isolates file), and a filtered run resolves each payload twice.

Configured via `PREGO_MIN_CONFIDENCE` (matching the repo's env-var idiom) or the constructor arg. Out-of-range values are refused. A filtered run writes `confidence_calibration.tsv` next to the outputs.

`poetry run tox` green: **872 passed**.

realmarcin:
## Quality anchor added — with a result that changes how this PR should be described

`prego/quality.py` implements fold enrichment against a gold standard (the TISSUES protocol). Measured against 333,407 curated taxon→GO pairs from `metatraits` + `metatraits_gtdb` + `madin_etal`, **PREGO's score is anti-correlated with agreement**:

| continuous-channel quintile | fold |
|---|---:|
| 0.001–0.761 | **1.61x** |
| 0.761–1.640 | 1.59x |
| 1.640–2.775 | 1.56x |
| 2.775–4.000 | 1.34x |
| 4.000–4.007 | **1.17x** |

Monotonically decreasing, and it survives stratifying by GO term (30 of 47 terms; pooled 0.180 low-half vs 0.151 high-half). The flat channels — which carry the top star tier — sit at **1.00x, exactly random**, while the continuous channel reaches 1.45x.

### What this means for this PR

The mechanism is sound and stays. What changes is the claim:

- **Valid:** `--min-confidence` as a *size* lever. Retention is predictable and validated (73.50% at 2.0), which is what #693 needed.
- **Not valid:** describing it as a quality filter. Raising it selects edges that agree with curated knowledge *less* often.

I have not re-scoped the code, because the filter still does exactly what it says — it thresholds a documented axis deterministically. But it should not be merged with a description that implies quality improvement, and the module docstrings now say so explicitly.

### Not shipped

No quality calibration. Only ~1% of PREGO taxon→GO edges are comparable over 70 shared GO terms (0.24% of its GO space) — fitting a calibration on that and applying it to 44.7M edges would be the exact overreach this work was meant to avoid.

Full detail, caveats, and next steps in #696.

`poetry run tox` green: **885 passed**.

realmarcin:
## Correction — the quality finding reverses against a broader gold standard

The previous comment reported PREGO's score as anti-correlated with agreement. Built a UniProt gold standard from `kg-microbe-function` (14,411,869 taxon→GO pairs, 4,209 GO terms, **41.5%** of PREGO edges comparable vs 1.0% before) and it **rises** instead: 0.94x → 0.96x → 1.04x → 1.19x across the continuous channel.

Cause of the disagreement: provenance alignment. UniProt is genome-derived and PREGO's flat channels are genome-derived, so each gold standard flatters the channel sharing its provenance. Neither is neutral.

Also fixed a real bug in `enrichment_by_window`: it sliced by index, splitting tie blocks so that sort tiebreaks decided which side rows fell on. That fabricated the original 0.44x/1.95x adjacent windows. Windows now break only on score changes, with regression tests.

**Net effect on this PR:** the earlier "do not describe this as a quality lever" caveat softens to "the quality effect is weak (~1.0–1.2x) and its direction is gold-standard dependent." The size-lever justification (#693) is unaffected. The consistent, decision-relevant property is that raising the threshold strips rare, specific annotations first (Spearman +0.26 between a GO term's edge count and its mean score).

Full detail in #696. `poetry run tox` green: **888 passed**.

realmarcin:
## Codex review complete — 11 findings, 4 fixed here

Ran the review prompt from `TODO/2026-08-06_codex_review_prompt_pr697_prego_calibration.md` (which extends the standard six dimensions with a seventh, statistical validity). Result: **0 CRITICAL, 3 HIGH, 7 MEDIUM, 1 LOW** — and **5 of 11 findings are statistical**, including all three HIGHs. Adding that dimension was worth it.

### Fixed in this PR, all verified by execution

1. **`is_monotone_increasing` was vacuously true.** `all()` over an empty sequence returns True, so a fully tied channel — no score ordering whatsoever — reported as monotonically increasing. Now requires >=2 usable windows.
2. **Non-finite scores were inconsistent across passes.** `_bin_index(inf) == 0` shaped the bottom histogram bin, yet `inf >= cutoff` survived every threshold. Bottom-ranked in both now.
3. **The two passes could disagree.** My comment saying they "can never disagree" was false — the emit pass also rejects empty IDs, GO-prefix mismatches, and DOIDs without a MONDO xref. Cheap checks mirrored; residual gap now stated honestly. This reframes the 1,200-row delta I had reported as reassuring: it is a symptom, not confirmation.
4. **`run()` twice produced an incomplete `nodes.tsv`.** The node de-dup set persisted while the file was truncated. State now resets at the top of `run()`, before `_load_dictionary` populates counters. Mutation-verified.

### Docstring overclaims corrected

The review quoted six unsupported statements. The substantive ones: "Fold 1.0 means the score carries no information" (only relative to a uniform-cell baseline); "The reversal is a provenance-alignment artifact" (a hypothesis, now labelled as one); "spans roughly 1.0-1.2x" (wrong — that pooled two benchmarks; now stated separately as UniProt 0.94-1.19x and trait-derived 1.56-1.61x); and "byte-identical" (false, since this branch adds `prego_source`).

### One place the review is mistaken

It flagged the "0.00x and 5.00x tied windows" docstring as unsupported. It is supported — the metatraits flat-channel run produced exactly those figures. Codex only had the 0.44x/1.95x example I put in the prompt, so it was a fair challenge on the evidence given, but the claim stands and I have left it.

### Filed, not fixed here

- #698 — the three HIGH statistical findings: Cartesian baseline null, UniProt circularity for genome-derived channels, missing clustered uncertainty.
- #699 — unify the row-eligibility predicate across the two passes (structural fix).
- #700, #701 — tar-scan cost on cached payloads; O(N) memory in `enrichment_by_window`.

### On merging

The review's Q1 matches my read: **safe as an opt-in size lever** — default no-op, bounded tau, deterministic inversion, ties intact, bounded histogram memory, realized retention recorded. Not an exact size target, because flat tiers, cap plateaus and uncalibratable rows create discontinuities; the calibration table and actual emitted count are authoritative.

No quality claim should ride on this until at least #698's baseline and uncertainty items are done.

`poetry run tox` green: **891 passed**.

realmarcin:
## Adversarial review complete — four blockers found, all fixed

Rebased cleanly onto master (#691, #692, #704 are ancestors; the reviewer independently verified the merge-base is master's tip and nothing was reverted). Then an adversarial review found four blocking defects, **three of them only visible by running the code**.

### 1. CRITICAL — the calibration table was never written

`atomic_io.atomic_write` is a `@contextmanager`. It was called as a plain function, so the returned generator was never entered: no temp file, no `os.replace`, no output — while the run printed `"wrote calibration table"`. The second positional argument also bound to `mode`. All nine other call sites in the repo use `with`.

A filtered run therefore shipped a reduced edge set with **no record of what was dropped**, plus console confirmation that the record existed. My earlier report that "a filtered run writes confidence_calibration.tsv next to the outputs" was wrong.

### 2. HIGH — the table and the filter measured retention differently

`as_row` used histogram bin membership; `star_for_row` used the raw score. Not interchangeable: `int(score/1e-4)*1e-4` exceeds `score` for **11.5%** of representable 4-dp values — including **1.71**, the measured p50 of the real continuous channel. On a probe the table reported 0.600 kept while 0.200 shipped. Both now compare against `cutoff_bin`.

### 3. HIGH — the threshold ran before eligibility checks

Rows that would have been dropped anyway were attributed to `below_min_confidence`: **418 reported vs 198 actually lost** on a probe, and `unmapped_associations.tsv` lost every other drop reason at any non-default tau — the file whose stated purpose is giving curators fix targets. Moved to immediately before the write.

### 4. HIGH — withdrawn claims shipped inside the measurement module

`quality.py`'s docstring still asserted 1.57x/18-of-19, 1.94x/7-of-9, and "two independent edge types agree" — all three listed as corrected in `PREGO_SCORE_VALIDATION.md`. Replaced with the current status (**NOT ESTABLISHED**) and a pointer to the record.

### The root cause of the test gap

The fixture had **zero continuous-channel rows**, so pass 1 always returned an empty histogram and the whole histogram → cutoff → filter path was dead in every test. Four mutations survived: stubbing `_collect_calibration`, no-oping the table write, forcing `kept_fraction` to 1.0, and coarsening `BIN_WIDTH` 100×.

Added 12 continuous rows spanning 0.20–4.00 with a tie block at 1.71, plus a test asserting the table exists and its `kept_fraction` matches the emitted edge count. **All four mutations now fail.** Also replaced the default-is-a-no-op assertion — it only checked `min_confidence == 0.0` and would have passed if the default dropped 90% of edges.

### Filed, not fixed

#708 (the bootstrap discards multiplicity — a ~63% subsample, not *n*-out-of-*n*; conservative, so the verdict survives), #709 (`PREGO_MIN_CONFIDENCE` undocumented, and the docs argue against a single global knob), #710 (`quality.py` unused by the package, plus five smaller inconsistencies).

### What this PR is justified as

An opt-in, default-off per-resource percentile threshold as a **size** lever on a 44.7M-edge / 7.38 GiB source — bounded, auditable data reduction. **Not** as evidence that it improves edge quality; the score's quality signal is not established. `merge.noprego.yaml` (#691) remains the coarser alternative that needs none of this machinery.

`poetry run tox` green: **909 passed**.

Issue #695, opened 2026-08-06
prego edges have empty knowledge_level and agent_type


realmarcin:
Found while inspecting PREGO edge attributes for #693.

## Problem

Every edge in `data/transformed/prego/edges.tsv` has **empty** `knowledge_level` (column 6) and `agent_type` (column 7):

```
NCBITaxon:100  biolink:capable_of  GO:0000034  RO:0002215  infores:prego  <empty>  <empty>  4  Isolates  TRUE  https://img.jgi.doe.gov/...
```

Other transforms populate these. BacDive, for example, sets them per-edge via `_add_edge_metadata`, and the ontologies transform sets `knowledge_assertion` / `manual_agent` for every row.

## Why it matters

PREGO is **text-mined** (Zafeiropoulos et al. 2022 — a JensenLab tagger pipeline over PubMed co-mentions). It is exactly the source where these fields carry the most signal, because its edges are the least like a curated assertion. Biolink's intended values here would be something on the order of `knowledge_level: prediction` (or `statistical_association`) and `agent_type: text_mining_agent` / `automated_agent`.

Leaving them blank means a consumer cannot distinguish 44.7M text-mined co-mention associations from manually curated assertions anywhere in the merged KG — PREGO contributes the single largest block of edges in the graph.

## Suggested fix

Set both in the prego transform. The values are constant per source, so this is a small change. Worth confirming the exact biolink enum values against the current model rather than guessing:

```bash
poetry run python -c "
from bmt import Toolkit; t=Toolkit()
print(t.get_element('knowledge level').permissible_values.keys())
print(t.get_element('agent type').permissible_values.keys())
"
```

Also worth a general check for other transforms leaving these blank — `kg-model-review` does not currently assert their presence, which is why this went unnoticed.

realmarcin:
Closing: fixed in code by `9b1c48c06` (#703) and now in the graph. Rebuilt `prego_habitat/edges.tsv` (FRESH, 2026-09-08): `knowledge_level`/`agent_type` are `statistical_association`/`data_analysis_pipeline` on 193,335 edges, `knowledge_assertion`/`automated_agent` on 37,749, `prediction`/`text_mining_agent` on 5,360 — 0 empty of 237,745.

Issue #686, opened 2026-08-06
kg-path-review: add 'lpsn' to _CROSS_TRANSFORM_SUPPLIED_PREFIXES


realmarcin:
Raised by the adversarial review of #680.

## Problem

`_CROSS_TRANSFORM_SUPPLIED_PREFIXES` in `.claude/skills/kg-path-review/kg_path_review.py:769-775` lists `NCBITaxon`, `GTDB`, `GenBank`, and others, but **not `lpsn`**.

Consequence: the `orphan-edges` archetype reported both halves of the bacdive LPSN cross-ref as orphaned —

```
bacdive_orphan_subject = 62096
bacdive_orphan_object  = 62096
```

#680 fixes the subject half (the subject CURIE was genuinely never emitted as a node). The object half is a **tooling false positive**: all 20,420 `lpsn:*` objects resolve in `data/transformed/lpsn/nodes.tsv`, and `merge.yaml:193` includes the `lpsn` source, so the merged KG is clean. Without this change the next `orphan-edges` run will still emit 62,096 CRITICAL findings that are not real.

## Suggested fix

Add `"lpsn"` to the frozenset.

## Caveat worth noting in the skill

`merge.minimal.yaml` has **no** `lpsn` source, so under that config the objects genuinely do dangle. That was already true before #680 (both endpoints dangled), so it is a strict improvement rather than a regression — but if the skill ever grows a merge-config-aware mode, this is a case where the answer differs by config.

realmarcin:
Closing: fixed in `9b1c48c06` (#703) — `lpsn` is in `_CROSS_TRANSFORM_SUPPLIED_PREFIXES` (`kg_path_review.py:824-830`, with the comment that the lpsn transform owns those ids and merge.yaml includes it).

Issue #682, opened 2026-08-06
kg-path-review: stray data/transformed/microbedecoder_bak/ is scanned as a transform


realmarcin:
Found during a `kg-path-review` sweep.

## Problem

`data/transformed/microbedecoder_bak/` is a backup directory, but the `orphan-edges` archetype (and any other aggregate archetype that enumerates transform dirs) treats it as a live transform. Its stats are reported alongside the real one and are byte-identical:

```
microbedecoder_nodes = 5224
microbedecoder_edges_scanned = 514641
microbedecoder_orphan_subject = 514641
...
microbedecoder_bak_nodes = 5224
microbedecoder_bak_edges_scanned = 514641
microbedecoder_bak_orphan_subject = 514641
```

Every count from that source is double-reported, which inflates aggregate totals and makes the finding list misleading.

## Notes

The skill already documents this hazard for merge snapshots — see the "Operational gotchas" section and `NON_TRANSFORM_DIR_PREFIXES` / `_list_transform_dirs()` in `.claude/skills/kg-path-review/kg_path_review.py` — but a `_bak` suffix is not covered by the existing prefix filter.

## Suggested fix

Either add a `_bak` / `.bak` suffix filter to `NON_TRANSFORM_DIR_PREFIXES` handling, or prune the directory. A suffix filter is preferable since backups will recur.

## Separate observation (may warrant its own issue)

The same run reported large cross-transform unresolved-object counts that are worth a look independently:

- `rhea_mappings_cross_transform_orphan_obj = 180524`
- `microbedecoder_cross_transform_orphan_obj = 39123`
- `bacdive_cross_transform_orphan_subj = 59385`, `..._obj = 21834`
- `metatraits_gtdb_cross_transform_orphan_obj = 6600`

`microbedecoder` emitting edges only against `lpsn` nodes is documented and expected (see CLAUDE.md), but the object-side counts above are not obviously accounted for. Worth confirming whether these resolve at merge time via the ontologies transforms or are genuine dangling references.

realmarcin:
Closing: fixed in `9b1c48c06` (#703) — `kg_path_review.py:52` skips `_bak`, `.bak`, `_old`, `_backup`, `_orig` directories — and the directory itself is gone (`data/transformed/microbedecoder_bak/` no longer exists).

PR #703, opened 2026-08-07
fix(prego,path-review): real channels, populated edge metadata, fewer false orphans

### LINKED ISSUE #694: prego_channel is not a channel: it holds sample counts, PMIDs, sequencing strategies and body sites
Found while evaluating whether PREGO edges can be filtered (#693).

## Problem

`data/transformed/prego/edges.tsv` column 9 is named `prego_channel`, and `docs/PREGO_INGEST_PLAN.md:115` makes it a first-class filter axis:

> **Threshold policy for v1:** emit all associations from all three channels. Downstream consumers can filter by `score`, `channel`, or `direct_flag`.

But the column does not hold a channel. Normalizing its values across all 44,716,161 edges:

| value class | share | example |
|---|---:|---|
| `<N of M samples>` | **53.00%** | `402 of 487 samples` |
| `Isolates` | 28.66% | |
| `Genome annotation` | 9.17% | |
| `Metagenome-Assembled` | 7.27% | sequencing strategy |
| `Single Amplified Genome` | 1.79% | sequencing strategy |
| `<PMID:...>` | 0.05% | `PMID:24914180` |
| body sites / habitats | trace | `Kidneys`, `Large intestine`, `Brain`, `Stomach`, `Nodule`, `Plants`, `Pulp and paper waste` |

So the column mixes at least five distinct kinds of thing: an evidence tally, a provenance class, a sequencing strategy, a citation, and an anatomical/environmental context. The sample-count strings are effectively unique per row, so the column has huge cardinality and cannot be selected on.

The three channels the plan describes (literature / environmental / isolates) are not recoverable from it.

## Why it matters

1. **The documented v1 filter policy does not work.** `channel`-select is one of the two escape hatches `PREGO_INGEST_PLAN.md:115` names for when edge volume becomes a problem — and volume *is* now the problem (#693: PREGO is 76% of merge input, 44.7M edges).
2. **It blocks the fix for #693.** The revised recommendation there is per-channel score thresholds, because score semantics differ sharply by channel (53% continuous, mean 1.96; ~46% flat integer 3 or 4). That cannot be implemented while the channel column is unusable.
3. **It is misleading in the shipped KG.** Anyone consuming `prego_channel` will assume it names a channel.

## Suggested fix

Decide what the column is meant to carry and split accordingly. Likely:

- `prego_channel` — the actual channel (literature / environmental / isolates), derived from which source archive the row came from rather than from this field.
- a separate column for evidence tallies (`402 of 487 samples`), ideally parsed into two numeric columns.
- citations belong in a publication/`has_evidence` field, not here.
- body sites and habitats look like they should be ontology-mapped (UBERON / ENVO) rather than free text — worth checking whether they are being dropped on the floor today.

Verify against the raw archives which field the transform is reading; the mixing may originate upstream, in which case the transform should normalize rather than pass through.

## Reproduce

```bash
awk -F'\t' 'NR>1{ch=$9;
  if(ch ~ /^[0-9]+ of [0-9]+ samples$/) ch="<N of M samples>";
  else if(ch ~ /^PMID:/) ch="<PMID:...>";
  c[ch]++; n++} END{for(k in c) printf "%d\t%s\t%.2f%%\n", c[k], k, 100*c[k]/n}' \
  data/transformed/prego/edges.tsv | sort -rn | head -20
```

realmarcin:
Closes #694, #695, #686, #682.

Four issues across two areas. No shared cause beyond all four making the graph — or the tools that review it — describe themselves inaccurately.

## #694 — `prego_channel` was not a channel

It carried PREGO's column 6 verbatim. Across the real archives that column holds **~24k distinct values** of at least four different kinds:

| kind | examples | count in the isolates archive |
|---|---|---:|
| resource class | `Isolates`, `Genome annotation`, `Metagenome-Assembled Genome`, `Single Amplified Genome` (each also `* GOLD`) | 41.9M |
| evidence tally | `402 of 487 samples` | all of the environmental archive |
| citation | `PMID:24914180` | 30,931 |
| habitat | `Aquatic`, `Groundwater`, `Plants`, `Terrestrial` | ~13k |

That made **channel-selection impossible** — one of the two filters `PREGO_INGEST_PLAN.md:115` promises for when edge volume becomes a problem. It also silently shaped every score-validation measurement in `docs/PREGO_SCORE_VALIDATION.md`, which had to reconstruct "channel" by pattern-matching the tally shape.

**Fix.** The channel is a property of the *archive*, which is also what PREGO's paper means by one:

- `prego_channel` — derived from the archive: `environmental_samples` / `annotated_genomes_isolates` / `literature`
- `prego_evidence` — column 6 preserved verbatim, named for what it actually holds
- `prego_evidence_class` — `sample_count` / `resource_class` / `publication` / `habitat`, so the grab-bag is filterable

## #695 — edges shipped with empty `knowledge_level` and `agent_type`

44.7M text-mined and statistically-derived associations were indistinguishable from curated assertions anywhere in the merged KG — and PREGO is the single largest edge block in it.

Set per channel, because the channels are generated by genuinely different processes and one constant would misdescribe most of them:

| channel | knowledge_level | agent_type |
|---|---|---|
| `environmental_samples` | `statistical_association` | `data_analysis_pipeline` |
| `annotated_genomes_isolates` | `knowledge_assertion` | `automated_agent` |
| `literature`, or any row with a citation | `prediction` | `text_mining_agent` |
| unrecognised | `not_provided` | `not_provided` |

A citation overrides the channel default: those rows combine curated metadata with text mining over the linked abstract, which PREGO's authors themselves score one tier lower. All values verified against biolink's `KnowledgeLevelEnum` / `AgentTypeEnum`.

## #686 — 62,096 false CRITICAL orphans in every path review

`lpsn` was missing from `_CROSS_TRANSFORM_SUPPLIED_PREFIXES`, even though the lpsn transform owns those nodes and `merge.yaml` includes it. Verified against the re-run bacdive output: the orphan-edges archetype now returns **0 CRITICAL** where it previously reported 40 findings covering 62,096 edges.

Noted in the code that `merge.minimal.yaml` has no lpsn source, so the frozenset is not merge-config aware.

## #682 — backup directories scanned as transforms

`data/transformed/microbedecoder_bak/` carries a full `edges.tsv`, so every aggregate archetype counted it twice. Filtered by **suffix** rather than exact name, since these vary (`_bak`, `.bak`, `_old`, `_backup`, `_orig`). Transform-dir count drops from 14 to 13.

## Note for #697

That branch adds `calibration.py` / `quality.py`, whose `is_continuous_channel` matches the `"N of M samples"` shape against `prego_channel`. After this change that shape lives in `prego_evidence`, and the channel is directly available — so the rebase should **simplify** that code rather than complicate it.

## Testing

11 new tests covering archive→channel derivation, evidence classification, per-channel metadata, the unrecognised-channel fallback, and an end-to-end assertion that no emitted edge has empty metadata.

`poetry run tox` green: **850 passed**.


realmarcin:
## Rebased onto master (now includes #697)

Force-pushed. `mergeStateStatus` is back to `MERGEABLE`.

The textual conflicts were all additive and confined to `prego.py` (4) and `tests/test_prego_transform.py` (1) — resolved keep-both. But the merge being *textually* clean hid a real semantic defect, which is the part worth reviewing:

### The rebase broke calibration, and neither PR was wrong on its own

#703 moves the channel from a data column to the archive filename. #697 added a calibration pass that reads the channel. After the rebase they disagreed on what "channel" means:

- `_process_archive` derived it from the archive name (#703 semantics).
- `_collect_calibration` still read it from `row[5]` (#697 semantics) — which under #703 holds the **evidence** string, not the channel.

So `is_continuous_channel` was false for every row in the calibration pass. The histogram came back empty, no cutoff was computed, and no calibration table was written — while the emit pass starred rows against a cutoff that did not exist. This is the same producer/measurer split that `cutoff_bin` was introduced to close in #697, reopened by the *combination* rather than by either change.

Fixed: `_collect_calibration` derives the channel from the archive name and skips non-continuous archives outside the row loop.

### The test suite could not have caught it

The fixture wrapped every row in a single `literature.tar.gz`. Once the channel came from the archive name, there was no continuous-channel row anywhere in the suite — the histogram → cutoff → filter path was dead in every test, which is the same gap that let four mutations survive in #697.

Split into two archives mirroring production: `literature.tar.gz` (flat) and `environmental_samples.tar.gz` (the 12 continuous rows). Verified by mutation — pinning the calibration channel to a constant fails 2 tests.

### Also in the rebase

- `is_continuous_channel` / `flat_channel_star` re-keyed off the channel constants instead of score-shape matching.
- Fixed a stale assertion identifying continuous edges by `" of " in prego_channel`; that raw string moved to `prego_evidence`.

`poetry run tox` green: **915 passed**.

realmarcin:
## Adversarial review of the rebased branch

Read-only pass over `git diff origin/master...HEAD`, focused on where #703's semantics change (channel from archive filename) could still disagree with #697's calibration.

### 1. Other producer/consumer disagreements — one found, already fixed above

Swept every read of `row[4]`/`row[5]`/`row[6]` and every call site of `is_continuous_channel`, `flat_channel_star`, `channel_for_archive`, `classify_evidence`, `cutoff_bin`, `star_for_row`, `PREGO_CHANNEL_COLUMN`, `PREGO_EVIDENCE_COLUMN`.

Clean now. `row[5]` is read in exactly one place (`prego.py:707`) and bound to `evidence`; both `channel_for_archive` sites (`541` calibration, `687` emit) agree. The only remaining textual mention of `row[5]` in the calibration pass is the comment explaining why it is *not* read there.

### 2. `channel_for_archive` robustness — fails open, by design, matching production

The three production archives in `download.yaml:310-319` match the three constants exactly (`literature`, `environmental_samples`, `annotated_genomes_isolates`). An unrecognised name returns the normalised stem, which makes `flat_channel_star` return `None`, which makes `star_for_row` return `None`, which makes `keep_row` return `True` — kept but uncalibrated.

That is the documented intent (dropping rows we cannot calibrate would delete data for a reason unrelated to confidence). The residual risk — upstream renames `environmental_samples.tar.gz` and the continuous channel silently stops being calibrated — is covered: the run prints `WARNING: no continuous-channel rows found` when the histogram comes back empty. Now documented in CLAUDE.md via #711.

### 3. The early `continue` — safe, and improved

Confirmed nothing downstream depends on `_collect_calibration` pre-populating `_payload_cache`: `_process_archive` calls `_ensure_payload` itself. So I moved the channel test *above* `_ensure_payload` — the calibration pass no longer decompresses the 8.7 GB isolates tarball only to skip it one line later.

This also fixed genuinely misleading logging. The payload member is named `database_pairs.tsv` inside all three tarballs, so the old line printed `calibration pass over database_pairs.tsv` three times, indistinguishably, including for archives it skipped.

### 4. Fixture realism — one vacuous-assertion risk found and closed

Both fixture files are well-formed and the malformed-row fixture (3 fields) correctly stayed in `database_pairs.tsv`. The risk was real though: `test_calibration_table_is_written_and_matches_what_ships` identified continuous edges by `" of " in prego_channel`, which under #703 matches *nothing* — the loop body would never execute and the test would have passed vacuously. Re-anchored to `prego_channel == CHANNEL_ENVIRONMENTAL`.

### 5. Anything else — one MEDIUM, filed as #712 and partly fixed here

`FLAT_CHANNEL_STARS` is keyed on PREGO resource classes while `flat_channel_star` keys on archive-derived channels, so **no key of the dict is a valid argument to the function in the same module**. `estimate_retention` scored unrecognised channels as 0.0 and excluded them, returning **0.265 where the answer was 0.734** — a 2.8x error, no exception.

Fixed the fail-closed half here (it was introduced by this PR's semantics move); left the rename to #712 as churn.

---

**Not merging** — awaiting review. `poetry run tox`: **916 passed**.

realmarcin:
## Second review pass — two more findings

Follow-up sweep specifically hunting the *vacuous assertion* class, since one instance of it had already bitten (`" of " in prego_channel`, which matched nothing under the new semantics, so the loop body never ran).

### Fixed here: the threshold test asserted on the wrong axis

`test_threshold_drops_only_rows_below_it` asserted `float(prego_score) >= 3.5` over **every** kept edge. That conflates two axes:

- flat rows are filtered on their **raw score**;
- continuous rows are filtered on the **calibrated star axis**, where any row at or above its resource's cutoff rates `STAR_MAX` regardless of raw score.

So a continuous row scoring 2.0 can legitimately survive a 3.5 cut. Probed the actual run: at `min_confidence=3.5` exactly two continuous rows survive, **both happening to score 4.0** — which is the only reason the assertion held. A fixture row scoring 2.0 in the top 12.5% of its resource would have failed the test for entirely correct behaviour.

Split into two arms, each with an explicit non-emptiness guard so neither can go vacuous:

- flat rows: raw score >= 3.5, as before;
- continuous rows: raw score >= that resource's cutoff, **read back from the emitted calibration table** — so the test also fails if the table and the filter ever diverge again, which is the #697 defect class.

Verified by mutation: returning `STAR_MAX` unconditionally from `star_for_row` now fails 2 tests. Previously it failed **none**.

### Filed, not fixed: #713 — the genomes channel is never instantiated

The fixture builds only `literature.tar.gz` and `environmental_samples.tar.gz`. `annotated_genomes_isolates` — **~47% of production edges** — has no end-to-end coverage at all.

Worse, `literature.tar.gz`'s payload actually holds genome/isolate rows (`src=JGI IMG ch=Isolates`), which in production live in the genomes archive. Since #703 derives the channel from the filename, those rows are now labelled `literature`. Harmless in effect — `star_for_row` uses the row's own score for any recognised flat channel — but it means the literature channel is never tested in its real shape (PMID at flat 3.0), and a regression isolated to the genomes channel would pass CI.

Left for #713 because fixing it shifts expected counts across several existing tests, which is more than a rebase PR should carry.

---

`poetry run tox`: **916 passed**. Still not merging — awaiting your review.

realmarcin:
## Independent review landed — 2 HIGH findings, both fixed

An independent reviewer found things my own passes did not. Verified each claim before acting.

### HIGH — the PREGO validation scripts had gone silently dead

`ubiquity_check.py`, `fold_enrichment_go.py` and `precision_assay_go.py` each carried a local `is_cont` matching the tally shape (`402 of 487 samples`) against `prego_channel`. Since that column became archive-derived it holds `environmental_samples`, so the match returns False for **every row, forever** — and does not crash. `ubiquity_check.py` printed "0 GO terms" with exit 0; the other two filed 100% of edges under "flat", collapsing the continuous-vs-flat contrast that is their entire purpose.

**These scripts produced the numbers in `docs/PREGO_SCORE_VALIDATION.md`** — the evidentiary basis for #697's design. Same producer/consumer class as the defect already fixed here, just outside `kg_microbe/`, which is exactly why my sweep missed it: I scoped the grep to the package.

Fixed with `scripts/prego_validation/channel_compat.py`, which detects the layout from the header (`prego_evidence` present ⇒ post-#703) so each script works on an `edges.tsv` from either era, and `assert_non_empty` refuses to report a result computed over zero continuous rows. Verified on both header shapes.

### HIGH — the fail-open was labelled "flat", which is the one thing it isn't

A flat channel is rated and thresholded on its own score. An unrecognised one is not rated at all, so `keep_row` returns True and every row bypasses the threshold. My skip message called both "flat channel", making a silent fail-open read like routine output.

The reviewer demonstrated it live using a near-miss archive already in the suite (`environmental.tar.gz` at line 618): 3 edges survived `min_confidence=4.0` — the maximum — on channel rather than score, with `not_provided` provenance and a normal-looking log. Production scenario: PREGO renames the genome archive and ~47% of edges quietly exempt themselves.

Now warns and names the three expected channels. New test asserts the warning fires, is not mislabelled as flat, and is *true* — the rows really do survive `min_confidence=4.0`.

### MEDIUM/LOW — two assertions could not detect the failure they described

- `all(e["knowledge_level"] for e in edges)` passes when every edge is `not_provided` (truthy). The detector that should have caught the finding above could not. Now asserts membership in the documented value sets.
- `all("prego_evidence" in e for e in edges)` tests the **header**, not the value — `e` is a DictReader dict, so it passes with every cell empty, which is precisely what the surrounding comment claims to guard. Now asserts the values.

### Filed, not fixed here

- **#713** genome channel (~47% of production edges) never exercised end-to-end
- **#714** `classify_evidence` asserts `habitat` for every unrecognised value — with the tradeoff spelled out, since the naive fix would destroy a real signal
- **#715** `edges.tsv` grows ~44% (7.38 → ~10.6 GiB), relevant to #693
- **#716** three smaller findings, incl. one pre-existing tautological test

### Confirmed clean by the reviewer

The producer/consumer sweep inside `kg_microbe/`; fixture well-formedness; `kg_path_review.py`; and the `FLAT_CHANNEL_STARS` treatment in b358d5e. They also verified on the **real payloads** that the old shape-match and the new archive-name select an identical row set — so shipped cutoffs are unchanged, which my probe could not establish from fixtures alone.

---

`poetry run tox`: **917 passed**. Still not merging — awaiting your review.

Issue #683, opened 2026-08-06
Re-run transforms and re-merge: shipped merged KG carries stale + orphaned content


realmarcin:
Tracking issue for repairing the shipped artifact. #680 fixes the code; this issue covers regenerating the data.

## Current state

`data/merged/merged-kg.tar.gz` is dated **Aug 3 21:01** and contains at least four defects:

1. **Two phantom nodes** with non-CURIE IDs — literally `subject` and `object`, both `biolink:NamedThing` / `provided_by=Graph`. These come from a doubled header row in `data/transformed/ontologies/ec_edges.tsv`. Fixed in code by #680; requires re-running the `ontologies` transform (or at least the `ec` post-process) and re-merging.

2. **62,521 orphan `bacdive:*` stub nodes**, all `provided_by=Graph` with no name, synthesized by KGX from the 62,096 `bacdive:<id> -in_taxon-> lpsn:<n>` edges whose subject was never emitted as a node. Fixed in code by #680; requires re-running the `bacdive` transform.

3. **Stale MediaDive content.** `data/transformed/mediadive/edges.tsv` (Jul 24 15:38) predates `data/raw/mediadive/*.json` (Jul 25 00:14) and was built from 2023-era YAML + 2025 HTTP caches. See #681. Requires re-running the `mediadive` transform with bulk data present.

4. **Merge is older than two transforms.** `kg-path-review` prints:
   `data/merged/merged-kg.tar.gz is older than transform output(s): microbedecoder, prego`

## Suggested order

1. Merge #680.
2. Re-run `ontologies` (ec), `bacdive`, and `mediadive` transforms. Confirm `data/raw/mediadive/` is fully downloaded first — #680's guard will now refuse otherwise.
3. `poetry run kg merge -y merge.yaml`.
4. Re-run `kg-model-review --merged` and `kg-path-review` archetypes to confirm the ERROR and the two CRITICALs are gone.

## Verification checks

After re-merge, these should all return zero:

```bash
# phantom header nodes
awk -F'\t' 'NR>1 && ($1=="subject"||$1=="object")' merged-kg_nodes.tsv | wc -l
# orphan bacdive stubs
awk -F'\t' 'NR>1 && $1 ~ /^bacdive:/ && $6=="Graph"' merged-kg_nodes.tsv | wc -l
# mediadive wrapper solution should now exist
grep -c '^mediadive.solution:6570' merged-kg_nodes.tsv   # expect >= 1
```

[...]

PR #704, opened 2026-08-07
fix(ontologies): drop stray header rows by content, not position

realmarcin:
Reopens and properly fixes the EC duplicate-header defect that #680 claimed to close.

## #680's fix did not work

Re-running the ec transform on master, with that fix in place, produced **byte-identical output** — stray header row included. The transform ran (mtime updated), rewrote both files, and reproduced the defect exactly.

## Why

It addressed the wrong stage. #680 skipped index 0 in the ec branch's read loop, assuming the branch was failing to consume the incoming header. In fact **the parse output already carries two header lines**: pandas consumes the first, the second survives as a data row, and skipping index 0 removes only the one already being consumed.

The signature is visible in the shipped file — the two lines differ:

```
line 1:  subject  predicate  object  relation  primary_knowledge_source  knowledge_level  agent_type
line 2:  subject  predicate  object  relation  knowledge_source          knowledge_level  agent_type
```

Line 1 is the canonical header `_normalize_schema` renames into place. Line 2 still says the deprecated `knowledge_source` because it was already body content when the rename ran.

## The fix

Moved into `_normalize_schema`, which runs for **every** ontology rather than just ec, and filtered **by content**: any row whose subject column is literally `"subject"` is dropped. No real edge has that subject.

Position-based filtering is what let this survive a full re-run, so it is deliberately not used.

## Verified end to end

```
[_normalize_schema] ec_edges.tsv: dropped 1 stray header row(s)
```

`ec_edges.tsv` goes from 12,023 to 12,022 lines, zero strays. This is what was producing the two empty `biolink:NamedThing` nodes in the merged KG.

Scanned all 14 ontology edge files: `ec` was the only one affected, but the fix is general.

## Testing

The existing test could not catch this — its fixture only ever had one header, so skipping index 0 always sufficed. The new test builds a file with a duplicate header **in the body** and asserts it is removed.

`poetry run tox` green: **845 passed**.

## Context

Found while running #683 (re-run transforms, re-merge). The canary — one scoped transform, verified before fanning out — is what caught it. bacdive, microbedecoder and mediadive were already re-run correctly; ec was the last outstanding transform and would have carried the defect into the new merge.

### DIFF
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -975,6 +975,22 @@ class OntologiesTransform(Transform):
 
         if edges_file.is_file():
             df = pd.read_csv(edges_file, sep="\t", low_memory=False)
+            # Drop header rows that leaked into the body. The ec pipeline has
+            # produced a file whose parse output already carried two header
+            # lines: the first is consumed as the header, the second survives
+            # as a data row and reaches the merged KG, where KGX reads
+            # "subject"/"object" as endpoint CURIEs and synthesizes two empty
+            # biolink:NamedThing nodes.
+            #
+            # Filtering by content rather than by position deliberately: an
+            # earlier fix skipped index 0 in the ec branch, which removes one
+            # header but not a second, and left this defect in place through a
+            # full re-run. No real edge has "subject" as its subject.
+            if SUBJECT_COLUMN in df.columns:
+                stray = df[SUBJECT_COLUMN].astype(str) == SUBJECT_COLUMN
+                if stray.any():
+                    print(f"  [_normalize_schema] {edges_file.name}: dropped {int(stray.sum())} stray header row(s)")
+                    df = df[~stray]
             dropped_edge_cols = []
             renamed = False
             if "id" in df.columns:
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -858,16 +858,18 @@ class OntologiesTransform(Transform):
                         line = _replace_special_prefixes(line)
                         line = replace_category_ontology(line, id_index, category_index)
                         new_nf_lines.append(line + "\n")
-                # Update prefixes in edges file. The incoming header must be
-                # dropped by position: a canonical header is written below, and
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review complete — and it overturned my diagnosis twice over

**#680 did not fail to fix this defect. It introduced it. And my follow-up explanation in this PR was also wrong.**

### Verified on disk

`data/transformed_pre20260801/ontologies/ec_edges.tsv` (Jul 27, pre-#680) has **zero** stray header rows. #680 replaced

    if line.startswith("id"):

with a guard matching `SUBJECT_COLUMN`, asserting in its comment that *"an edges header starts with subject (not id)"*. That is backwards — KGX's edges TSV leads with an `id` column, so its header's first field **is** `id`. The old guard matched the real header; the new one never fires.

My second explanation — that the parse output carried two header lines — was also false. A direct probe of KGX's obojson→tsv output shows **one** header, 265,500 lines, zero data rows starting with `id`. The second header is written by the ec branch itself.

### What changed in this PR

- **Producer repaired** at source: the guard now accepts both `id` and `subject`, so it survives a schema change in either direction.
- **`_normalize_schema` filter demoted to an explicit backstop**, not the primary fix. On a real ec run it is now **silent** — which is the signal the leak is stopped at source rather than swept up.
- **Both comments corrected.** This file had accumulated three consecutive incorrect explanations of the same five lines.

### Test fidelity — the part that mattered most

The old fixtures used the canonical header (first field `subject`). That is exactly how #680 shipped green: its guard fired in the test and never in production. Fixtures now use the real KGX shape (`id` first, `urn:uuid:` values) and drive `post_process` rather than calling `_normalize_schema` directly.

**My first mutation test was insufficient.** Restoring #680's broken guard still passed — because the backstop silently repaired it. Defence in depth had made the producer bug untestable through the output. The test now also asserts the backstop does **not** fire, so a broken producer fails even when the final artifact is correct. With that, restoring #680's guard fails the suite.

### Filed, not fixed here

- **#705** — the ec edges path relies on unguarded pandas index promotion. A 7-column header over an 8-field body works only by coincidence; one extra upstream column shifts every field silently, exit 0. Highest real risk, pre-existing.
- **#706** — test fixtures must use the real KGX header shape; plus the layered-guard lesson above.
- **#707** — no equivalent guard on the nodes path (symmetry gap, currently no producer).

`poetry run tox` green: **846 passed**.

PR #711, opened 2026-08-07
docs(prego): document PREGO_MIN_CONFIDENCE and its relation to merge.noprego.yaml

### LINKED ISSUE #709: prego: PREGO_MIN_CONFIDENCE is undocumented, and the docs argue against a single global knob
From the adversarial review of #697. MEDIUM.

## Undocumented

`PREGO_MIN_CONFIDENCE` appears in no README, no `.env.example`, not in CLAUDE.md's environment-variable table, and not in `docs/PREGO_SCORE_VALIDATION.md`. Every sibling knob (`KG_SEMSQL_BUILD`, `KG_GO_VERSION_CHECK`, `KG_MEDIADIVE_ALLOW_STALE_CACHE`, `METATRAITS_WORKERS`) is documented in CLAUDE.md.

## And the design is contested by its own analysis

`docs/PREGO_SCORE_VALIDATION.md` recommendation #2:

> **Per-predicate thresholds, not one global knob.** A single `min_confidence` is a validated quality filter on `location_of` and an arbitrary cut on `capable_of` simultaneously.

So #697 ships precisely the knob its own document recommends against. That was defensible while the predicate split looked real; the split is now **NOT ESTABLISHED**, which weakens the argument for per-predicate thresholds too — but it does not restore the argument for a single one.

## Worth stating plainly in the docs

`merge.noprego.yaml` (#691) already exists as a size lever and needs no calibration, no second pass, and no threshold semantics. #697 should be documented as a **finer-grained alternative** to it — keep some PREGO rather than none — not as the only or the primary mechanism.

## Suggested

- Add `PREGO_MIN_CONFIDENCE` to CLAUDE.md's environment-variable section with the size-lever framing and an explicit "not a quality filter" note.
- State the relationship to `merge.noprego.yaml` in `PREGO_SCORE_VALIDATION.md`.

realmarcin:
Closes #709.

## What

`PREGO_MIN_CONFIDENCE` shipped in #697 undocumented. This adds it to CLAUDE.md's environment-variable section and states its relationship to `merge.noprego.yaml` in `docs/PREGO_SCORE_VALIDATION.md`.

## Why the per-channel table matters

The knob does not mean the same thing in every channel, so a one-line entry would have been misleading:

| channel | share | what the threshold does |
|---|---|---|
| `environmental_samples` | 53.00% | Genuine ranking. Retains ~`1 - τ/4` per resource. |
| `annotated_genomes_isolates` | 46.89% | All-or-nothing. Flat 4.0 assigned by PREGO's authors. |
| `literature` | 0.05% | All-or-nothing. Flat 3.0 — the channel vanishes once `τ > 3`. |

Two consequences are stated explicitly: above `τ = 3` you delete the literature channel on **provenance, not quality**; and within the continuous channel the score correlates with GO-term ubiquity, so raising `τ` strips rare and taxon-specific annotations before wrong ones.

## The honest counterweight

The issue's substantive point was that `docs/PREGO_SCORE_VALIDATION.md` recommendation 2 argues *against* a single global knob, and #697 shipped exactly that. Rather than paper over it, the new section says so, and frames the two mechanisms:

- **`merge.noprego.yaml` is the primary size lever** — drops ~76% of merge input (44.7 M edges, 7.38 GiB) with no calibration, no histogram, no cutoffs, no threshold semantics.
- **`PREGO_MIN_CONFIDENCE` is the finer-grained alternative** for keeping *some* PREGO rather than none.

## Verification

- Channel shares checked against the measured table at `PREGO_SCORE_VALIDATION.md:102-107` (53.00 / 28.66 + 9.17 + 7.27 + 1.79 / 0.05).
- The Spearman correlation was **not** quoted as a number: the script measures GO-term *degree*, not sample count as an earlier draft claimed, and the value is not recorded in the doc. Described qualitatively with a pointer to `scripts/prego_validation/ubiquity_check.py`.

Docs only — no code paths touched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,11 +250,17 @@ Optional, for the ontology transforms:
 
   Two consequences worth internalising. Above `τ = 3` you delete the literature
   channel outright — on provenance, not on quality. And within the continuous
-  channel the score correlates positively with how many taxa carry a GO term
-  (`scripts/prego_validation/ubiquity_check.py`), so raising `τ` preferentially
-  keeps common, well-covered terms and strips rare, taxon-specific ones before
-  it strips wrong ones. An unrecognised channel is kept rather than dropped, so
-  a newly added archive fails open.
+  channel the score tracks how *common* a GO term is rather than how well
+  supported it is: measured Spearman +0.2592 of term degree against mean score,
+  with mean score climbing steeply across the bottom three deciles (0.67 → 1.98)
+  then flat at ~2.0 for the rest. So raising `τ` strips the rare,
+  taxon-specific tail first and barely discriminates within the bulk — the
+  opposite of what you want if the goal is to keep specific, informative edges.
+  Full decile table in `docs/PREGO_SCORE_VALIDATION.md`; reproduce with
+  `scripts/prego_validation/ubiquity_check.py`.
+
+  An unrecognised channel is kept rather than dropped, so a newly added or
+  renamed archive fails open — the transform warns when it sees one.
 
   Calibration is per-resource (MGnify and MG-RAST have different marginals) and
   runs as a first pass over the archives; each run writes the cutoffs it
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -226,6 +226,38 @@ Optional, for the ontology transforms:
   `strict` (raise) or `warn`. GO defaults to strict because a mismatch silently
   miscategorises terms; the other two default to warn.
 
+- `PREGO_MIN_CONFIDENCE`: star threshold in `[0, 4]` applied to the PREGO
+  transform. Default `0` — a no-op that emits all ~44.7 M edges, so unset
+  behaviour is unchanged. **This is a size lever, not a quality filter.** See
+  the table below before setting it.
+
+  PREGO's score is not one scale, so the knob does not mean the same thing in
+  every channel (`docs/PREGO_SCORE_VALIDATION.md` has the measurements):
+
+  | channel | share | what the threshold does |
+  |---|---|---|
+  | `environmental_samples` | ~53% | Genuine ranking. Retains about `1 - τ/4` of each resource, by within-resource empirical CDF. |
+  | `annotated_genomes_isolates` | ~47% | All-or-nothing. Every row is a flat 4.0 assigned by PREGO's authors, so nothing is dropped until `τ > 4`. |
+  | `literature` | ~0.05% | All-or-nothing. Flat 3.0, so the entire channel vanishes the moment `τ > 3`. |
+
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review of this PR — one factual error found and fixed

Reviewed the doc claims against the **real payloads** rather than against the summary they were drawn from. One was wrong.

### #717 — the genome channel is not uniformly 4.0

The table asserted:

> All-or-nothing. Every row is a flat 4.0 assigned by PREGO's authors, so nothing is dropped until `τ > 4`.

Measured on `annotated_genomes_isolates_extracted/database_pairs.tsv` (8.68 GB) at four offsets:

| sample | score 4 | score 3 |
|---|---:|---:|
| head 3M rows | 2,992,738 | 7,262 (0.243%) |
| offset 2 GB | 799,237 | 762 (0.095%) |
| offset 5 GB | 799,999 | 0 |
| offset 8 GB | 799,999 | 0 |

Roughly 0.1% carry a 3, and **not only the PMID-evidenced ones** — `Isolates` (1,762) and `Single Amplified Genome` (553) appear at 3.

This matters because `star_for_row` deliberately uses each flat row's *own* score rather than the channel constant, so those rows drop once `τ > 3` — on the order of 20k edges. The table asserted the opposite of what happens, in the table someone consults when choosing τ. Corrected.

### The ubiquity claim is now measured, not gestured at

I originally refused to quote a Spearman figure here because the number was not recorded and I could not verify the operand. Ran it: full 7.93 GB `edges.tsv`, 3,830 GO terms with ≥50 continuous-channel edges.

**Spearman (GO term degree vs mean score): +0.2592**

The coefficient alone is misleading, so the decile table is now in `PREGO_SCORE_VALIDATION.md`. The relationship is *not* a smooth gradient — mean score climbs steeply across the bottom three deciles (0.666 → 1.977) then sits flat at ~2.0 for deciles 4–10. The score separates rare terms from everything else and carries almost no ordering among the common ones.

That sharpens the operational warning rather than softening it: raising τ strips the rare, taxon-specific tail first and barely discriminates within the bulk.

### Bonus: this was also the real-data canary for #703

The on-disk `edges.tsv` is the **pre-#703** layout, so the run exercised `channel_compat`'s backwards-compatibility path on the real 7.93 GB file — it correctly reported `layout: pre-#703 (tally shape in prego_channel)` and returned a non-empty selection. Previously that path was only covered by synthetic headers.

### Verified and unchanged

Channel shares (53.00 / 46.89 / 0.05) against `PREGO_SCORE_VALIDATION.md:102-107`; the `1 - τ/4` retention identity against `ScoreHistogram.cutoff`; and the literature-channel claim (flat 3.0, vanishes at `τ > 3`) against `flat_channel_star` + `keep_row`.

Issue #716, opened 2026-08-07
prego: three smaller review findings (habitat rows in genome archive, fail-before-write, tautological retention test)


realmarcin:
From the adversarial review of #703. Grouped because each is **LOW** on its own.

## 1. Habitat-valued rows in the genome archive get `knowledge_assertion`

`kg_microbe/transform_utils/prego/utils.py` — `edge_metadata_for`

The real `annotated_genomes_isolates` payload carries column-6 values like `Marginal Sea`, `Hydrothermal vents`, `Birds` at **score 3** — the lower BioProject/text-mined tier, not 4. `edge_metadata_for` overrides the channel default only on `EVIDENCE_PUBLICATION`, so these fall through to the genome-channel default and receive `knowledge_assertion` / `automated_agent`: the *highest* provenance tier, despite carrying the same score PREGO assigns its text-mined rows.

Measured at roughly 756 of ~300k rows in a 60 MB sample — small, but systematically mis-tiered in the confident direction, which is the worse direction. Related to #714 (the habitat catch-all).

## 2. The filtered path lost fail-before-write for flat archives

`kg_microbe/transform_utils/prego/prego.py` — the calibration-pass skip vs the emit loop

`_ensure_payload` also validates that a tarball actually contains a `database_pairs.tsv` member (raises `SystemExit`). Before the fix that moved the channel test above `_ensure_payload`, a run with `min_confidence > 0` extracted all three archives in pass 1 — i.e. **before** `nodes.tsv` / `edges.tsv` are opened for writing. Now flat archives are first touched inside the write block, and `sorted()` puts `annotated_genomes_isolates.tar.gz` first.

Consequence: a corrupt genome tarball aborts *after* the output headers are written, replacing the previous run's outputs with a 1-line file.

Note the default path (`min_confidence == 0`, which skips calibration entirely) has always behaved this way, so this removes an incidental property of the filtered path rather than a designed guarantee. Still worth closing — a cheap pre-flight that opens each tarball and checks for the member, before any output file is created, would restore it for both paths.

## 3. `test_tau_four_is_provenance_dominated` is tautological

`tests/test_prego_calibration.py`

At `tau = STAR_MAX` the continuous term of `estimate_retention` is exactly `continuous_share * (1 - 4/4) = 0`, so `retained` and `flat_only` evaluate the same expression and `flat_only / retained` is always exactly 1.0. The assertion `> 0.99` cannot fail for **any** implementation.

[...]

Issue #715, opened 2026-08-07
prego: #703 metadata columns grow edges.tsv ~44% (7.38 -> ~10.6 GiB), relevant to #693


realmarcin:
From the adversarial review of #703. **MEDIUM** — not a defect, a cost that should be sized before the next merge rather than discovered during it.

## Measurement

Per emitted row, #703 adds:

| addition | approx bytes/row |
|---|---:|
| `prego_channel` now holds a channel name (was col-6 verbatim) | ~23 (weighted) |
| `prego_evidence_class` | ~13 |
| `knowledge_level` | ~21 |
| `agent_type` | ~19 |
| 2 extra tabs | 2 |

The raw column-6 string moves to `prego_evidence` rather than being duplicated, so it is not double-counted.

**~78 B/row x 44.7M rows ≈ 3.3 GiB**, taking `data/transformed/prego/edges.tsv` from **7.38 GiB to ~10.6 GiB (+44%)**.

## Why it matters now

#693 (merge RAM exhaustion) is open, and PREGO is the single largest edge block in the merged KG — KGX holds all source graphs in the parent simultaneously. A 44% growth in the largest block is material to that.

The metadata is the *point* of #703 and worth having; `knowledge_level` / `agent_type` were shipping empty, which made 44.7M text-mined and statistically-derived associations indistinguishable from curated assertions. This issue is only to make sure the cost is stated out loud and factored into the #693 sizing, not to argue against the columns.

## Suggested

- Re-measure the real `edges.tsv` after the next full transform and confirm the estimate.
- Fold the new number into the #693 analysis, and into whichever of `merge.noprego.yaml` / `PREGO_MIN_CONFIDENCE` (see #709 / #711) is chosen as the lever.

realmarcin:
Measured in #718 (merged as 8cddc45), over all 44,716,161 real rows rather than a sample:

| | |
|---|---:|
| `edges.tsv` before | 7.38 GiB |
| bytes added | 3.54 GiB (**85.1 B/row**) |
| projected | **10.93 GiB (+48.0%)** |

**Above this issue's 78 B/row estimate.** Recorded in `docs/PREGO_SCORE_VALIDATION.md` under a new sizing section, alongside the #693 implications.

Leaving this open: the remaining action is budgeting the next merge against the measured figure, not a code change.

realmarcin:
[...]

Issue #714, opened 2026-08-07
prego: classify_evidence asserts 'habitat' for every unrecognized value


realmarcin:
From the adversarial review of #703. **MEDIUM.**

`kg_microbe/transform_utils/prego/utils.py` — `classify_evidence` returns `EVIDENCE_UNKNOWN` **only for the empty string**. The terminal `return EVIDENCE_HABITAT` is the catch-all, so anything that is not a tally, not `PMID:`-prefixed, and not one of four resource-class prefixes is *positively asserted* to be a habitat name.

## Failure scenario

PREGO adds a fifth resource class (e.g. `Metagenome bin`), or ships a truncated column-6 value. It is emitted as `prego_evidence_class=habitat`. A consumer doing the exact thing this column was added for — `WHERE prego_evidence_class='habitat'` — silently gets habitats mixed with unclassified junk, with nothing flagging the drift.

## The tradeoff, which is why this is filed rather than hotfixed

The obvious fix (make the fallback `EVIDENCE_UNKNOWN`) is **not clearly correct**. The real `annotated_genomes_isolates` payload carries genuine habitat names in column 6 — `Marginal Sea`, `Hydrothermal vents`, `Birds` — and there is no habitat vocabulary in the transform to recognise them by. So:

- Keeping `habitat` as the catch-all: correctly labels real habitats, silently absorbs new resource classes.
- Switching to `unknown`: never mislabels junk, but destroys a real signal for the majority of values that genuinely *are* habitats.

Neither is free. Options worth weighing:

1. **Count the fallthroughs in `_stats`** and print them in the run summary, so a new resource class shows up as a jump in the habitat count rather than silently. Non-destructive; makes drift visible without a vocabulary.
2. Match habitats against an actual controlled vocabulary (ENVO labels are already loaded elsewhere in the pipeline) and reserve `unknown` for genuine non-matches. Correct, but materially more work.
3. Rename the class to something honest about being a residual (`uncategorised`), and accept that it mixes both.

Option 1 is the cheapest thing that makes the problem observable, and does not require deciding the semantics question first.

realmarcin:
Closed by #718 (merged as 8cddc45).

[...]

Issue #713, opened 2026-08-07
prego tests: annotated_genomes_isolates channel (~47% of production edges) is never instantiated in any test


realmarcin:
From the adversarial review of #703. **MEDIUM.**

## The gap

The test fixture builds exactly two archives (`tests/test_prego_transform.py:63-66`):

- `literature.tar.gz`
- `environmental_samples.tar.gz`

`annotated_genomes_isolates.tar.gz` is never built, so `CHANNEL_GENOMES` has **no end-to-end coverage at all** — despite being ~47% of production edges (`Isolates` 28.66% + `Genome annotation` 9.17% + MAG 7.27% + SAG 1.79%).

This matters more since #703, because the channel is now derived from the archive filename. `channel_for_archive("annotated_genomes_isolates.tar.gz")` is exercised only as a unit test on the string; no test ever runs a row through the transform under that channel.

## Made worse by the fixture mislabelling what it does have

`database_pairs.tsv` — the payload of `literature.tar.gz` — actually contains genome/isolate-shaped rows:

```
src=JGI IMG      ch=Isolates
src=BioProject   ch=PMID:12345678
```

In production those `JGI IMG` / `Isolates` rows live in `annotated_genomes_isolates.tar.gz`, not in `literature.tar.gz`. Under the archive-derived channel they are now labelled `literature`, which is wrong. It happens to be harmless — `star_for_row` returns the row's own score for any recognised flat channel, so a score-4 row in the literature archive still rates 4.0 — but it means:

1. The `literature` channel is never tested in its real production shape (PMID rows at a flat 3.0).
2. `flat_channel_star(CHANNEL_GENOMES) == 4.0` is never exercised on the emit path.
3. A regression that broke the genomes channel specifically would pass CI.

## Suggested

Add a third fixture archive `annotated_genomes_isolates.tar.gz` and move the `JGI IMG` / `Isolates` rows into it, leaving `literature.tar.gz` holding only the `BioProject` / PMID rows at 3.0. That makes all three channels real and matches production layout.

Note this will shift some existing expected counts — e.g. `test_edges_carry_the_originating_resource` asserts `{"BioProject", "JGI IMG", "MGnify"}`, and `test_taxon_to_go_edges_use_capable_of` asserts 15 edges. Those numbers should stay the same in total, but the per-channel split changes; check anything that filters by `prego_channel`.

## Context

[...]

PR #718, opened 2026-08-07
fix(prego): close the five deferred review findings (#712–#716)

### LINKED ISSUE #712: prego: FLAT_CHANNEL_STARS keys no longer resolve through flat_channel_star(); estimate_retention fails silently
From the adversarial review of #703. **MEDIUM.**

## The inconsistency

`kg_microbe/transform_utils/prego/calibration.py` defines two things that look like they belong together but no longer interoperate:

- `FLAT_CHANNEL_STARS` (line 68) is keyed on **PREGO's resource-class names**: `Isolates`, `Isolates GOLD`, `Genome annotation`, `Metagenome-Assembled Genome`, `Single Amplified Genome`.
- `flat_channel_star()` (line 98) keys on the **archive-derived channel**: `annotated_genomes_isolates`, `literature`.

#703 moved the channel from a data column to the archive filename, which re-keyed the function but left the constant behind. Measured:

```
'Isolates'                      documented=4.0  flat_channel_star()=None
'Isolates GOLD'                 documented=4.0  flat_channel_star()=None
'Genome annotation'             documented=4.0  flat_channel_star()=None
'Metagenome-Assembled Genome'   documented=4.0  flat_channel_star()=None
'Single Amplified Genome'       documented=4.0  flat_channel_star()=None
```

Every key of the dict returns `None` from the function in the same module.

## Why it matters: estimate_retention fails silently

`estimate_retention` (line 336) scores unrecognised channels as `0.0` via `(flat_channel_star(channel) or 0.0)` and excludes them, rather than rejecting them:

```
estimate_retention({'Isolates':0.2866,'Genome annotation':0.0917}, 2.0, 0.53)  -> 0.265
estimate_retention({'annotated_genomes_isolates':0.4689,'literature':0.0005}, 2.0, 0.53) -> 0.734
```

**A 2.8x error with no exception raised.** A caller who reasonably reaches for `FLAT_CHANNEL_STARS.keys()` — the only enumeration of flat channels in the module — gets a confidently wrong retention estimate. `estimate_retention` exists precisely to answer "what will this threshold cost me before I run it", so a silent 2.8x error defeats its purpose.

## Not currently a production bug

- `estimate_retention` has no production call site (tests only, and `tests/test_prego_calibration.py` was correctly re-keyed to `MEASURED_FLAT_SHARES = {CHANNEL_GENOMES: 0.4689, CHANNEL_LITERATURE: 0.0005}`).
- `star_for_row` returns the row's own score for recognised flat channels, so `FLAT_CHANNEL_STARS` is not consulted on the emit path.

So this is a latent footgun, not a live defect — filed rather than hotfixed.

## Suggested

1. Make `estimate_retention` **fail closed**: raise on a channel key that `flat_channel_star` does not recognise, instead of silently scoring it 0.
2. Keep `FLAT_CHANNEL_STARS` (it documents the paper's per-resource-class constants, which is genuinely useful provenance) but rename it to say what it is keyed on — e.g. `PREGO_RESOURCE_CLASS_STARS` — and state in its docstring that it is **documentation of upstream values, not a lookup table for our channels**.

realmarcin:
Closes #712, #713, #714, #716. Addresses #715 with a measurement.

Follow-up to #703. Each of these was filed rather than fixed there, deliberately; this is the cleanup pass.

## #713 — the genome channel had no end-to-end coverage

`annotated_genomes_isolates` is **~47% of production edges** and no test ever ran a row through it. Worse, the rows that should have been in it sat in `literature.tar.gz`, so since the channel is derived from the archive filename they were emitted as `literature` — the exact inverse of production for those same bytes.

The fixture now builds all three production archives. Edge count is **unchanged at 18**, so the split is behaviour-preserving. Verified all three channels now appear:

```
channels        : {annotated_genomes_isolates: 4, environmental_samples: 12, literature: 2}
knowledge_level : {knowledge_assertion: 4, statistical_association: 12, prediction: 2}
evidence_class  : {resource_class: 4, sample_count: 12, publication: 2}
```

A new test asserts the **full expected sets**, so dropping an archive or adding a channel upstream fails here rather than silently narrowing coverage.

## #714 — measured before fixing, because the obvious fix was wrong

`classify_evidence` returns `habitat` for anything unrecognised. The tempting fix — make it `unknown` — would have been a mistake. Measured over 3M real genome rows first:

```
resource_class    2,993,360
publication           4,947
habitat               1,693   <- 42 distinct values, ALL genuine habitats
```

`Aquatic`, `Plants`, `Terrestrial`, `Oceanic`, `Wastewater`, `Hydrothermal vents`… Reclassifying those as `unknown` would destroy a real signal to guard against a hypothetical one.

So instead the distinct values are **tracked and reported**. The real value space is small and stable, so a fifth PREGO resource class shows up as an unfamiliar entry rather than being silently asserted to be a habitat. Bounded at 1000 keys — a dict on a free-text column is an O(N) memory risk across 44.7M rows — with occurrences still counted past the cap and a warning when it trips.

## #716 — three findings

1. **Habitat rows overclaimed provenance.** They inherited the genome channel's `knowledge_assertion` (highest tier) despite carrying score 3 — PREGO's *lower* tier, the same one it gives text-mined rows. They come from sample metadata, not the annotation pipeline the channel is named for. Now `observation`, verified against the loaded Biolink `KnowledgeLevelEnum` rather than invented.

2. **Fail-before-write restored.** `_ensure_payload` validates lazily and the genome archive sorts first, so a corrupt tarball aborted *after* headers were written — replacing a good previous run's outputs with a one-line file. A pre-flight now checks every archive before anything is opened for writing, reading only the tar index so it stays cheap on the 8.7 GB archive. Test asserts the prior outputs survive.

3. **A tautological test.** At tau exactly `STAR_MAX` the continuous term is `share * (1 - 4/4)` = 0, so both sides of the ratio reduced to the same expression — it could not fail for any implementation. Moved to tau=3.9 and now asserts the continuous contribution is actually present.

## #712 — rename

`FLAT_CHANNEL_STARS` → `PREGO_RESOURCE_CLASS_STARS`. It is keyed on PREGO resource classes, not our channels, and no key of it is a valid argument to `flat_channel_star`.

## #715 — measured, not estimated

Over all **44,716,161 real rows**, not a sample:

| | |
|---|---:|
| `edges.tsv` before | 7.38 GiB |
| bytes added | 3.54 GiB (85.1 B/row) |
| projected | **10.93 GiB (+48.0%)** |

**Above the 78 B/row estimate in the issue.** Recorded in `PREGO_SCORE_VALIDATION.md` against #693 — KGX holds every source graph in the parent, and PREGO is the largest edge block. Leaving #715 open since the action item is budgeting the next merge, not a code change.

---

`poetry run tox`: **923 passed**.


realmarcin:
## Review pass — mutation-tested rather than read

I wrote these tests, so reading them proves nothing about whether they can fail. Every new test was checked by mutation instead.

| mutation | tests failed |
|---|---:|
| M1 — drop the `habitat → observation` branch | **2** |
| M2 — remove the archive pre-flight entirely | **1** |
| M3 — stop tracking habitat values | **1** |
| M4 — `estimate_retention` ignores the continuous term | **2** |
| M5 — fixture drops the genome archive (the #713 regression) | **11** |
| *(restored)* | 0 — 85 pass |

**M4 is the one worth calling out.** It confirms the tautology fix has real content: `test_tau_four_is_provenance_dominated` now fails when `estimate_retention` drops its continuous term. At the old `tau = STAR_MAX` that mutation was *undetectable*, because the continuous term was already exactly zero on both sides of the comparison.

**M5** confirms #713 is genuinely closed rather than papered over — reintroducing the two-archive fixture now breaks 11 tests, where before it broke none.

## One self-correction, found by checking my own claim

The pre-flight's comment said it "reads only the tar index, not the members". **A `.tar.gz` has no central directory**, so that reasoning is wrong — even though the conclusion happens to hold. Measured on the real 281 MB / 8.7 GB-payload genome archive:

```
any() short-circuit, payload is first member : 0.00s
full scan, no matching member (corrupt case) : 4.30s
```

The happy path is free because `any()` stops at the first matching member and the payload *is* that member — not because anything is indexed. Reaching a later member would mean decompressing everything before it, which matters if the archive layout ever gains members. Comment corrected in b0e9e2f rather than left as a plausible-sounding wrong explanation.

## On #714: measured before fixing, and the measurement changed the fix

The issue proposed making the catch-all `unknown`. Measuring first showed that would have been a mistake — the bucket is 1,693 rows across 42 distinct values, **all genuine habitat names**. Reclassifying them would have destroyed a real signal to guard against a hypothetical one. Tracking-and-reporting keeps both properties.

## Caveat

An independent reviewer was spawned but has not reported. Everything above is my own analysis of my own work, with mutation testing standing in for a second pair of eyes. That is better than reading it twice, but it is not the same as independent review.

`poetry run tox`: **923 passed**.

realmarcin:
## Closing the two questions mutation testing could not answer

Mutations prove a test *can* fail. They cannot tell you whether a refactor changed behaviour it shouldn't have, or whether a provenance change has wider blast radius than claimed. Both checked directly.

### A — the fixture restructure is behaviour-preserving on graph content

Ran the transform twice over identical row content, once with the old 2-archive layout (everything in `literature.tar.gz`) and once with the new 3-archive layout, then compared:

```
old-layout edges: 19   new-layout edges: 19
core triples identical (subject, predicate, object, score): True
rows whose channel/provenance changed: 5 of 19
```

The graph content is byte-identical. What changed is exactly the 5 genome rows, and they changed *toward* correctness:

| subject → object | before | after |
|---|---|---|
| `ENVO:00000011 → NCBITaxon:693444` | `literature` / `prediction` | `annotated_genomes_isolates` / `knowledge_assertion` |
| `NCBITaxon:100 → GO:0000034` | `literature` / `prediction` | `annotated_genomes_isolates` / `knowledge_assertion` |
| `NCBITaxon:100 → GO:0005634` | `literature` / `prediction` | `annotated_genomes_isolates` / `knowledge_assertion` |
| `NCBITaxon:100 → GO:0006355` | `literature` / `prediction` | `annotated_genomes_isolates` / `knowledge_assertion` |
| `NCBITaxon:100 → GO:0008150` | `literature` / `prediction` | `annotated_genomes_isolates` / **`observation`** |

That last row is the habitat-evidenced one, correctly landing on `observation` rather than inheriting the channel default. This is the #713 defect visible in miniature: those rows were being emitted as text-mined literature when they are genome-annotation rows.

### C — the `observation` change does not reach the environmental channel

The concern was scope creep: `edge_metadata_for`'s new habitat branch sits above the channel checks, so it would also reclassify environmental rows from `statistical_association` to `observation` if any carried habitat evidence.

Checked the real 5.48 GB environmental payload at two offsets (2M rows at the head, 500k at 3 GB):

```
sample_count    2,000,000     <- head
sample_count      499,999     <- offset 3 GB
```

**100% `sample_count`, zero habitat rows.** So the change is confined to genome-channel habitat rows, as the PR claims. Not asserted from reading the branch order — measured.

---

The independent reviewer was pinged twice more and has still not reported anything. Everything on this PR is my own analysis, with mutation testing and these two direct measurements standing in for a second reader.

realmarcin:
## Independent review landed — and it caught a false claim I had written into the code

This is the review I said was outstanding. It found things my own mutation testing structurally could not, and one of them was a factual error I introduced.

### The important one: my justification was false

`edge_metadata_for`'s docstring justified the habitat demotion with *"Those rows carry score 3 in the real genome archive — PREGO's lower tier."*

Measured over the first 3M rows of the real archive:

```
score distribution of HABITAT rows:
  score 4: 1,693
```

**All of them score 4 — PREGO's highest tier. Zero at score 3.**

That claim originated in the #716 issue text, which came from an earlier review comment, and I propagated it into code without checking. It's the same 3M-row sample I had already measured for its *value space* — I just never measured its *scores*. A reviewer auditing the rule against the data would have found the stated reason contradicted by the data it cites.

The rule itself survives on different grounds: Biolink `knowledge_level` is an **unordered** description of how knowledge was produced, not how confident it is. So a high-confidence observation is coherent, and "habitat ≠ annotation-derived" is the real argument. The docstring now says that, and says plainly that this is not a confidence demotion.

### C2 — my fix introduced a provenance regression

Putting the habitat check above the channel checks made it answer for channels the code knows nothing about:

```
edge_metadata_for("metagenomes", "habitat")  ->  ('observation', 'automated_agent')
                                    was ->  ('not_provided', 'not_provided')
```

Reachable, since `channel_for_archive` returns the raw stem for an unrecognised archive and those archives *are* processed — as another test in this very PR establishes. `test_unknown_channel_declines_to_assert_metadata` asserts exactly this invariant but probed only `evidence_class="unknown"`, so the hole sat right next to its own test. Now nested inside the genome branch, and the test sweeps every evidence class.

### B3 — "verified against the loaded model" verified nothing

The docstring claimed `observation` was checked against Biolink. The test hardcoded the string. **A typo — `observations` — passed the entire suite** and would have shipped ~25k edges with an invalid enum value.

New test imports `KnowledgeLevelEnum` / `AgentTypeEnum` and asserts membership, with a guard so an empty enum can't make it pass vacuously.

### Mutation-verified

| mutation | before | after |
|---|---:|---:|
| M6 — re-hoist the habitat branch | 0 failed | **1 failed** |
| M7 — typo `observations` | 0 failed | **4 failed** |

### Two vacuous assertions removed

`len(counts) <= CAP` against a 1-value fixture (passes with the cap deleted), and `flat_only < retained`, which is entailed by the two lines above it — a re-tautologized assertion *inside the test whose purpose was removing one*. Also swapped a `dict.fromkeys` for a `defaultdict`, since `run()`'s reset loop type-sniffs and would leave a plain dict as `0`, raising on a second run.

### Correction to this PR's own description

I claimed **"edge count unchanged at 18, behaviour-preserving."** It is **19**. The split *is* behaviour-preserving — core triples are identical and nothing was lost — but I also added a habitat row, and I reported the move without the addition. No test asserts a total, so nothing was papered over, but the claim was wrong as written.

### Filed, not fixed

- **#719** — habitat drift-detection is blind to dropped rows
- **#720** — the `observation` branch has no before/after coverage, which is *the* gap that let the two MEDIUMs above through my own mutation testing

`poetry run tox`: **924 passed**.

realmarcin:
## Final pass on the fixes — the strongest question answered definitively

My fixes to the review findings had not themselves been reviewed, which the review showed is not good enough. So I answered its sharpest question — *"does any (channel, evidence_class) pair now return something different, other than genome+habitat?"* — by diffing the **entire matrix against master**:

```
cells differing from master: 1 of 30
  ('annotated_genomes_isolates', 'habitat')
      master:  ('knowledge_assertion', 'automated_agent')
      this PR: ('observation', 'automated_agent')
```

Exactly one cell moves, and it is the intended one. The C2 regression is fully closed, and the literature/publication precedence and environmental branch are untouched.

That matrix is now **pinned as a test** rather than left as a one-off check. Point assertions were what let C2 through in the first place: hoisting the habitat rule changed six cells while the test probing that exact invariant kept passing, because it sampled one of the other twenty-four. Mutation-verified — re-hoisting now fails 2 tests.

### The other questions

- **C1 accuracy** — re-measured: all 1,693 habitat rows score 4. The docstring now states that and says plainly this is not a confidence demotion. No new unsupported claim.
- **B3** — the enum test can fail: mutating `observation` → `observations` fails 4 tests, where before it failed none.
- **B1/B2 deletion vs replacement** — deletion was right in both cases. B1's real coverage (the overflow path) remains and does catch a removed cap; B2's load-bearing `flat_only / retained > 0.95` remains. Only the provably-entailed lines went.

`poetry run tox`: **925 passed**.

Merging on the author's explicit instruction. Filed and left open: #715 (merge budgeting), #719, #720.

Issue #720, opened 2026-08-08
prego tests: the observation branch has no regression coverage over pre-existing rows


realmarcin:
From the independent adversarial review of #718 (finding A corollary). **LOW.**

The only habitat-evidenced row in the fixture is one **added by #718** (`tests/resources/prego/database_pairs_genomes.tsv`, the `JGI IMG / Aquatic / 4` row).

So every assertion about the `observation` knowledge_level tests a row with no "before" state. No test can catch the habitat branch flipping provenance on a row that previously emitted something else.

## Why this is the gap that mattered

This is precisely the hole that let #718's two MEDIUM findings through my own mutation testing:

- Re-hoisting the branch above the channel checks changed `("metagenomes", "habitat")` from `not_provided` to a confident `observation` — no fixture row exercised that combination.
- A typo in the enum value shipped cleanly, because nothing compared emitted values against the Biolink model.

Both are now covered by targeted tests, but the underlying fixture property still holds: **habitat rows have no before/after coverage.**

## Suggested

Add a habitat-evidenced row to a shape that already emitted before #718 — e.g. give one of the existing `Isolates` genome rows a habitat value instead of adding a new row — so a provenance flip on a pre-existing edge is detectable. The A/B harness used in the #718 review (run the transform under two fixture layouts and diff by `(subject, predicate, object)`) is the pattern worth keeping.

## Context

`edge_metadata_for` is the single point where 44.7M edges acquire their `knowledge_level` and `agent_type`. It deserves before/after coverage, not just point assertions.

realmarcin:
Closed by #722 (merged as fb025ac).

`test_emitted_provenance_snapshot` now pins channel, evidence class, `knowledge_level` and `agent_type` for all 19 fixture edges, so a provenance flip on a pre-existing edge fails rather than going unnoticed.

The 19 expected values were cross-checked against the fixture **source** rather than copied from transform output — generating a snapshot from current behaviour is how a bug becomes the expectation. Each is derivable from the archive a row lives in plus its column-6 value. Independent review re-derived all 19 and confirmed no bug was enshrined.

Mutation-verified: reverting the #713 archive split fails 13 tests; re-hoisting the habitat rule fails 2.

One extra defect surfaced while closing this: the snapshot was keyed on `(subject, object)`, so duplicate edges collapsed silently and the transform does not dedupe. A refactor processing an archive twice would have doubled 44.7M edges with every test still green. Now counts before comparing.

PR #722, opened 2026-08-08
fix(prego): drift detection over dropped rows, provenance snapshot, merge budget

### LINKED ISSUE #719: prego: habitat drift-detection is blind to dropped rows
From the independent adversarial review of #718 (finding F1). **LOW.**

`_record_habitat_value` is called from the emit-row path, so habitat values on rows dropped by shape or by threshold are never tracked.

Verified: the fixture's `Human` habitat row is a taxon-taxon pair, drops as `taxon_taxon_host`, and is silently absent from the run summary — which reports `1 rows across 1 distinct values (top: Aquatic(1))` rather than 2.

## Why it matters

The tracking added in #714 exists for **drift detection** — to make a new PREGO resource class visible instead of letting it be silently absorbed into the habitat catch-all. A new resource class that happens to land only on dropped shapes (taxon-taxon rows, inverse directions, DOID rows with no MONDO xref) would be invisible to exactly the mechanism built to catch it.

## The tension

Moving the call to classify time would cover dropped rows, but then the summary describes rows that do **not** ship, which is confusing for a report printed alongside `edges_emitted`. Options:

1. Track at classify time and label the count explicitly as "rows seen" rather than "rows emitted".
2. Track both, and report them separately.
3. Accept the blind spot and document it.

Option 2 is the most informative and barely more code — a divergence between seen and emitted is itself a useful signal.

## Not urgent

The drift this guards against is hypothetical (PREGO adding a fifth resource class), and the emit-path coverage catches the common case. Filed so the limitation is recorded rather than discovered later by someone trusting the summary to be complete.

realmarcin:
Closes #719, #720. Completes #715 with a merge-level measurement.

## #719 — drift detection was blind where it mattered

The habitat tracking added in #714 ran on the emit path, so a new PREGO resource class landing only on dropped shapes was invisible to the check built to catch it.

Demonstrated on the fixture — the `Human` habitat row drops as `taxon_taxon_host`:

```
before: habitat is a residual bucket: 1 rows across 1 distinct values (top: Aquatic(1))
after:  habitat is a residual bucket: 2 rows seen across 2 distinct values, 1 emitted (top: Aquatic(1), Human(1))
```

Values are now recorded over every well-formed row, with a separate emitted count. A divergence between seen and emitted is itself a signal, so both are reported.

## #720 — the coverage gap that let two defects through

The only habitat row in the fixture was one #718 *added*, so no test could catch the branch flipping provenance on a row that previously emitted something else. This is precisely why mutation testing missed the two MEDIUMs in #718: **mutations show a test can fail, not that a combination is untested.**

`test_emitted_provenance_snapshot` now pins channel, evidence class, `knowledge_level` and `agent_type` for all 19 fixture edges. Where `test_edge_metadata_matrix_is_pinned` covers the pure function, this covers the *wiring* — which archive a row lives in, what `channel_for_archive` returns, what `classify_evidence` makes of column 6. A bug in any of those leaves the matrix test green while changing what ships.

Mutation-verified:

| mutation | tests failed |
|---|---:|
| genome rows back in `literature.tar.gz` (the #713 defect) | **13** |
| habitat rule re-hoisted above channel checks (the C2 defect) | **2** |
| drift tracking back on the emit path (#719) | **1** |

## #715 — the merge budget, measured

PREGO alone was never the number that matters: KGX holds every source graph in the parent simultaneously. Measured across the 20 `merge.yaml` sources present on disk:

| | total input | prego | prego share |
|---|---:|---:|---:|
| today | **9.72 GiB** | 7.40 GiB | **76.2%** |
| after the #703 columns | **13.26 GiB** | 10.94 GiB | **82.5%** |
| `merge.noprego.yaml` | **2.32 GiB** | — | — |

1. Merge input grows **36.4% overall** from a change touching one source. PREGO's own +48% dilutes to +36% only because the other 19 sources together are 2.32 GiB.
2. PREGO goes from dominant to **overwhelming** — 76.2% → 82.5%. Future work on merge memory that isn't about PREGO is rounding error.
3. `merge.noprego.yaml` is a **4.2x reduction today, 5.7x after #703**. Deleting every other source combined would save less than PREGO's growth alone.

Stated explicitly in the doc: these are **input bytes, not peak RSS** — KGX's in-memory representation is several times larger and not a fixed multiple — and **no merge benchmark has been run at the new size.**

---

`poetry run tox`: **926 passed**.

### DIFF
--- a/docs/PREGO_SCORE_VALIDATION.md
+++ b/docs/PREGO_SCORE_VALIDATION.md
@@ -236,10 +236,33 @@ and `agent_type`, plus `prego_channel` changing from PREGO's verbatim column 6
 to a channel name. The raw column-6 string is not double-counted — it *moves*
 to `prego_evidence`, so it cancels.
 
-This matters for #693: KGX holds every source graph in the parent
-simultaneously, and PREGO is the single largest edge block in the merged KG. A
-48% increase on that block is material to the merge's peak RSS, and it lands on
-the same axis the two levers below are trying to reduce.
+#### What that does to the merge budget
+
+KGX holds every source graph in the parent simultaneously, so the number that
+matters for #693 is total merge input, not PREGO alone. Measured across the 20
+`merge.yaml` sources present on disk:
+
+| | total input | prego | prego share |
+|---|---:|---:|---:|
+| today | **9.72 GiB** | 7.40 GiB | **76.2%** |
+| after the #703 columns | **13.26 GiB** | 10.94 GiB | **82.5%** |
+| `merge.noprego.yaml` | **2.32 GiB** | — | — |
+
+Three things follow, and they are the actual budgeting answer:
+
+1. **Merge input grows 36.4% overall** (9.72 → 13.26 GiB) from a change that
+   touches one source. PREGO's own +48% dilutes to +36% at the merge level only
+   because the other 19 sources together are 2.32 GiB.
+2. **PREGO goes from dominant to overwhelming** — 76.2% → 82.5% of input. Any
+   future work on merge memory that is not about PREGO is rounding error.
+3. **`merge.noprego.yaml` is a 4.2x reduction today and 5.7x after #703.**
+   Nothing else available comes close: even deleting every non-PREGO source
+   would save less than PREGO's growth alone.
+
+Peak RSS is not measured here — this is input bytes, and KGX's in-memory
+representation is several times larger and not a fixed multiple. Treat these as
+the ratio to reason with, not as an RSS prediction. **No merge benchmark has
+been run at the new size.**
 
 The columns are still worth having — `knowledge_level` and `agent_type` were
 shipping empty, which made 44.7M text-mined and statistically-derived
--- a/docs/PREGO_SCORE_VALIDATION.md
+++ b/docs/PREGO_SCORE_VALIDATION.md
@@ -253,16 +253,46 @@ Three things follow, and they are the actual budgeting answer:
 1. **Merge input grows 36.4% overall** (9.72 → 13.26 GiB) from a change that
    touches one source. PREGO's own +48% dilutes to +36% at the merge level only
    because the other 19 sources together are 2.32 GiB.
-2. **PREGO goes from dominant to overwhelming** — 76.2% → 82.5% of input. Any
-   future work on merge memory that is not about PREGO is rounding error.
+2. **PREGO's share of bytes goes 76.2% → 82.5%**, so it dominates input size
+   more than before. But see the caveat below before treating that as a
+   memory ranking — it is not one.
 3. **`merge.noprego.yaml` is a 4.2x reduction today and 5.7x after #703.**
    Nothing else available comes close: even deleting every non-PREGO source
    would save less than PREGO's growth alone.
 
-Peak RSS is not measured here — this is input bytes, and KGX's in-memory
-representation is several times larger and not a fixed multiple. Treat these as
-the ratio to reason with, not as an RSS prediction. **No merge benchmark has
-been run at the new size.**
+#### Bytes and objects rank differently — do not read the table as memory
+
+Peak RSS is not measured here. This is input bytes, and KGX's in-memory
+representation is several times larger and not a fixed multiple. **No merge
+benchmark has been run at the new size.**
+
+More than a magnitude gap, though: KGX peak RSS is dominated by per-object
+Python overhead, which scales with the **number** of nodes and edges, not their
+width. On that axis the picture shifts:
+
+| source | rows | share of rows | share of bytes |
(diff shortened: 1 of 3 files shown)


realmarcin:
## Self-review found a hole I had just created

Before handing this to a reviewer I worked my own priority list, and the first question — *does the #719 move actually cover every drop reason?* — turned up a real defect.

**It did not.** The empty-id guard returns from **inside the same `try` block**, so tracking placed below it silently excluded `empty_id` rows. That is the identical blind spot #719 exists to fix, reintroduced one drop reason further along.

Tracking moved above the guard. New test exercises three distinct fates in a single run:

| habitat value | fate |
|---|---|
| `Aquatic` | emitted |
| `Human` | dropped by shape (taxon-taxon) |
| `Subglacial Lake` | dropped by the empty-id guard |

Mutation-verified: moving the call back below the guard fails 1 test.

## Measured the cost I had asserted was negligible

`classify_evidence` now runs twice per emitted row. I had called that "cheap string ops" without checking:

```
classify_evidence: 319 ns/call
extra cost at 44.7M rows: 14.3 s
```

Under 1% of a run that streams 14 GB of archives. Threading the result through would remove it at the cost of a wider signature on every emit path — not worth it at that ratio, but the number is now in the comment instead of an assumption.

## The snapshot's expectations were cross-checked, not copied

I generated the 19 expected values *from current output*, which is exactly how a bug becomes the expectation. So I re-derived each one from the fixture **source** instead — every value follows from the archive a row lives in plus its column-6 content, with no reference to what the transform emitted:

```
literature                 + PMID:*     -> publication    -> prediction / text_mining_agent
annotated_genomes_isolates + Isolates   -> resource_class -> knowledge_assertion / automated_agent
annotated_genomes_isolates + Aquatic    -> habitat        -> observation / automated_agent
environmental_samples      + "N of M samples" -> sample_count -> statistical_association / data_analysis_pipeline
```

They agree. The snapshot pins correct behaviour, not merely self-consistent behaviour.

---

`poetry run tox`: **927 passed**. An independent review has been requested and has not yet reported.

realmarcin:
## Independent review landed — two real findings, both fixed

### 1. The snapshot was blind to duplicate emission (MEDIUM)

`actual` is keyed on `(subject, object)`, so two edges sharing that pair collapse into one entry — and **the transform does not dedupe**. Reproduced:

```
edges emitted: 20 | distinct (subject,object): 19
duplicate emitted? True     <- and `actual == expected` still held
```

So a refactor that processed an archive twice would double PREGO's 44.7M edges while every prego test passed — **including this one**, written to be the tripwire. The others all use `>=`, membership, or set equality.

Now counts before comparing. Mutation-verified: `AssertionError: expected 19 edges, got 20 (duplicates?)`.

This is a good catch precisely because it's the failure mode a snapshot test is supposed to own, and mine didn't.

### 2. My conclusion 2 was a memory ranking argued from a byte measurement

It read *"any future work on merge memory that is not about PREGO is rounding error."* KGX peak RSS is dominated by per-object Python overhead, which scales with the **number** of nodes and edges, not their width. Counted rows across all 20 sources:

| source | rows | share of rows | share of bytes |
|---|---:|---:|---:|
| prego | 44,768,164 | **68.6%** | **76.2%** |
| metatraits | 4,651,069 | 7.1% | |
| bacdive | 4,516,195 | 6.9% | |
| metatraits_gtdb | 4,239,260 | 6.5% | |
| **total** | **65,258,838** | | |

Non-PREGO is **~31% of objects (20.5M)**, not ~24%, concentrated in four sources — halving metatraits + metatraits_gtdb + bacdive is a real win, not rounding error.

Sharper still: **the #703 columns add zero rows**, only ~85 B of width. In object terms PREGO stays at 68.6% before *and* after, so if RSS is object-dominated the peak grows by considerably less than the 36.4% the byte table implies — good news the earlier text wasn't claiming.

Row table added; the overreaching sentence retracted in place rather than quietly edited.

### 3. Confirmed already fixed

The review reproduced the empty-id hole against 8194a99 — I'd caught and fixed it in c2b40e6 before the review arrived. Verified against their exact probe: both the empty-id row and the taxon-taxon row are now tracked.

Also corrected a comment calling these rows "well-formed" — they parse fine and fail a *content* check, which is exactly why they need covering.

### Confirmed clean by the reviewer

The 19 snapshot values re-derived from the fixture source (all correct, no bug enshrined); the arithmetic 5 + 2 + 12 = 19; `classify_evidence` cost re-measured at 289 ns over real column-6 values (~25s across both real archives, under 1%); the `merge.yaml` / `merge.noprego.yaml` diff being exactly `{prego}`; and `evidence_habitat_emitted` surviving `run()`'s reset loop correctly.

`poetry run tox`: **927 passed**. Merging on the author's instruction.

Issue #742, opened 2026-08-11
LPSN synonym-chain walk hangs instead of failing if the cycle guard regresses


realmarcin:
Found reviewing the #684 fix (round 1), by mutation testing.

`_resolve_accepted_records` walks `record_lnk` until it reaches a correct name. Its only protection against a cyclic chain is the `link in seen` check:

```python
while True:
    link = (current.get("record_lnk") or "").strip()
    if not link or link in seen:
        break
    ...
```

Removing that check does not make the test suite fail — it makes it **hang forever**. I hit this while mutation-testing: the run had to be killed at the 40-minute timeout, and because the kill pre-empted the restore step it left the mutated file on disk.

A hanging test is a materially worse failure mode than a failing one. In CI it burns the job's whole time budget and reports a timeout rather than pointing at the defect, and locally it looks like a stuck machine.

The shipped GSS has **no cycles** — 7,018 synonyms resolve in one hop, 250 in two, 2 in three — so this is defence against a future LPSN release, not a current bug.

Fix: add a bounded iteration count alongside the `seen` set, so a cycle terminates by hitting the bound even if the set check is lost. Belt and braces on a loop whose failure mode is a hang rather than a wrong answer.

realmarcin:
Fixed in the #684 PR — the walk is now bounded (_LPSN_MAX_SYNONYM_HOPS) as well as cycle-guarded. Verified: with the seen-set removed the cyclic case completes in 2s instead of hanging.

PR #743, opened 2026-08-11
Re-point LPSN synonym matches to the accepted name (#684)

### LINKED ISSUE #684: bacdive: 992 strain→LPSN subclass_of edges point at deprecated LPSN records
Raised by the adversarial review of #680.

## Problem

`_load_lpsn_name_index()` (`kg_microbe/transform_utils/bacdive/bacdive.py:462-500`) builds its name→`record_no` index from **every** GSS row regardless of the `status` column. Only *ambiguous* names (2+ record_no for one name) are suppressed. A name with a single row that happens to be a synonym or an illegitimate name still matches and produces an edge.

## Measured impact

- 20,420 distinct `lpsn:*` records are matched across 62,096 edges.
- **468 of those 20,420 records carry `deprecated=True`**, accounting for **992 edges** (1.6%).

## Why it matters more now

#680 changes this edge from `biolink:in_taxon` to `biolink:subclass_of`. Under `in_taxon` a deprecated target was a loose association. Under `subclass_of` it places a living strain in the **subsumption hierarchy beneath a deprecated class**, which reasoners and any closure/ELK step will propagate.

This is pre-existing selection logic — #680 did not introduce it — but the predicate change raises the stakes, so it is worth deciding on deliberately.

## Suggested fix

Prefer `status == "correct name"` rows in `_load_lpsn_name_index()`, falling back to other statuses only when no correct-name row exists for that name. Alternatively, skip deprecated targets entirely for the subclass edge.

Worth confirming with a curator which behaviour is wanted before changing it — dropping 992 edges vs. re-pointing them at the accepted name are meaningfully different outcomes.

realmarcin:
Closes #684.

## The problem

`_load_lpsn_name_index` matched on name alone, so a strain whose BacDive-reported species is an LPSN **synonym** was placed under the synonym's record. #680 made that edge `subclass_of` rather than `in_taxon`, which puts a living strain in the subsumption hierarchy **beneath a deprecated class** — 992 of 62,096 edges, propagated by reasoners and any closure step.

## Drop or re-point — measuring decided it

The issue offered both and suggested asking a curator. The data answered:

- `record_lnk` is **LPSN's own crosswalk** to the accepted name
- all **7,270** links resolve to a real GSS row
- **96.2%** land directly on a correct-name row

Dropping would lose real information — the strain genuinely *is* that organism, LPSN has simply renamed it. So this follows the source's mapping rather than deleting edges. That isn't a curator judgement; it's what LPSN asserts.

Chains are walked transitively: **7,018** synonyms reach a correct name in one hop, **250** in two, **2** in three, and nothing in the shipped GSS cycles. **192** dead-end with no link and keep their own record — there is nothing better to point at.

## Effect on the real GSS

| | before | after |
|---|---:|---:|
| index entries resolving to a deprecated record | 7,225 | **49** |

Example: `Abiotrophia adiacens` `lpsn:772466` → `lpsn:776611` `Granulicatella adiacens`.

## Review round 1 — #742, fixed

Mutation testing found the walk's cycle guard was its **only** protection, and removing it **hangs rather than fails**. A hang burns a CI job's entire budget and reports a timeout instead of pointing at the defect; it also bit me directly — the kill pre-empted my restore step and left a mutated file on disk.

The walk is now bounded as well as cycle-guarded, so the behaviour survives losing either. Verified: with the `seen` set removed, the cyclic case completes in 2s instead of hanging.

## Tests

Five, covering re-point, multi-hop chain, dead-end, cycle, and the untouched correct-name case (26,866 of 34,301 rows). Mutation-checked: disabling re-pointing and truncating the chain walk each fail.

`poetry run tox` green.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -479,6 +479,7 @@ class BacDiveTransform(Transform):
             )
             return {}
         name_index: Dict[str, list] = {}
+        gss_rows: Dict[str, dict] = {}
         with open(lpsn_csv, newline="") as fh:
             reader = csv.DictReader(fh)
             for row in reader:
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review loop — 2 rounds

**Round 1 — 1 finding (#742), fixed.** Mutation testing showed the chain walk's cycle guard was its only protection and that removing it **hangs rather than fails**, which burns a CI job's whole budget and reports a timeout instead of the defect. It bit me directly: the kill pre-empted my restore step and left a mutated file on disk. The walk is now bounded as well as cycle-guarded; with the `seen` set removed the cyclic case completes in 2s.

**Round 2 — clean.** Three checks against the real GSS and the shipped `lpsn` output:

- **No orphans created.** All 27,058 distinct index targets exist as `lpsn:` nodes; re-pointing never points at something the LPSN transform does not emit.
- **Fan-in is the expected synonym merge.** 5,449 records are now referenced by more than one name, largest fan-in 21 — that is precisely what collapsing synonyms onto accepted names looks like.
- **The 49 residual deprecated targets are genuinely terminal.** 22 have no `record_lnk` at all. The other 27 have one, but their chains end at records LPSN itself offers no replacement for — `orphaned subspecies`, `rejected name; in need of a replacement`, `illegitimate name; in need of a replacement`, each with an empty `record_lnk`. Keeping the original rather than hopping to another dead record is the right call.

Down from 7,225 to 49 (99.3% resolved), with the remainder being cases the source cannot resolve either.

Issue #744, opened 2026-08-12
The MicrobeDecoder BacDive crosswalk is a type-strain relation, but nothing expresses that


realmarcin:
Deferred from #687.

MicrobeDecoder's `BacDive_ID` column is not an identifier equivalence like its `NCBI_Taxonomy_ID` / `GTDB_ID` siblings. It names the **nomenclatural type strain** of the LPSN record, and the source proves it: over the 21,247 rows matching a GSS record, MicrobeDecoder's strain designation

- **exactly equals** LPSN's own `nomenclatural_type` for **21,010 (98.9%)**
- shares ≥1 culture-collection number for another **233 (1.1%)**
- has no overlap for **4**

That is what makes the relation strictly 1:1 (one type strain per name) where bacdive's is many-to-one (up to 1,870 strains per record).

#687 removed the contradiction by emitting the same `strain -subclass_of-> lpsn` bacdive already asserts, so the two transforms agree and merge collapses the 18,425 shared pairs while keeping both provenances. **But that loses the type-strain specificity** — subsumption says the strain belongs to the name, not that it is *the* type.

Nothing in the graph can currently say it:

- **Biolink** has no type-strain or nomenclatural-type predicate (checked `type of`, `has nomenclatural type` against BMT).
- **METPO** has no such term either — `TYPE_STRAIN` in `constants.py` is a TSV *column* name in the bacdive intermediate, not a predicate.
- The `lpsn` transform parses `nomenclatural_type` into culture-collection CURIEs, but never asserts the relation.

Options: propose a METPO predicate (`.claude/skills/metpo-proposal` covers the workflow), or adopt an RO term if one fits. Worth doing — LPSN, BacDive and MicrobeDecoder all carry the information and the graph currently drops it on the floor.

Scope note: `GOLD_Organism_ID` and `IMG_Genome_ID` have the same shape problem — an organism/genome is not another name for a taxon — but neither has a contradicting edge today, so they were left alone.

PR #745, opened 2026-08-12
microbedecoder: assert subsumption, not close_match, on the BacDive crosswalk (#687)

### LINKED ISSUE #687: microbedecoder: 19,114 orphan close_match edges to bacdive:N, and a modeling clash with bacdive's strain→LPSN edges
Found while validating the predicate/prefix change in #680.

## 1. Same orphan bug as bacdive had

`microbedecoder` emits **19,114** edges of the form:

```
lpsn:10084  biolink:close_match  bacdive:166782   (skos:closeMatch)
```

`bacdive:N` is **never emitted as a node row by any transform** — verified across all of `data/transformed/*/nodes.tsv`. In the merged KG all 62,521 `bacdive:*` nodes are KGX-synthesized stubs (`provided_by=Graph`, `biolink:NamedThing`, no name); **zero** have a real source.

Meanwhile the real strain nodes exist under a different CURIE: 99,392 `kgmicrobe.strain:bacdive_N` rows. All **19,079** distinct microbedecoder BacDive IDs resolve there (0 misses), so this is a pure prefix mismatch, same class as the bacdive bug fixed in #680.

## 2. But a naive retarget creates a semantic contradiction

Retargeting to `kgmicrobe.strain:bacdive_N` would resolve the orphans, but **18,425 (lpsn, bacdive) pairs are asserted by both transforms**. After #680 those would become:

```
lpsn:X                        --close_match--> kgmicrobe.strain:bacdive_Y   (microbedecoder)
kgmicrobe.strain:bacdive_Y    --subclass_of--> lpsn:X                       (bacdive)
```

`skos:closeMatch` asserts near-identity; `subclass_of` asserts proper subsumption. Asserting both over the same node pair is contradictory.

## 3. Cardinality shows which one is right

| transform | pairs | distinct LPSN records | shape |
|---|---:|---:|---|
| bacdive | 62,096 | 20,420 | **many-to-one** — up to **1,870 strains** on a single LPSN record |
| microbedecoder | 19,114 | 19,114 | **strictly 1:1** |

- BacDive's relation is many strains per species name, so identity semantics are impossible and `subclass_of` is correct (as changed in #680).
- microbedecoder's is a strict 1:1 crosswalk — almost certainly one representative/type strain per LPSN name — so `close_match` is defensible *there*, but it is not the same relation bacdive is expressing.

## Suggested resolution

Not a blind prefix swap. Options:

1. **Retarget + re-predicate**: point microbedecoder at `kgmicrobe.strain:bacdive_N` and change the predicate to something that does not collide with subsumption — e.g. a type-strain relation, or drop the edge where bacdive already asserts `subclass_of` for the same pair.
2. **Keep close_match but target the right entity**: if the crosswalk really means "this LPSN name ↔ this BacDive record", the object should be whatever node represents the *record*, which today is nothing. That argues for option 1.
3. Confirm with a curator whether the microbedecoder crosswalk is type-strain designation; if so, model it explicitly rather than as `close_match`.

Blocked on a modeling decision, so deliberately not folded into #680.

realmarcin:
Closes #687. Defers the type-strain modelling to #744.

## State of the issue

**Orphans — already fixed.** `CROSSWALK_COLUMNS` emits `kgmicrobe.strain:bacdive_N`; the shipped output has **0** bare `bacdive:` targets.

**Semantic clash — live.** With the prefix fixed, the contradiction the issue predicted now exists:

| | count |
|---|---:|
| microbedecoder `lpsn:X -close_match-> strain:Y` | 19,114 |
| bacdive `strain:Y -subclass_of-> lpsn:X` | 62,096 |
| **same pair asserted both ways** | **18,425** |

## The modelling question, answered by the source

The issue said this needed a curator. It doesn't — LPSN's own GSS settles it. Comparing MicrobeDecoder's strain designation against GSS `nomenclatural_type` for the same `LPSN_ID`, over the 21,247 rows that match a GSS record:

| | rows | share |
|---|---:|---:|
| **exactly equals** `nomenclatural_type` | 21,010 | **98.9%** |
| shares ≥1 culture-collection number | 233 | 1.1% |
| no overlap | 4 | 0.0% |

So the column means **"this BacDive record is the type strain of this LPSN name"** — which is why it is strictly 1:1 where bacdive's is many-to-one (up to 1,870 strains per record).

## What this does, and what it deliberately doesn't

That rules `close_match` out — the target is a *strain*, not another name for the taxon. It does not pick a replacement, because **no predicate expresses "is the type strain of"**: BMT has neither `type of` nor `has nomenclatural type`, and METPO's `TYPE_STRAIN` is a TSV column name in the bacdive intermediate, not a predicate.

Rather than invent an ontology term mid-fix, this emits the **same subsumption bacdive already asserts**, so the two transforms agree instead of contradicting. Merge collapses the 18,425 shared pairs keeping both provenances; **689** are unique to microbedecoder and become new subsumption edges.

The type-strain specificity is **deferred to #744**, not dropped silently.

**Only the BacDive column changes.** NCBITaxon, GTDB, GOLD and IMG are genuine identifier crosswalks where `close_match` is correct — a test pins that they're untouched, and a mutation re-predicating all five fails it.

## Verification

Canary on the real 27,010-row source: 84,244 crosswalk edges, **0** close_match to a strain, **19,114** subsumption edges, 18,425 agreeing with bacdive.

Three pre-existing tests assumed the BacDive edge was subject-side `lpsn:` and were updated rather than bypassed. Two mutations checked. `poetry run tox` green, 25 tests.

### DIFF
--- a/kg_microbe/transform_utils/microbedecoder/microbedecoder.py
+++ b/kg_microbe/transform_utils/microbedecoder/microbedecoder.py
@@ -79,14 +79,17 @@ from kg_microbe.transform_utils.constants import (
     PRIMARY_KNOWLEDGE_SOURCE_COLUMN,
     PRODUCES_PREDICATE,
     PROVIDED_BY_COLUMN,
+    RDFS_SUBCLASS_OF,
     RELATION_COLUMN,
     SMALL_MOLECULE_CATEGORY,
+    SUBCLASS_PREDICATE,
     SUBJECT_COLUMN,
     TRAIT_PREFIX,
     TROPHICALLY_INTERACTS_WITH,
     VPI_KNOWLEDGE_SOURCE,
 )
 from kg_microbe.transform_utils.microbedecoder.utils import (
+    BACDIVE_CROSSWALK_COLUMN,
     BACDIVE_SNAPSHOT_COLUMNS,
     CROSSWALK_COLUMNS,
     LPSN_ID_COLUMN,
@@ -348,15 +351,29 @@ class MicrobeDecoderTransform(Transform):
                         local_id = local_id[len(candidate) :]
                         break
                 object_curie = f"{prefix}{local_id}"
-                edge_writer.writerow(
-                    self._make_edge_row(
-                        subject,
-                        CLOSE_MATCH_PREDICATE,
-                        object_curie,
-                        CLOSE_MATCH_RELATION,
-                        self.knowledge_source,
+                if column == BACDIVE_CROSSWALK_COLUMN:
+                    # Strain -> name subsumption, matching what the bacdive
+                    # transform asserts for the same pair. See the note on
+                    # BACDIVE_CROSSWALK_COLUMN: close_match contradicted it.
+                    edge_writer.writerow(
+                        self._make_edge_row(
+                            object_curie,
+                            SUBCLASS_PREDICATE,
+                            subject,
+                            RDFS_SUBCLASS_OF,
+                            self.knowledge_source,
+                        )
                     )
-                )
+                else:
+                    edge_writer.writerow(
+                        self._make_edge_row(
+                            subject,
+                            CLOSE_MATCH_PREDICATE,
+                            object_curie,
+                            CLOSE_MATCH_RELATION,
+                            self.knowledge_source,
+                        )
+                    )
                 self._stats["crosswalk_edges"] += 1
 
     # ------------------------------------------------------------------
--- a/kg_microbe/transform_utils/microbedecoder/utils.py
+++ b/kg_microbe/transform_utils/microbedecoder/utils.py
@@ -41,6 +41,24 @@ CROSSWALK_COLUMNS: Tuple[Tuple[str, str, Optional[str]], ...] = (
     ("IMG_Genome_ID", "IMG:", None),
 )
 
+# The BacDive crosswalk is not an identifier equivalence like the others. Its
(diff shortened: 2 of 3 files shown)


realmarcin:
## Review loop — 2 rounds

**Round 1 — clean.** All 19,079 distinct strain subjects exist in `bacdive/nodes.tsv`; **zero** would orphan. The flip preserves `knowledge_level`/`agent_type` (`knowledge_assertion`/`manual_agent` on all 19,114) and `relation` (`rdfs:subClassOf`). Orphan risk is unchanged by the flip — the same two endpoints, swapped.

**Round 2 — one finding, filed as #746, deliberately not fixed here.**

Making both transforms use `subclass_of` exposed **239 strains where they name different LPSN parents** — a conflict that could not exist while the predicates differed, so this PR surfaces it rather than causes it.

**158 of the 239 resolve** if microbedecoder's LPSN record is re-pointed through the same accepted-name walk #684 added to bacdive; microbedecoder has 5,263 edges (1.0%) on deprecated LPSN records and was never given that treatment. The remaining **81** are genuine snapshot-vs-current differences between MicrobeDecoder's 2024-11-07 data and BacDive's.

Scoped out because it is a distinct defect with its own measurement, and fixing it here would mean moving `_resolve_accepted_records` off `BacDiveTransform` into shared code — a refactor that deserves its own review.

PR #747, opened 2026-08-12
Surface corrections that lose a rank tie to a raw column (#723)

### LINKED ISSUE #723: best_primary ranks raw source columns above corrected ones, so upstream corrections lose ties — 35 rows diverge, held back only by a hand-maintained retraction list
Filed from the MediaIngredientMech side as the consumer half of CultureBotAI/MediaIngredientMech#138, which covers the source-data half. That issue notes "no action needed here" on the kg-microbe side because the retraction pass keeps the graph clean — this issue is about the mechanism that makes the retraction pass necessary, and what it costs.

**The published graph is not currently wrong.** All three known-bad groundings are on the retraction list and are dropped every consolidation. This is a design/maintainability issue, not a live data defect.

## The mechanism

`scripts/consolidate_chemical_mappings.py:593`:

```python
def best_primary(candidates) -> str:
    ...
    for c in candidates:
        r = prefix_rank(c)
        if r > best_rank:       # strictly greater -> first candidate wins a tie
```

The tiebreak is documented and intentional ("Within a tied rank, the first candidate in iteration order wins — callers that care about a deterministic tiebreak should pass candidates in their preferred order"). The issue is the order the caller actually passes, at line 1219:

```python
primary_id = best_primary([chebi_id, culturemech_term, mim_id, kg_node_id, cas_curie])
```

`CHEBI`, `FOODON`, `UBERON`, `ENVO`, `NCIT` all rank **100**, so within that tier the tiebreak decides everything — and it puts the two **raw upstream columns** (`chebi_id`, `culturemech_term_id`) ahead of the two **corrected** ones (`mim_id`, `kg_microbe_node_id`).

Consequence: when a curator corrects a grounding by writing the new id to `mim_id` / `kg_microbe_node_id` but the stale value survives in `chebi_id` / `culturemech_term_id`, **the consolidator grounds to the stale value and demotes the correction to an xref.** The correction is present in the file and silently loses.

## Measured against the current `UNIFIED_INGREDIENT_MAPPING.tsv` (3915 rows)

Reordering to `[kg_node_id, mim_id, chebi_id, culturemech_term, cas_curie]` changes the primary on **35 rows** — matching the "~35" estimate in #138. Splitting them by whether the current primary is already on `mappings/kgm_retracted_groundings.tsv`:

- **5 rows: known-wrong**, already retracted every run — `4-Aminobenzoic acid`, `p-Amino Benzoic Acid`, `p-amino benzoic acid` (`CHEBI:194474` → `CHEBI:30753`), `Cobalamine` (`CHEBI:30411` → `CHEBI:28911`), `Infusion from Potatoes` (`FOODON:03316428` *Peptone* → unmapped)
- **30 rows: unreviewed** — nobody has ruled on whether the raw or the corrected column is right. A sample:

| ingredient | current primary | corrected columns say |
|---|---|---|
| Cholesterol / cholesterol | `CHEBI:140435` | `CHEBI:16113` |
| Fructose | `CHEBI:28757` | `CHEBI:28645` |
| Casein | `FOODON:03420180` | `CHEBI:3448` |
| Thiamin | `CHEBI:18385` | `CHEBI:26948` |
| Beef heart | `UBERON:0000948` | `FOODON:00004410` |
| Iron(II) sulfate | `CHEBI:29033` | `CHEBI:75832` |
| 8× malate variants | `CHEBI:91260` | `CHEBI:91261` |

I am not claiming the right-hand column is correct in these 30 — several look like genuine judgement calls (`Casein` as CHEBI protein vs FOODON food; `Beef heart` as anatomy vs food). That is exactly the point: **the current behaviour resolves 30 open curation questions by column position, silently, and nobody chose that.**

## Why the retraction list is not a fix

`apply_retractions` is well-built — conservative, only drops an object whose sole contributing source is the retracting one, runs before `propagate_synonyms_via_xrefs` so phantom xrefs cannot leak labels. It works.

But it is a **hand-maintained 179-row list that must grow by one entry for every future upstream correction**, and it only fires for corrections someone noticed and enumerated. A correction made upstream that nobody adds to the list is silently reverted to the stale grounding, with no signal. The mitigation scales with curator vigilance rather than with the data.

## Options, in increasing order of commitment

1. **Make it visible without changing behaviour** (cheapest, and probably the right first step). Emit a warning at load when a row's corrected columns disagree with the chosen primary — the 35 above become a report instead of silence. That closes the "silently reverted with no signal" hole without touching any grounding.
2. **Reorder the candidate list** to `[kg_node_id, mim_id, chebi_id, culturemech_term, cas_curie]`. Fixes the class outright and lets 5 retraction entries retire — but shifts 30 unreviewed groundings, so it should follow the curation review, not precede it.
3. **Fix it at the source** so each row has one consistent value — CultureBotAI/MediaIngredientMech#138, which is where the stale columns come from. Even with that done, the precedence rule stays latent for the next tool that writes a correction to only one column, so option 1 is still worth having.

Happy to supply the full 35-row table or the script that produced it.

Related: CultureBotAI/MediaIngredientMech#138 (source-data half).

realmarcin:
Closes #723.

## What and why

`best_primary` breaks a rank tie by candidate order, and the call site passes the raw upstream columns (`chebi_id`, `culturemech_term_id`) **ahead of** the corrected ones (`mim_id`, `kg_microbe_node_id`). CHEBI, FOODON, UBERON, ENVO and NCIT all rank 100, so inside that tier order decides everything — a curator who writes a correction to a corrected column while a stale value survives in a raw one is **silently overruled**, their correction demoted to an xref.

## Report, don't resolve — the issue's own option 1

Reordering the candidates fixes the class outright and would let 5 retraction entries retire. But it moves **35** groundings, of which only 5 are known-wrong; the other **30 are unreviewed curation questions** — `Casein` as a CHEBI protein or a FOODON food, `Beef heart` as anatomy or food.

Resolving those by column position is *the defect being reported*. So this makes the divergence visible and leaves the call with a curator. **No emitted mapping changes.**

## Verification

Against the real 3,915-row `UNIFIED_INGREDIENT_MAPPING.tsv`: **exactly 35 rows flagged**, matching both the issue's count and its named examples.

| ingredient | grounds to | corrected column says |
|---|---|---|
| `Beef heart` | `UBERON:0000948` | `FOODON:00004410` |
| `4-Aminobenzoic acid` | `CHEBI:194474` | `CHEBI:30753` |
| `Fructose` | `CHEBI:28757` | `CHEBI:28645` |
| `CoSO4 x 7 H2O` | `CHEBI:91244` | `CHEBI:53470` |

The check is extracted as `corrected_column_primary` so it is unit-testable rather than buried in the loader. Four tests, mutation-checked (making the corrected order equal the production order fails them). One test pins that **ranking still beats order** — this reports tie-breaks, it does not invert the ranking.

`poetry run tox` green.

## Not addressed here

The upstream half is CultureBotAI/MediaIngredientMech#138 — each row should carry one consistent value. Even once that lands, the precedence rule stays latent for the next tool that writes a correction to only one column, so this report is worth having regardless.

### DIFF
--- a/scripts/consolidate_chemical_mappings.py
+++ b/scripts/consolidate_chemical_mappings.py
@@ -612,6 +612,32 @@ def best_primary(candidates) -> str:
     return best
 
 
+def corrected_column_primary(kg_node_id, mim_id, chebi_id, culturemech_term, cas_curie) -> str:
+    """
+    Return the primary that would be chosen if corrected columns outranked raw ones.
+
+    Same candidates and same ranking as the production call, only the order
+    differs: ``kg_microbe_node_id`` and ``mim_id`` are the columns a curator
+    writes a correction into, while ``chebi_id`` and ``culturemech_term_id``
+    carry the raw upstream value. CHEBI, FOODON, UBERON, ENVO and NCIT all rank
+    100, so within that tier order alone decides, and production puts the raw
+    columns first (#723).
+
+    Comparing the two orders is what makes a silently-overruled correction
+    visible. It deliberately does not change the grounding: 30 of the 35 rows it
+    flags are unreviewed curation questions, and resolving those by column
+    position is the defect, not the fix.
+
+    :param kg_node_id: ``kg_microbe_node_id`` cell.
+    :param mim_id: ``mim_id`` cell.
+    :param chebi_id: ``chebi_id`` cell.
+    :param culturemech_term: ``culturemech_term_id`` cell.
+    :param cas_curie: CAS CURIE, last-resort candidate in both orders.
+    :return: The CURIE a corrected-first order would pick, or ``''``.
+    """
+    return best_primary([kg_node_id, mim_id, chebi_id, culturemech_term, cas_curie])
+
+
 def is_chemical_curie(curie: str) -> bool:
     """Return True if the CURIE's prefix belongs in the unified chemical file."""
     if not curie or ":" not in curie:
@@ -1198,7 +1224,10 @@ class ChemicalMappingConsolidator:
         df = pd.read_csv(filepath, sep="\t", dtype=str).fillna("")
 
         added = 0
+
         skipped = 0
+
+        diverged: list = []
         for _, row in df.iterrows():
             ingredient_name = row.get("ingredient_name", "").strip()
             cas_rn = row.get("cas_rn", "").strip()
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review round 1 — no defects; the check is precise

Against the real 3,915-row file:

| | |
|---|---:|
| rows flagged | **35** of 3,915 (0.89%) |
| flagged with **no** corrected column populated (false positives) | **0** |
| flagged where both CURIEs share a rank (a genuine tie, not a rank difference) | **35 of 35** |

So every flag is a real tie-break between a populated corrected column and a raw one — the exact class the issue describes, with no noise for a curator to wade through.

## One deliberate difference from the issue's split

The issue splits the 35 into 5 already-retracted and 30 unreviewed. I measure **4 and 31**.

The fifth is `Infusion from Potatoes`, whose correction is to **unmap** (`FOODON:03316428` *Peptone* → nothing). This check requires both orders to yield a CURIE, so a row whose corrected columns are empty is not reported: an absent corrected column is silence, not disagreement, and treating it as a correction would flag every uncorrected row in the file.

That case is real but a different signal — "the raw grounding is wrong and there is no replacement" — and it is already handled by the retraction list. Worth knowing the report will not surface it.

PR #748, opened 2026-08-12
microbedecoder: anchor on accepted LPSN names, sharing the resolver (#746)

### LINKED ISSUE #746: microbedecoder attaches 5,263 edges to deprecated LPSN records (same class as #684)
Found reviewing #745 (round 2).

#684 fixed this for bacdive: names that resolve to an LPSN *synonym* are now re-pointed to the accepted record via LPSN's own `record_lnk`. microbedecoder has the same problem and was not touched.

| | |
|---|---:|
| microbedecoder edges | 514,641 |
| distinct `lpsn:` endpoints | 26,170 |
| of those, deprecated | **252** |
| **edges touching a deprecated record** | **5,263 (1.0%)** |

microbedecoder anchors *everything* — metabolism, BacDive snapshot, crosswalk — to `lpsn:<LPSN_ID>` taken straight from the source's `LPSN_ID` column, so a synonym ID drags every edge for that row onto a deprecated node.

## Why it surfaced now

#745 makes microbedecoder assert `strain -subclass_of-> lpsn`, the same relation bacdive asserts. That exposed **239 strains where the two transforms name different LPSN parents** — impossible to see while the predicates differed.

**158 of the 239 disagreements disappear** if microbedecoder's LPSN record is re-pointed through the same `_resolve_accepted_records` walk #684 added. Example: strain `bacdive_17241`, microbedecoder says `lpsn:780735`, bacdive says `lpsn:782997` which re-points to `lpsn:780733`.

The remaining **81** are genuine differences between MicrobeDecoder's 2024-11-07 snapshot and BacDive's current data, not something a re-point fixes.

## Suggested fix

Share the resolution. `_resolve_accepted_records` currently lives on `BacDiveTransform`; it takes a `{record_no: gss_row}` mapping and returns `{record_no: accepted_record_no}`, so it has no BacDive-specific state and could move to a shared LPSN utility that both transforms call.

Deliberately out of scope for #745, which is about the crosswalk predicate.

realmarcin:
Closes #746.

## The problem

#684 fixed this for bacdive. microbedecoder never got it, and had **5,263 edges (1.0%)** on deprecated LPSN records — worse per-row than bacdive's case, because it anchors *everything* (crosswalk, metabolism, BacDive snapshot) on a single `lpsn:<LPSN_ID>` taken straight from the source column. One synonym ID drags an entire row's edges onto a deprecated class.

## Shared, not duplicated

`_resolve_accepted_records` held no BacDive-specific state, so it moved to `kg_microbe/utils/lpsn_utils.py`. Both transforms now call the same walk — which is the point: they anchor on the same records and should resolve them identically.

## Effect

| | before | after |
|---|---:|---:|
| edges touching a deprecated LPSN record | 5,263 | **269** (−94.9%) |
| synonym records re-pointed | — | 7,243 |

## A measurement trap worth recording

Comparing against the shipped `data/transformed/bacdive/edges.tsv` shows parent conflicts **rising 239 → 436**, which looks like a regression.

It isn't. **That file predates #684**, so bacdive's on-disk parents are not re-pointed, and the comparison pits one transform's new behaviour against another's old output. Applying #684's resolution to both sides gives **89** — down from 239, close to the 81 predicted in the issue. The full benefit needs a bacdive re-run.

## Contract preserved

A missing `lpsn_gss.csv` stays non-fatal — the GSS export is account-gated and not shipped — matching bacdive. IDs are then used as given and the run prints why.

## Verification

Canary on the real 27,010-row source. Nine tests: seven on the shared resolver (synonym, chain, correct-name, dead-end, cycle, over-length chain, dangling link) and two on the transform. Two mutation-checked. bacdive's 15 existing LPSN tests pass unchanged against the moved resolver. `poetry run tox` green.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -487,6 +487,13 @@ class BacDiveTransform(Transform):
                 record_no = (row.get("record_no") or "").strip()
                 if not record_no:
                     continue
+                # Collected before the name check so the resolver sees every
+                # record, exactly as microbedecoder passes it. Both callers share
+                # one walk (#746); feeding it different subsets would let a chain
+                # hopping through a nameless record resolve for one transform and
+                # dead-end for the other. No GSS row is nameless today, so this
+                # is fragility rather than a live difference.
+                gss_rows[record_no] = row
                 genus = (row.get("genus_name") or "").strip()
                 sp = (row.get("sp_epithet") or "").strip()
                 subsp = (row.get("subsp_epithet") or "").strip()
@@ -496,7 +503,6 @@ class BacDiveTransform(Transform):
                 if not full_name:
                     continue
                 name_index.setdefault(full_name, []).append(record_no)
-                gss_rows[record_no] = row
         accepted = resolve_accepted_records(gss_rows)
         repointed = 0
         for full_name, record_nos in name_index.items():
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -187,6 +187,7 @@ from kg_microbe.utils.isolation_source_mapping_utils import (
     load_isolation_source_mappings,
     normalize_isolation_source_label,
 )
+from kg_microbe.utils.lpsn_utils import resolve_accepted_records
 from kg_microbe.utils.mapping_file_utils import (
     _build_metpo_tree,
     load_assay_kit_mappings,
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review round 1 — one finding, fixed in-branch

**Both callers now feed the shared resolver the same thing.** bacdive collected `gss_rows` *after* a `continue` for rows with no assembled name; microbedecoder passes every row with a `record_no`. Two callers, one walk, different inputs — a chain hopping through a nameless record would resolve for one transform and dead-end for the other.

Measured: **zero impact today** — all 34,301 GSS rows have a name, both inputs produce the identical 7,243-entry map, 0 records resolve differently. So it was latent fragility, not a live bug, and the fix is one line moved. bacdive's index is unchanged (34,246 names; `Abiotrophia adiacens` → `lpsn:776611`).

**Also checked and clean:** the two accepted maps converge — 0 of bacdive's index targets would be moved further by microbedecoder's map, i.e. both paths reach the same fixed point.

Issue #755, opened 2026-08-12
METPO pairing test: stem heuristic misses multi-word predicates, incl. the canonical grows-in counter-example


realmarcin:
Found in adversarial review of the re-landed #751 work, against my own test.

`test_does_not_stem_siblings_are_paired` derives a positive predicate's stem
with `label.lower().rstrip().removesuffix("s")`. That only works when the verb
is a single word ending in a bare `s`. Measured against real METPO predicate
shapes:

| positive | derived stem | negative | stem | result |
|---|---|---|---|---|
| `tolerates` | `tolerate` | `does not tolerate` | `tolerate` | match |
| `produces` | `produce` | `does not produce` | `produce` | match |
| `grows in` | `grows in` | `does not grow in` | `grow in` | **MISSED** |
| `shows activity of` | `shows activity of` | `does not show activity of` | `show activity of` | **MISSED** |
| `denitrifies` | `denitrifie` | `does not denitrify` | `denitrify` | **MISSED** |

3 of 5 miss, and the miss is silent: no sibling is found, so the loop `continue`s
and the assertion passes. The test reports green while checking nothing.

The worst case is `grows in` / `does not grow in` — **METPO:2000517/2000518**,
the exact unpaired pair the skill documents as its counter-example and the
motivating case for the whole synonyms-column fix. A test written to prevent
that defect cannot currently detect it.

This is false confidence rather than a missing check, which makes it worse than
having no test: the other three assertions are sound, so the suite looks like it
covers pairing.

### Ask

Inflect the **first word** and match on a candidate set rather than a single
`removesuffix("s")`:

- `-ies` -> `-y` (`denitrifies` -> `denitrify`)
- `-es` -> strip `es` and strip `e` (`oxidizes` -> `oxidize`)
- `-s` -> strip `s` (`grows` -> `grow`)

then re-join the remaining words. Verify by asserting the heuristic pairs all
five shapes above, so the table itself becomes the regression test.

Issue #753, opened 2026-08-12
METPO proposal: intra-proposal pattern-key clashes are unguarded


realmarcin:
Found reviewing #751. Low severity, latent.

`validate_against_metpo` checks each proposed label/synonym against *existing*
METPO, but nothing checks proposed terms against **each other**.

`_build_metpo_lookups` keys `metpo_pattern_to_predicate` on every synonym plus
the label, and writes last-wins:

```python
self.metpo_pattern_to_predicate[pattern_key]["positive"] = curie
```

So two proposed properties of the same polarity sharing a synonym silently
collapse: one predicate wins the key, the other becomes unreachable through
that pattern, with no warning at generation or load time.

There is no clash today (verified across all 9 property rows). The reason to
file it now is that #751 makes property synonyms actually reach the
submittable artifact for the first time, so the surface exists where it
previously did not — every property synonym is now a live pattern key.

### Ask

In `validate_against_metpo` (or alongside it), fail when two proposed terms of
the same polarity normalise to the same key. Cheap to fold into the pairing
test proposed in the sibling issue.

realmarcin:
Fixed on master by `105615fe` (PR #759), which did not carry a closing keyword — hence this issue outliving its fix.

Verified on `859a1853`. `tests/test_extract_metpo_proposals.py::test_no_same_polarity_pattern_key_clash` builds the same `(pattern key, polarity)` map that `_build_metpo_lookups` writes last-wins into, and fails if two proposed properties of the same polarity claim one key.

That is the latent collapse this issue described, now caught at test time rather than becoming an unreachable predicate with no warning at generation or load. Still no clash across the 9 property rows, which is the point — the guard exists before the proposal grows into one.

Issue #752, opened 2026-08-12
METPO proposal: no test asserts that a proposed paired predicate actually pairs


realmarcin:
Found reviewing #751.

#749 was a defect where the properties ROBOT template had no `synonyms`
column, so `Term.synonyms` was silently dropped for every property and no
proposed predicate could auto-pair via `_build_metpo_lookups`. It shipped and
survived because **nothing in the test suite could see it**.

`tests/test_extract_metpo_proposals.py` is a regenerate-and-diff gate. It
compares the generator's output against the committed artifact, so it would
catch someone *deleting* the synonyms column (the artifact would change). It
cannot catch the case that actually matters: a curator proposes a new
`X` / `does not X` pair and forgets the shared related synonym. Output is
self-consistent, the gate is green, and the pair is silently unpaired —
`_get_negative_predicate` returns `None` and false-majority observations are
dropped downstream with no error.

The skill (`.claude/skills/metpo-proposal`) already names this check:

> verify by simulating `_build_metpo_lookups` against the proposal OWL

but it is a manual step in a checklist, which is why it was not run.

### Ask

Add a test that reimplements the loader's pairing rule (label lowercased,
`does not ` prefix ⇒ negative; key on label + every synonym) against the
emitted properties template, and asserts:

1. Every property whose label starts with `does not ` has a positive partner
   reachable through a shared key — i.e. no orphan negatives.
2. Symmetrically, any positive whose stem has a `does not <stem>` sibling in
   the proposal is paired with it.

The loader reads `node["meta"]["synonyms"]` — all synonym types, not just
`hasRelatedSynonym` — so the test should key on the synonyms column
regardless of which annotation directive it carries.

realmarcin:
Fixed on master by `105615fe` (PR #759), which did not carry a closing keyword — hence this issue outliving its fix.

Verified on `859a1853`. `tests/test_extract_metpo_proposals.py` is no longer only a regenerate-and-diff gate; it now has three tests that fail on the case this issue described — output self-consistent, pairing broken:

- `test_real_predicate_shapes_are_paired` — each positive must generate the stem its `does not` partner uses, checked against real shapes including the canonical `grows in` / `does not grow in` (METPO:2000517/2000518)
- `test_no_orphan_negative_predicates` — a `does not X` with no reachable positive fails
- `test_does_not_stem_siblings_are_paired` — covers the multi-word case that #755 found the original stem heuristic missing

[...]

Issue #750, opened 2026-08-12
7 of 9 rows in the METPO properties proposal still carry TODO:add_citation


realmarcin:
Noticed while adding METPO:2000802 (#744).

`mappings/metpo_proposal_properties_robot.tsv`:

| row | definition_source |
|---|---|
| METPO:2000064 `tolerates` | `TODO:add_citation` |
| METPO:2000065 `does not tolerate` | `TODO:add_citation` |
| METPO:2000067 `isolated from host with quality` | `TODO:add_citation` |
| METPO:2000068 `isolated from environment with quality` | `TODO:add_citation` |
| METPO:2000717 `has growth temperature optimum value` | `TODO:add_citation` |
| METPO:2000718 `has growth pH optimum value` | `TODO:add_citation` |
| METPO:2000719 `has growth salinity optimum value` | `TODO:add_citation` |
| METPO:2000069 `has environmental niche` | `DOI:10.3390/microorganisms10020293` |
| METPO:2000802 `is nomenclatural type of` | `PMID:41797679` |

The classes template has 37 more.

The `metpo-proposal` skill is explicit: *"A proposal with only `TODO:add_citation` placeholders should NOT be submitted upstream"*, and `IAO:0000119` is listed as required on every term. These are all from the `metpo_proposal_2026_04` cohort; the two later cohorts carry real citations.

Not blocking the 2026_08 cohorts, which can be filed on their own subset tag — but the 2026_04 cohort cannot go upstream until these are filled, and it is worth knowing that before someone attaches the whole file to a `berkeleybop/metpo` issue.

PR #756, opened 2026-08-12
feat(metpo): propose "is nomenclatural type of" (METPO:2000802) and fix silently-dropped property synonyms

realmarcin:
Re-lands #751, which was reverted in #754 for a process reason — merged with
`--admin` over the required-review rule — not a technical one. Stacked on the
revert branch so this diff shows the work rather than a net-zero. GitHub will
retarget it to `master` when #754 merges.

**Please review before merging. I will not merge this one myself.**

## 1. `METPO:2000802 is nomenclatural type of` (#744)

#746 established that MicrobeDecoder's BacDive crosswalk column is a
*type-strain* relation: 21,010 of 21,247 rows (98.9%) carry a strain
designation byte-identical to the LPSN GSS `nomenclatural_type` for the same
`LPSN_ID`. We currently emit `subclass_of`, which is not what it means, and
neither METPO nor RO has a predicate for it.

- **ID**: opens a `METPO:20008xx` nomenclature family. The documented free
  slots (`2000066`, `2000070`) are all in the chemical-interaction range,
  which is the wrong neighbourhood; the skill's rule is "match the family",
  and no matching family existed, so this creates one.
- **Citation**: `PMID:41797679`, the ICNP 2025 Revision — looked up, not guessed.
- **Range** `owl:Thing`, following METPO:2000069: METPO has no taxon-name class.
- **No `does not` partner**: absence of a type-strain assertion is not a claim
  that a strain is *not* the type, so the negative form would carry no
  observations.

## 2. The properties template dropped every synonym (#749)

The metatraits loader pairs a positive predicate with its `does not <stem>`
partner via a shared synonym. The properties ROBOT template — the artifact
submitted upstream — **had no synonyms column**, so `Term.synonyms` was
accepted by the dataclass and silently discarded for every property the script
has ever emitted. No proposed predicate could auto-pair.

Already biting: `METPO:2000064`/`2000065` (`tolerates`/`does not tolerate`)
carry the shared `tolerance` synonym in the curation TSV and lost it on the way
to the submittable one, reproducing upstream's own broken `2000517`/`2000518`.

## 3. Tests, because this defect was invisible to the suite (#752, #753, #755)

The regenerate-and-diff gate cannot see an unpaired predicate: a proposal
omitting the shared synonym is self-consistent, so the artifact matches and the
gate stays green. Added five assertions and mutation-verified each.

The second commit is a review finding **against the first commit's own test**:
its stem heuristic (`removesuffix("s")`) missed 3 of 5 real predicate shapes
*silently* — including `grows in`/`does not grow in`, i.e. METPO:2000517/2000518,
the exact pair this whole change exists to prevent. A test written to catch
that defect could not catch it. Now inflects the first word into a candidate
set, with the shape table asserted so it cannot regress.

## Verification

- `robot template` (both) + `merge` + `reason --reasoner ELK`: exit zero,
  **0 UNSAT**, delta 18 lines. `IAO:0000119` axiom confirmed on 2000802.
- Loader pairing simulated against the compiled OWL:
  `tolerance -> {positive: 2000064, negative: 2000065}` — previously **empty**.
- Mutations, each with the artifact restored and re-confirmed after: dropping
  the synonyms column fails 3 of 4; blanking its directive fails 1; removing the
  shared `tolerance` synonym fails 2; a same-polarity key clash fails 1;
  injecting an unpaired `2000517`/`2000518` fails the fixed heuristic and
  passed the old one.
- `poetry run tox` green.

## Not in scope

- **#750** — 7 of 9 property rows carry `TODO:add_citation`, which per the skill
  blocks upstream submission. Each needs a real source looked up; nine
  literature lookups would bury this change.
- **#744** stays open until METPO mints the term; only then does microbedecoder
  swap `subclass_of` for it.


realmarcin:
Closed as collateral: GitHub closes open PRs whose **base** branch is deleted, and it does not auto-retarget them (that only happens for PRs based on a merged PR's *head*). Deleting `revert/751-merged-without-review` after merging #754 took this one with it, and a closed PR's base cannot be changed, so it cannot be reopened.

No work was lost — `feat/metpo-type-strain-predicate-v2` is untouched. Continues in the replacement PR against `master`.

Issue #770, opened 2026-08-13
BacDive transform drops @ref, so per-statement paper attribution is impossible downstream


realmarcin:
The BacDive transform extracts phenotypes comprehensively — **4,243,914 edges** with METPO predicates. But it discards the one thing needed to attribute a statement to the paper it came from.

## What the transform records

`primary_knowledge_source` holds the **strain record**, not the reference:

```
kgmicrobe.strain:bacdive_1  biolink:subclass_of  NCBITaxon:435  ...  ['infores:bacdive', 'bacdive:1']
kgmicrobe.strain:bacdive_1  METPO:2000517  mediadive.medium:105  ...  ['infores:bacdive', 'bacdive:1']
```

Verified on 50,000 edges: the id in `primary_knowledge_source` equals the subject's strain id in **50,000 of 50,000** cases. It identifies which BacDive record an edge came from, never which citation.

## What BacDive actually provides

BacDive attaches an `@ref` to **every individual data item**, and each strain record's `Reference` section maps those ids to citations via a `doi/url` field. So the raw data supports statement-level paper attribution.

`kg_microbe/transform_utils/bacdive/bacdive.py` (3,419 lines) contains **no handling of the `Reference` section** — the only two `@ref` occurrences are in comments.

## Why this matters

The `@ref` linkage is lost at transform time and cannot be reconstructed from the output, because the edge no longer records which data item it came from. Anyone needing "which paper says this organism grows on this medium" has to **re-parse the raw JSON in parallel**, duplicating extraction logic that already exists here.

That is what we ended up doing in `kg-microbe-projects` (`bacdive_doi_triples/`), and the cost of duplicating is exactly what you'd expect: it covers **9 sections** and produces **313,778** DOI-attributed triples, against the transform's 4.24M edges. Roughly **50% of referenced BacDive statements** are missed (excluding phylogeny and provenance sections) — morphology, halophily, culture pH, spore formation and the API panels are absent entirely.

## Suggested change

Emit the `@ref` id per edge — an extra column, or an additional entry in `primary_knowledge_source` alongside the existing strain id. Optionally resolve it to a DOI from the `Reference` section during the transform.

Either way the downstream join becomes trivial and the parallel extraction can be deleted, taking coverage from ~313k attributed statements to potentially all 4.24M.

## Checked against

- `data/transformed/bacdive/edges.tsv` (4,243,914 edges; 3,859,944 carry `bacdive:<id>`)
[...]

Issue #737, opened 2026-08-11
BacDive per-reference observations disagree: 4,278 strains get contradictory phenotype edges


realmarcin:
Surfaced while fixing #474.

BacDive stores some phenotype blocks as an **array of per-reference observations**. Once #474 is fixed and those arrays are traversed, references that disagree produce contradictory edges for the same strain.

Measured across 99,392 strains, on the eight routed phenotype paths:

| key | multi-valued | disagreeing | example |
|---|---:|---:|---|
| oxygen tolerance | 4,599 | **2,643** | `anaerobe` vs `obligate anaerobe` |
| biosafety level | 4,963 | 825 | `1` vs `2` |
| cell shape | 2,033 | 405 | `coccus-shaped` vs `oval-shaped` |
| motility | 1,751 | 249 | `no` vs `yes` |
| gram stain | 2,361 | 104 | `negative` vs `variable` |
| spore formation | 393 | 20 | `no` vs `yes` |
| nutrition type | 30 | 29 | `chemoorganotroph` vs `photoheterotroph` |
| halophily level | 14 | 3 | `halotolerant` vs `slightly halophilic` |
| **total** | **16,144** | **4,278 (26.5%)** | |

**This is not a regression from #474.** The previous behaviour was worse and less honest: it dropped array-shaped blocks entirely, so whichever strains happened to store the block as a dict got a single value and the rest got nothing. Nothing chose *between* conflicting references; the shape of the JSON decided. Making the disagreement visible is an improvement — but it is now visible, and something should decide what the graph asserts.

Options, roughly in increasing effort:

1. **Emit all** (current behaviour after #474). Honest, but a consumer asking "is this strain motile?" gets `yes` and `no` with nothing to choose between them.
2. **Emit all, carry the reference.** BacDive gives each observation an `@ref`. Attaching it as edge provenance would let a consumer resolve conflicts themselves. Closest to what the source actually says.
3. **Majority / most-specific wins.** Cheap, but silently discards evidence, and "anaerobe vs obligate anaerobe" is a specificity relation rather than a true conflict — the two need different handling.
4. **Flag the conflict** so downstream can filter.

Note the largest bucket is arguably not a conflict at all: `anaerobe` vs `obligate anaerobe` (2,643 of 4,278) is one reference being more specific than another, and METPO may already relate the two. Splitting genuine contradictions (`yes`/`no` motility) from specificity differences would shrink the real problem considerably.

[...]

PR #776, opened 2026-08-13
Emit the reference DOI on bacdive metabolite-utilization edges

realmarcin:
Addresses #770.

BacDive attaches an `@ref` to every data item and lists the matching citations in each record's `Reference` section. The transform doesn't read it, so an edge can name the **strain record** it came from but never the **paper**. That linkage can't be recovered downstream, because the edge no longer records which data item produced it.

## The change

`reference_dois()` maps a record's `@ref` ids to publication DOIs; the metabolite-utilization path threads it through. Where a DOI is known the edge carries it alongside the existing provenance:

```
['infores:bacdive', 'bacdive:1', 'doi:10.1099/ijs.0.02862-0']
```

Same list-literal convention already used for the strain id, so nothing changes shape.

## Two `doi/url` kinds are skipped deliberately

Catalogue URLs (DSMZ, StrainInfo) aren't publications, and BacDive's own record DOI (`10.13145/bacdive…`) points back at the record `primary_knowledge_source` already identifies. Either would attach confident-looking provenance that leads nowhere.

## Backward compatible by construction

`_StrainProvenanceWriter.writerow` takes `publication` as an **optional keyword**, so every call site that doesn't pass one emits exactly what it emitted before. That's what makes this landable one section at a time, and there's a test asserting it rather than a claim.

## Scope

**Metabolite utilization only** — the carbon-source relations. The pattern for the rest is identical: capture the item's `@ref` where the item is read, pass the resolved DOI at the write site. I stopped at one section so the change is reviewable and the convention can be agreed before it's repeated ~8 times.

## Testing

- **51 bacdive tests pass**, including 7 new ones covering the mapping, the single-dict form, the skip rules, malformed input, the appended DOI, and unchanged output without a publication.
- Verified `reference_dois()` against the real dump: **300/300** sampled records resolve at least one paper DOI.
- The 3 collection errors elsewhere in `tests/` are a missing `parameterized` dependency and are **identical on a clean checkout** — not from this change.

## Why it matters downstream

In `kg-microbe-projects` we needed statement-level citations and had to re-parse the raw JSON in parallel, which covers **9 of ~50** referenced sections. In a BacDive-vs-assay comparison there, only **81 of 996** rows could be attributed to a paper — and the one row where two papers disagreed with each other (*Alteromonas stellipolaris* on NAG: `10.1099/ijs.0.02862-0` says "does not assimilate", `10.1099/ijs.0.63521-0` says "uses as carbon source") turned out to be the most informative row in the analysis. Without DOIs that's indistinguishable from a curation error.

Threading `@ref` through the remaining sections would make every edge attributable and let that parallel extraction be deleted.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -113,6 +113,7 @@ from kg_microbe.transform_utils.constants import (
     MEDIUM_URL_COLUMN,
     METABOLITE_CATEGORY,
     METABOLITE_CHEBI_KEY,
+    REFERENCE_KEY,
     METABOLITE_KEY,
     METABOLITE_MAPPING_FILE,
     METABOLITE_PRODUCTION,
@@ -1936,6 +1937,9 @@ class BacDiveTransform(Transform):
                     isolation_dict = value.get(ISOLATION_SAMPLING_ENV_INFO, {})
                     morphology_dict = value.get(MORPHOLOGY, {})
                     physiology_dict = value.get(PHYSIOLOGY_AND_METABOLISM, {})
+                    # @ref -> DOI for this record, so a statement can name the
+                    # paper behind it rather than only the strain record.
+                    record_reference_dois = reference_dois(value)
                     safety_dict = value.get(SAFETY_INFO, {})
                     name_tax_classification = value.get(NAME_TAX_CLASSIFICATION, {})
 
@@ -2809,6 +2813,10 @@ class BacDiveTransform(Transform):
                                                 "utilization_type": utilization_type,
                                                 "metpo_predicate": metpo_predicate,
                                                 "metpo_label": metpo_label,
+                                                # The @ref of this specific item, not of the
+                                                # record: one strain record cites several
+                                                # papers and they can disagree.
+                                                "ref": metabolite.get(REFERENCE_KEY),
                                             }
                                         )
 
@@ -2869,6 +2877,7 @@ class BacDiveTransform(Transform):
                                 knowledge_level, agent_type = self._add_edge_metadata(
                                     item["metpo_predicate"], HAS_PARTICIPANT, item["chebi_key"]
                                 )
+                                doi = record_reference_dois.get(item.get("ref"))
                                 edge_writer.writerow(
                                     [
                                         organism_id,
@@ -2878,7 +2887,8 @@ class BacDiveTransform(Transform):
                                         self.knowledge_source,
                                         knowledge_level,
                                         agent_type,
-                                    ]
+                                    ],
+                                    publication=f"doi:{doi}" if doi else None,
                                 )
 
                     if phys_and_metabolism_metabolite_production:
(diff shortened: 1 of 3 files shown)


realmarcin:
## Adversarial review

Reviewed on request; not my PR, so nothing pushed. The core idea is right and `reference_dois()` is careful about what it refuses to call a citation (catalogue URLs, malformed entries). Three blockers, filed separately, and two design questions worth settling before the pattern is repeated eight times for #737.

Quantitative claims were measured by replaying the PR's own `reference_dois` logic over the real `data/raw/bacdive_strains.json` (99,392 records, 862,829 metabolite-utilization items with a `Chebi-ID`). Those are **derived** counts, not counts of emitted rows — the transform was not run.

### Blockers

* **#792 — `tox -e lint` fails.** 7 ruff errors, all in new code; master clean. Independently reproduced. Now a hard merge gate: `master` began requiring `build (3.10/3.11/3.12)` on 2026-08-15.
* **#793 — the single-dict metabolite-utilization branch drops `@ref`.** `:2819` (list branch) captures it; `:2843-2851` (dict branch) does not, so `:2880` reads `None`. **292 records take that shape; 157 have a `@ref` that resolves to a real DOI** — all silently unattributed. `test_reference_dois_accepts_a_single_dict` covers the single-dict shape of the *`Reference` section*, not of *metabolite utilization*, so the suite reads as though this is covered.
* **#794 — the scalar DOI un-dedupes 4,052 edge groups.** `drop_duplicates` is a full-row pandas dedupe; rows that used to collapse now differ only in the pks cell. 3,802 groups gain an unattributed twin alongside the attributed row.

### Design

**The coverage number in the PR body is measured on the wrong denominator.** "300/300 sampled records resolve at least one paper DOI" is true (98,448/99,392 records, 99.0%), but per *edge* — which is what is emitted:

| | items | share |
|---|---:|---:|
| resolve to a DOI | 201,976 | **23.4%** |
| ref has no `doi`/`url` | 608,545 | 70.5% |
| ref is a catalogue URL | 52,289 | 6.1% |

The 608,545 are not noise — by title they are `Automatically annotated from API 50CH acid` (303,167), `API rID32STR` (72,691), `API 20NE` (62,881) and so on: machine-read test panels. Worth carrying 23.4% in the PR body.

**Emitting only the DOI, and not the `@ref` id, discards more than it keeps.** #770 asked for the `@ref` id per edge and called DOI resolution *optional*; this PR does only the optional half. Carrying `bacdive.ref:N` as well costs nothing, raises the attribution ceiling from 23.4% to ~100%, and preserves a signal the transform currently gets wrong — those ~608k API-panel items are stamped `observation / manual_agent`, and the `@ref` this PR now reads is exactly what would let that be corrected later.

**`publications` looks like the right slot.** The DOI currently lands in `primary_knowledge_source`, whose Biolink range is a single infores CURIE; `fix_list_representations.py:22` then rewrites the cell to `infores:bacdive|bacdive:145|doi:10.1099/...`. That same module already lists `publications` at line 34 as a known multivalued edge field. Putting it in `Transform.edge_header` would make the DOI visible to Biolink-aware consumers, **dissolve #794 for free**, and give #737 somewhere to put per-reference observations. It does mean a real header change across all 36 write sites — as the PR stands there is no column added and no ragged-row risk.

### Tests

The production wiring is **entirely untested** — verified by mutation. Deleting both hooks (the `ref` capture at `:2819` and the `publication=` argument at `:2891`) in a throwaway worktree still gives `49 passed, 2 skipped`. All 7 new tests exercise `reference_dois` and `_StrainProvenanceWriter` in isolation; nothing touches the metabolite-utilization path, which is why #793 slipped through. `tests/resources/bacdive/yaml/{145,2567,10174}.yaml` already carry `metabolite utilization`, so an end-to-end assertion is cheap.

Also: `test_non_strain_rows_are_untouched_even_with_a_publication` asserts that a supplied DOI is **silently discarded** when the row has no strain endpoint — a hazard encoded as a requirement. The isolation-source path uses a bare `bacdive` pks, so threading a DOI there under #737 would be a no-op with no error and no warning.

### Minor

* `:3472` hand-builds a list literal with no escaping. Safe today (no DOI in the dump contains a quote or backslash), but one contains spaces — `10.1094/ MPMI -19-0181` — which becomes a CURIE with spaces once piped. `str(parts)` is safer.
* `:3446` `publication: str = None` — implicit Optional; want `str | None`.
* `:3476` `writerows` takes no `publication`; 5 of 36 write sites use it, which matters for #737.

### Verified non-issues

`@ref`/`@id` types match (all int, both sides, 383,408 / 862,829); `@ref` is never a list here; no record has duplicate `@id` with conflicting `doi`/`url`; 0 metabolite-utilization items point at a PMID-only reference (so skipping `pubmed` loses nothing *here* — it will matter for #737's other sections); lowercase `doi:` matches `microbedecoder/utils.py:250` and `lpsn_api/lpsn_api.py:93`; scoping of `record_reference_dois` is correct; `-k bacdive` gives 49 passed / 2 skipped on the branch.

### Not concluded

Post-merge behaviour of the duplicated rows in #794 (running the merge rewrites tracked artifacts); whether any DOI resolves to a real, correct paper (syntactic shape only).

Issue #757, opened 2026-08-12
Branch protection permits admin bypass of required review — the failure mode that caused #754


realmarcin:
Found reviewing #754 (the revert of #751). Root cause of that incident.

`master` requires an approving review, but the rule does not apply to
administrators, so `gh pr merge --admin` merges straight through it. That is
exactly what happened with #751: a plain merge failed on `REVIEW_REQUIRED`,
`--admin` succeeded, and work reached `master` reviewed by nobody but its own
author.

The revert fixes that instance. It does nothing about the mechanism, which
remains available for the next one — and the next one may not be caught.

### Ask

Enable **"Do not allow bypassing the above settings"** on the `master`
protection rule (classic protection: tick *Include administrators*).

```bash
gh api -X PATCH repos/Knowledge-Graph-Hub/kg-microbe/branches/master/protection \
  -f enforce_admins=true ...
```

### The tradeoff is real and it is yours to weigh

This also blocks *you*. Every change to `master`, including one-line docs fixes
and emergency reverts, would then need a second account's approval or a
temporary rule toggle. On a repo where you are frequently the only human
reviewer, that may be worse than the problem.

Middle options if full enforcement is too rigid:

- Leave admin bypass enabled but treat it as break-glass, and have the agent
  never invoke `--admin` without a go-ahead that names the bypass specifically
  (not just "merge"). This is a convention, not a control — it failed once.
- Require the branch to be up to date and CI green, so a bypass at least cannot
  land broken code, while review stays advisory.

Deliberately **not** applying this myself: it changes who can write to your
default branch, which is your call, not a review finding I should act on.

PR #778, opened 2026-08-14
fix(mappings): correct 14 wrong isolation-source ids, and check ids against the ontology

### LINKED ISSUE #777: isolation_source_to_ontology.tsv: 14 rows whose object_id names a different concept than object_label — and the validator trusts exactly those rows
14 rows in `mappings/isolation_source_to_ontology.tsv` have an `object_id` that does not denote what their `object_label` says. Six are semantically unrelated concepts. Checked against kg-microbe's own extracts in `data/transformed/ontologies/*_nodes.tsv`, so this is self-inconsistent within the repo — no external ontology version is involved.

## The rows

| subject_label | object_id | object_label claims | the ontology says |
|---|---|---|---|
| Nectar | `CHEBI:50292` | nectar | **cadmium sulfate** |
| Mushroom | `FOODON:03411260` | mushroom | **ginkgo tree** |
| Meat | `FOODON:00001027` | meat food product | **cheddar cheese food product** |
| Indoor-Air | `ENVO:01000855` | indoor air | **area of mixed forest** |
| Indoor | `ENVO:01000856` | indoor environment | **temperate marginal sea biome** |
| Heavy-metal | `CHEBI:25555` | monoatomic ion | **nitrogen atom** |
| Iron-mat | `ENVO:01000110` | microbial mat | hydrogenous sediment |
| Tick | `NCBITaxon:6939` | Ixodida | Ixodidae |
| Gastrointestinal-tract | `UBERON:0005409` | digestive tract | alimentary part of gastrointestinal system |
| Tundra | `ENVO:00000112` | tundra | area of tundra |
| Domestic-waste, Solid-animal-waste, Solid-plant-waste | `ENVO:00002264` | Waste | waste material |

The last four are label-wording differences and harmless. The first six are wrong mappings: a plant nectar isolation source resolves to a toxic metal salt, and mushroom to a tree.

There is no id offset — the intended concepts live at scattered ids (`microbial mat` is `ENVO:01000008`, `digestive tract` is `UBERON:0001555`, `monoatomic ion` is `CHEBI:24867`), and several intended labels are absent from kg-microbe's extracts entirely. So this looks like individually wrong ids rather than a misaligned column or a stale ontology build. **The `object_label` column appears to record what the curator meant, and is the recoverable signal.**

## Root cause: the validator skips precisely the rows that are wrong

`mappings/validate_isolation_source_mappings.py` only checks rows that pass `_row_is_trusted`, which is inverted relative to risk — it *skips* the checks for rows that assert they don't need them:

```python
return (confidence == "high") or (justification == "semapv:ManualMappingCuration")
```

10 of the 14 wrong rows are `('semapv:ManualMappingCuration', 'high')` — both trust conditions at once.

Error rate by the `curator` column, over rows whose target is resolvable in kg-microbe's own extracts:

```
ols4_auto              0 wrong / 150 checkable     <- the automated lookup is clean
ols4_search_synonym    3 wrong /  34
kg_review_promote      5 wrong /  30
family_mismatch_fix    2 wrong /  17
codex_review_fix       3 wrong /   5
codex_review_fix_v2    1 wrong /   1
```

The automated OLS4 pass is correct on all 150 checkable rows. Every error came from a later review/fix pass, and those passes marked their output high-confidence manual curation, which then exempted it from validation.

## Scope

Other mapping files were checked the same way. `special_chemical_mappings.tsv` (8/83), `enzyme_mappings.tsv` (4/13) and `chemical_mappings.tsv` (1/9) disagree only on label wording (`indole` vs `1H-indole`, a FoodEx2 code prefix, a GO label change) — worth tidying but not wrong. The semantically wrong mappings are concentrated in `isolation_source_to_ontology.tsv`.

I could not check `kgmicrobe_unified_entity_mappings.sssom.tsv.gz`, which has no `object_label` column in the copy I have, so its ids are unverifiable by this method.

## Suggested fix

1. Add a label check to the mapping validators: for every row with an `object_id` resolvable in `data/transformed/ontologies/`, fail when `object_label` disagrees. It is cheap and would have caught all 14.
2. Invert the trust policy. High confidence and manual curation are claims about a row, not evidence for it; they are a reason to check, not to skip.
3. Re-derive the six wrong ids from their `object_label` — the labels look correct throughout.

## Downstream impact

Found from [HabitatMech](https://github.com/CultureBotAI/HabitatMech), which consumes this table to ground habitat records. It shipped a record labelled "area of mixed forest" for GOLD's `Environmental > Air > Indoor Air`, and five further records inherited it as their parent, so one wrong id contaminated a whole branch. HabitatMech now verifies this table's labels before using it and drops the mapping when they disagree, which is a workaround rather than a fix — the ids are still wrong here.

## Open question

These may have entered via an SSSOM mapping shared with a sibling repo (MediaIngredientMech was suggested as a possible origin) rather than being authored in this file. I could not confirm that: the specific ids do not appear in `ingredient_mappings.sssom.tsv`, and I do not have that repo to check. Worth tracing whether the `codex_review_fix*` and `kg_review_promote` passes drew on a shared mapping source, since the same ids would then be wrong wherever else it is consumed.

realmarcin:
Issue #777 found rows in `isolation_source_to_ontology.tsv` whose `object_id`
denotes something other than what `object_label` claims. Some were badly wrong:
`Nectar` pointed at CHEBI:50292 (cadmium sulfate), `Meat` at FOODON:00001027
(cheddar cheese food product), `Mushroom` at FOODON:03411260 (ginkgo tree),
`Indoor-Air` at ENVO:01000855 (area of mixed forest).

Corrections, verified against OLS4 and against our own ontology extracts:

  id changed (6)
    Heavy-metal            CHEBI:24867      -> CHEBI:5631        heavy metal
    Iron-mat               ENVO:01000110    -> ENVO:01000008     microbial mat
    Tick                   NCBITaxon:6939   -> NCBITaxon:6935    Ixodida
    Nectar                 CHEBI:50292      -> BTO:0000537       nectar
    Mushroom               FOODON:03411260  -> FOODON:00002050   mushroom (whole or pieces)
    Meat                   FOODON:00001027  -> FOODON:03317170   meat (whole or pieces)

  label changed, id was already right (5)
    Gastrointestinal-tract UBERON:0005409   alimentary part of gastrointestinal system
    Tundra                 ENVO:00000132    area of tundra
    Domestic-waste / Solid-animal-waste / Solid-plant-waste        waste material

  unmapped (3)
    Indoor, Indoor-Air, Outdoor-Air — ENVO has no term for these, and the ids
    previously carried were invented. Object columns blanked.

`Gastrointestinal-tract` is the instructive one: UBERON:0005409 has
"gastrointestinal tract" and "GI tract" as exact synonyms, so the id was right
all along and only the label was wrong. That is why a first pass that trusted
the label and swapped the id would have made this row worse.

Prevention: the validator now resolves every `object_id` against
`data/transformed/ontologies/*_nodes.tsv` and fails when the resolved label
disagrees with `object_label`. Two deliberate choices:

  * The check is NOT behind `_row_is_trusted`. 10 of the 14 wrong rows were
    trusted and passed validation, so the trust gate was never what hid them —
    nothing resolved an id at all. It also matters because other projects read
    this TSV directly (HabitatMech grounds habitat records from it), where our
    loader's trust policy does not apply.
  * It skips silently when the extracts are absent, so the stdlib-only
    validator still runs in a lightweight CI container.

The check is retroactively effective: run against the pre-fix file it reports
all 14 rows with the true label for each.

Closes #777


realmarcin:

PR #786, opened 2026-08-15
fix(mappings): un-invert the cobalamin retraction, and retract at name level

### LINKED ISSUE #599: Harden apply_retractions: name-level granularity, index cleanup, orphan assertion
Follow-up from PR #597 (chemical-mapping retraction pass). The pass is safe today, but two design limitations are worth hardening before the retraction list grows.

### 1. Object-granular retraction depends on a frozen external analysis
`apply_retractions` drops at the **object** level (`del self.chemicals[oid]`), but the underlying data is *name→object* groundings. If a listed stale object is *also* the correct grounding for some name that is **not** in the retraction list — now or as the synced upstream (`UNIFIED_INGREDIENT_MAPPING.tsv`, the MIM SSSOM) evolves — that name is silently orphaned. The sole-source guard can't prevent this (the legitimate name would also be sole-sourced from `culturebotai_reviewed`).

Mitigations already in place (PR #597 review round): a **non-fatal orphan guard** warns if any `now_asserted` target is absent after the pass. Verified 0 orphans on the current 161-row list.

Possible further hardening:
- Move to **name-level** retraction (drop the specific `(name, object)` grounding, keep the object if other names still point to it), or
- Make the orphan guard **fatal** in CI once we trust it, or
- Regenerate/re-verify the retraction list from a diff of consecutive `culturebotai_reviewed` inputs rather than a frozen file.

### 2. `del self.chemicals[oid]` leaves dangling `name_index` / `formula_index` entries
The drop removes only the `self.chemicals` record; `name_index[norm_name]` / `formula_index[formula]` still point at the deleted CURIE. Harmless today (`export_unified_sssom` and `propagate_synonyms_via_xrefs` both iterate/snapshot `self.chemicals` and don't consult `name_index` afterward), but a latent trap for any future code that resolves a name via `name_index` after this pass. Cheap fix: prune those indices in the drop loop.

### 3. Test coverage
Added in review: idempotency, curator-qualified sole-source, empty-source-not-dropped. Still worth adding: an explicit orphan-detection test (a listed object that is also the sole-source correct target of a second unlisted name).

Context: root cause of 3 of the 16 stale objects is upstream and filed at CultureBotAI/MediaIngredientMech#138.

### LINKED ISSUE #784: kgm_retracted_groundings.tsv retracts the correct cobalamin term and keeps the wrong one
The cobalamin row in `mappings/kgm_retracted_groundings.tsv` is inverted:

    kgm.name:cobalamin  COBALAMIN  CHEBI:30411  Cobalamine  CHEBI:28911  culturebotai_reviewed

It drops `CHEBI:30411` as stale in favour of `CHEBI:28911`. Upstream has since concluded the opposite (CultureBotAI/MediaIngredientMech#362, merged 2026-08-14), and the argument checks out against our own ChEBI extract:

    CHEBI:30411  cobalamin        "...cobalamin with the oxidation state of the central cobalt atom unspecified."
    CHEBI:28911  cob(III)alamin   "A cobalamin in which the central cobalt atom has an oxidation state of +3."

and `data/transformed/ontologies/chebi_edges.tsv` has `CHEBI:28911 subclass_of CHEBI:30411`, i.e. 30411 is the parent and the generic term.

The +3 evidence was circular: `AUTO_BACKFILL_CHEBI_CHEMISTRY` copied `molecular_formula` / `inchi` / `smiles` out of `CHEBI:28911` on 2026-05-01 while the record was already mapped there, and five of the record eight synonyms are verbatim from `CHEBI:28911` own ChEBI synonym list. A formula cannot select an oxidation state in any case: ChEBI records identical `C62H88CoN13O14P` and identical average mass for the +1, +2 and +3 forms, which differ only in the separate charge field. The one independently-sourced field, `cas_rn 18534-66-2`, is cob(I)alamin, i.e. a third state, and is inconsistent with the record own formula.

Secondary defect in the same row, same class as #777: its `stale_object_label` column calls `CHEBI:28911` "Cobalamine". ChEBI label for it is "cob(III)alamin". "Cobalamine" is the bare vitamer name from the source media list, which is what makes `CHEBI:30411` the right target.

Fix: drop the cobalamin row from the retraction list so `CHEBI:30411` survives as the grounding. Blocked on nothing, but see the sibling issue about re-syncing the input snapshot — that sync may carry the correction in on its own.

### LINKED ISSUE #785: culturebotai_reviewed_ingredients.tsv is a 2026-07-20 snapshot; upstream rebuilt UNIFIED_INGREDIENT_MAPPING.tsv on 2026-08-14
Our copy of the CultureBotAI reviewed ingredients is frozen at 2026-07-20 (`mappings/culturebotai_reviewed_ingredients.tsv`, last written by `1145296`). Upstream rebuilt `UNIFIED_INGREDIENT_MAPPING.tsv` on 2026-08-14 in CultureBotAI/MediaIngredientMech#358, which had itself been stale since 2026-07-20, and then corrected the Cobalamine grounding in #362.

So our snapshot predates both. Concretely, our file still carries:

    Cobalamine            chebi_id=CHEBI:30411  kg_microbe_node_id=CHEBI:28911   (upstream now says 30411 is right — see #784)
    4-Aminobenzoic acid   chebi_id=CHEBI:194474 kg_microbe_node_id=CHEBI:30753   (lines 167, 1040, 3346)
    Infusion from Potatoes  culturemech_term_id=FOODON:03316428, mim_id=UNMAPPED_0251

Two things worth doing, in order:

1. **Re-sync the snapshot** and re-run `scripts/consolidate_chemical_mappings.py`. Check afterwards whether any of the 16 stale objects on `mappings/kgm_retracted_groundings.tsv` are now unnecessary — a shorter list is the point of the exercise, and #599 notes the list growing is the thing to avoid.
2. **Record how the sync happens.** There is no script or documented procedure in this repo for refreshing this file from upstream; it was hand-copied. That is why it silently aged a month. Even a make target plus a provenance header with the upstream commit would make staleness visible.

Note that a re-sync will *not* fix the other two rows. Per CultureBotAI/MediaIngredientMech#138, both stale values live in CultureMech `ingredients[].term.id`, which `best_primary` reads before `mim_id`, so no MIM-side change reaches them:

* `4-Aminobenzoic acid` — 251 CultureMech files carry `CHEBI:194474`, which is `4-ammoniobenzoate`, a zwitterion; every one of the 251 is a PABA label. Fix prepared and canaried upstream, not applied.
* `Infusion from Potatoes` — `FOODON:03316428` is `sweets` in FOODON (not "Peptone" as #138 says), while CultureMech stores `label: potato extract` against it. No FOODON term for potato extract exists, so unmapped is the correct state.

Related: #599, #723, #784.

realmarcin:
Two coupled fixes plus the machinery the first one needed.

**#784 — the cobalamin retraction row was backwards.** It read

    kgm.name:cobalamin  COBALAMIN  CHEBI:30411  Cobalamine  CHEBI:28911

i.e. it retracted `CHEBI:30411` in favour of `CHEBI:28911`. Checked against our
own extracts, that is the wrong way round: `CHEBI:30411` "cobalamin" is defined
as the cobalt oxidation state being *unspecified* and is the direct parent of
`CHEBI:28911` "cob(III)alamin" (`chebi_edges.tsv` carries the subclass_of), and
the source media list the bare vitamer name alongside Biotin and Thiamine.

The +3 evidence was circular: `AUTO_BACKFILL_CHEBI_CHEMISTRY` had copied the
formula, InChI and SMILES out of `CHEBI:28911` while the record was already
mapped there. A formula cannot select an oxidation state in any case — ChEBI
records the same `C62H88CoN13O14P` and the same average mass for the +1, +2 and
+3 forms, which differ only in the separate charge field. Upstream reached the
same conclusion independently in CultureBotAI/MediaIngredientMech#362.

The row never actually fired — the sole-source guard skipped it every run
because `CHEBI:30411` had gained other sources. So nothing shipped wrong from
it; it was a landmine that would have gone off the moment that source set
narrowed.

**#785 — re-synced from upstream.** `culturebotai_reviewed_ingredients.tsv` was
a 2026-07-20 snapshot; the sibling `../MediaIngredientMech` is at `5530527b`
with the Cobalamine correction. The consolidator syncs it automatically, so
this is just the result of running it. Contrary to what I first wrote on #785,
there was never a missing sync procedure — the file had simply aged because
nothing had run it. The other two rows in #138 are unchanged, as expected:
both stale values live in CultureMech's `ingredients[].term.id`, which
`best_primary` reads before `mim_id`, so no MIM-side fix reaches them.

**Name-level retraction (#599 item 1), which the above turned out to require.**
After the sync, `kgm.name:cobalamine` carried *two* exactMatch objects —
`CHEBI:30411` from the corrected input and `CHEBI:28911` kept alive by additive
re-seeding. Neither existing tool fixed that. An object-level retraction of
`CHEBI:28911` is refused by the sole-source guard, and should be: it is a real
ChEBI term whose KEGG, PDB-CCD and Gmelin xrefs would be collateral damage.

So `apply_retractions` now takes an optional `retract_names` column and strips
the enumerated names from the record instead of dropping it. Guard rails:

* A name is removed only once some `now_asserted` target actually carries it,
  so a retraction can never leave a name grounded nowhere. That check is the
  name-level equivalent of the sole-source guard, and is what makes these rows
  safe under mixed sources.
* `propagate_synonyms_via_xrefs` now refuses to re-add a retracted name. Without
  this the pass was useless: it copies names across the exactMatch xref, which
  points straight at the target that superseded them, so all four names
  reappeared as synonyms in the same run. Found by running it, not by reasoning.
* On retracting a canonical name, restore the term's *own* ontology label rather
  than promoting a surviving synonym. My first attempt promoted alphabetically
  and labelled cob(III)alamin "Coalpha-[alpha-(5,6-dimethylbenzimidazolyl)]-
  cobamide", which is worse than what it replaced.

Also #599 item 2: an object-level drop now prunes `name_index` / `formula_index`
instead of leaving them pointing at a deleted CURIE.

Result — `Cobalamine`, `COBALAMIN`, `Cbl` and `Cobalamin (B12)` ground only to
`CHEBI:30411`; `CHEBI:28911` keeps its xrefs, its own label, and its genuine
synonyms (`Cob(III)alamin`, `cobalamin(III)`, ...). Verified byte-identical
across two consecutive consolidator runs, which matters because the script seeds
from its own prior output.

`data/transformed/ontologies_stubs/` regenerated: the sync introduced 14 new
NCIT ids and 13 new mesh ids needing stub nodes.

Closes #784
Closes #785
Closes #599


realmarcin:

Issue #793, opened 2026-08-15
PR #776: the single-dict metabolite-utilization branch silently drops @ref — 157 records unattributed


realmarcin:
Blocker from the adversarial review of #776. Verified by reading the branch.

BacDive stores metabolite utilization as either a list or a single dict. The PR captures the per-item `@ref` in the **list** branch only:

* `kg_microbe/transform_utils/bacdive/bacdive.py:2819` — list branch, has `"ref": metabolite.get(REFERENCE_KEY)`
* `kg_microbe/transform_utils/bacdive/bacdive.py:2843-2851` — dict branch, builds its dict literal **without** it

So at `:2880`, `item.get("ref")` returns `None`, `record_reference_dois.get(None)` returns `None`, and no DOI is attached.

Measured against the real `data/raw/bacdive_strains.json`: **292 records take the dict shape here, and 157 of them have an `@ref` that resolves to a real DOI.** All 157 are silently unattributed — no error, no warning.

What hides it: `test_reference_dois_accepts_a_single_dict` covers the single-dict shape of the *`Reference` section*, not the single-dict shape of *metabolite utilization*. The suite reads as though the case is covered.

Deeper point for #737: capturing `@ref` per-section, per-shape is a bug class, not a bug. This PR got it wrong in one of two shapes within a single section; the eight phenotype paths in #737 are sixteen more chances at the same silent failure. Capturing `@ref` where the raw item is read — or a shared helper that builds the intermediate dict — removes the class instead of repeating it.

PR #795, opened 2026-08-15
feat(bacdive): report per-reference phenotype conflicts, split by kind (#737)

realmarcin:
#737 quoted "4,278 strains get contradictory phenotype edges" from an ad-hoc
measurement that was never persisted. Nothing on disk recorded which strains, or
which references disagreed: the bacdive transform writes only nodes.tsv,
edges.tsv and bacdive_media_links.txt, and the contradictory edges sit in
edges.tsv unmarked — same primary_knowledge_source on both rows, and no slot for
the `@ref` that would tell them apart.

This makes the number reproducible and inspectable, and splits it three ways.

Restricted to #737's eight fields the script reproduces **4,278 exactly**, which
is the check that it is measuring the same thing. Splitting that total:

  specificity     955  values on one METPO subsumption chain
  contradiction  3,342 strains  values that cannot both hold
  unresolved            a value with no METPO term at all

**The specificity share is far smaller than I assumed, and #737's note that
`anaerobe` vs `obligate anaerobe` is "arguably not a conflict at all (2,643 of
4,278)" reads as if subsumption explains most of the problem. It does not.**
All 955 specificity cases are oxygen tolerance, and there they are 36% of
disagreements, 22% of the 4,278 — 854 `aerobe | obligate aerobe` and 101
`anaerobe | obligate anaerobe`. The dominant oxygen-tolerance patterns are
genuine contradictions: `aerobe | facultative anaerobe` (440),
`anaerobe | microaerophile` (299), `facultative anaerobe | microaerophile` (266).
So ~3,342 strains really do need a resolution policy.

Getting that split required resolving through METPO **synonyms**, not labels.
BacDive writes `anaerobe`; METPO's label is `anaerobic` and carries `anaerobe`
as a synonym. A label-only match resolves nothing and reports every
oxygen-tolerance disagreement as a contradiction — my first cut did exactly
that and reported 4,686 strains, all "contradiction".

Third bucket is deliberate. `unresolved` means a value has no METPO term, so the
relation cannot be judged; calling it a contradiction would report a curation gap
as a data conflict. `colony color` is 1,196 rows, every one unresolved, because
colours have no METPO terms.

The script also scans four fields #737 did not — `colony color`, `colony shape`,
`type of spore`, `forms multicellular complex` — adding 1,793 more disagreeing
observations, of which `forms multicellular complex` (551) and `colony shape`
(23) are the ones worth looking at.

Deliberately not a gate: it exits 0 and reports. Wire it into CI once the counts
are expected to be stable.

One design note carried over from reviewing #776: `observations()` handles the
list and single-dict shapes in **one** code path. Handling them per-section is
how #793 happened — that PR captured `@ref` in the list branch and not the dict
branch of a single section, silently unattributing 157 records.

Refs #737.

### DIFF
--- a/scripts/bacdive_reference_conflicts.py
+++ b/scripts/bacdive_reference_conflicts.py
@@ -29,16 +29,16 @@ this reproduces its 4,278 exactly — and splits it:
   earlier guess that subsumption explained most of #737 was wrong; the
   dominant oxygen-tolerance patterns are genuine contradictions
   (``aerobe | facultative anaerobe`` 440, ``anaerobe | microaerophile`` 299).
-* **contradiction 3,342 strains** — the set that needs a policy.
+* **contradiction 2,828 strains** — the set that needs a policy.
 * **unresolved** — a value with no METPO term, so the relation cannot be
   judged. Evidence of a mapping gap, not of a conflict.
 
 The script also scans four fields #737 did not (``colony color``,
 ``colony shape``, ``type of spore``, ``forms multicellular complex``), which
-add 1,793 more disagreeing observations. ``colony color`` alone contributes
-1,196, every one ``unresolved`` — colours have no METPO terms, so that field is
-a curation gap rather than a contradiction, and should not be folded into a
-conflict headline.
+add 1,793 more disagreeing observations. Of those, 1,196 ``colony color`` and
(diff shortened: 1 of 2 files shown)


realmarcin:

Issue #797, opened 2026-08-16
kgm-freshness-check: git checkout bumps mtimes, so the merge verdict flips without anything changing


realmarcin:
The skill compares file **mtimes** against commit times. `git checkout` between branches whose content differs rewrites the working-tree file and sets its mtime to now — so simply visiting an old branch can change the verdict with no content change at all.

Observed twice on 2026-08-15 during a session that checked out two April-based branches (`add-security-md`, `add-dependabot-config`), both carrying a different `merge.yaml`:

    merge status : STALE_VS_INPUTS  (16 FRESH)   before
    merge status : STALE_VS_YAML    (15 FRESH)   after

`merge.yaml` at that point: `git status` clean, last commit `2026-08-12`, mtime `2026-08-15`. Nothing had changed; the file had merely been rewritten twice by checkout.

Two independent reviewers spent effort on the discrepancy before it was traced, which is the real cost — the skill's output looked like a substantive disagreement about repo state.

The docs already note the `cp -p` caveat ("Copying an output from another checkout with `cp -p` preserves mtime"), but not this one, which is far more common: anyone reviewing several branches in one sitting will hit it.

Options, roughly increasing in effort:

1. **Document it** in the Limitations section, and in the human output add a line when a compared file's mtime is newer than its last commit — that is the signature of a checkout-touched file, and it is cheap to detect (`git log -1 --format=%at -- <path>` vs `stat`).
2. **Prefer git times over mtimes for tracked inputs.** `merge.yaml`, `merge_utils/` and the transform code dirs are all tracked, so their real "when did this change" is the commit time, not the mtime. Only the *outputs* under `data/` are untracked and genuinely need mtimes.
3. **Hash tracked inputs** instead of timestamping them.

Option 2 looks right: it removes the whole class for every tracked input while leaving the output side, where mtime is the only signal available, unchanged.

Note the substantive verdict was correct throughout — the merge really is stale against `data/transformed/ontologies_stubs/`. The defect is that the *reason* given was wrong and unstable, which is enough to make the tool untrustworthy for the release pre-flight it is meant to gate.

PR #791, opened 2026-08-15
docs: reconcile NEXT_TASKS.md against the live backlog (2026-08-15)

realmarcin:
`NEXT_TASKS.md` tracked only the 2026-07-02 untracked-file triage, so answering "what's next?" from it missed every active thread.

Adds a live-backlog section ahead of the original: the three mergeable PRs and the re-merge they gate, the unblocked correctness work, the upstream-blocked items *with the reason*, and the governance debt.

Re-verified the original triage — all 8 remaining paths are still on disk, so that section is unchanged apart from the reconcile date.

Note for whoever reads it next: the merge is genuinely `STALE_VS_INPUTS`, not a timestamp artifact. #778 and #786 introduce `BTO:0000537`, `mesh:D010514` and 14 NCIT ids that the shipped graph has no nodes for.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #788, opened 2026-08-15
The upstream re-sync imports four grounding regressions and net-increases name ambiguity 572 -> 634


realmarcin:
Found by adversarial review of #786 and verified independently.

The re-sync in #786 pulls `mappings/culturebotai_reviewed_ingredients.tsv` and `mappings/ingredient_mappings.sssom.tsv` from `../MediaIngredientMech` at `5530527b`. It fixes Cobalamine, but brings in regressions.

**Live in the shipped SSSOM** (name -> CURIE resolution changes):

| name | master | after sync |
|---|---|---|
| `alpha-Lactose` | CHEBI:36219 alpha-lactose | **CHEBI:189432 Alpha-lactose monohydrate** |
| `Carboxymethyl cellulose` | CHEBI:85146 carboxymethylcellulose | **CHEBI:234035 carboxymethylcellulose sodium salt** |

Both arrive as `skos:exactMatch` on the `MIM:` subject — an anhydrous/hydrate and a free-acid/sodium-salt false identity, exactly what the `hydrate_equivalences` closeMatch machinery exists to prevent. The other new MIM rows use closeMatch correctly, so this looks like a slip rather than a policy change. Also wrong in the file though not yet moving resolution: `MIM:Berberine` -> `CHEBI:31271 "Berberine chloride (TN)"` exactMatch, and `MIM:36-Dihydroxyflavone` -> `CHEBI:24698 hydroxyflavone` (a class) with its CAS dropped.

**Latent in the vendored TSV** (held back only because the additive baseline still holds the correct record — verified in `mappings/culturebotai_reviewed_ingredients.tsv`):

* `D-Glucose`, 589 occurrences: `CHEBI:17634 D-glucose` -> **`CHEBI:42758 aldehydo-D-glucose`** (the open-chain tautomer), and `mapping_status` flipped `MAPPED` -> `REJECTED`.
* `Cysteine-HCl`, 103 occurrences: `CHEBI:91247 L-cysteine hydrochloride` -> **`CHEBI:52891 QSY9 succinimidyl ester(1+)`**, a fluorescence quencher. 52891 was already on master (pre-existing), but the sync now ratifies it and discards the correct `mim_id=CHEBI:91247`, which `best_primary` demotes to an xref.
[...]

PR #803, opened 2026-08-16
docs: reconcile NEXT_TASKS.md against the live backlog

realmarcin:
Replaces #791, which was accidentally branched from `fix/cobalamin-retraction-and-resync` and conflicted once that PR squash-merged.

`NEXT_TASKS.md` tracked only the 2026-07-02 untracked-file triage, so answering "what's next?" from it missed every active thread. Adds a live-backlog section ahead of it, reflecting state after today's batch merge.

Includes the corrections external review found: #550/#551 are closed not open, the 33 Dependabot alerts were stale rather than outstanding, and the #788 regressions are miscounted no longer. Also deliberately drops `kgm-freshness-check`'s status code, which is volatile — see #797.

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -1,11 +1,104 @@
-# NEXT_TASKS.md — untracked entries that need a per-file decision
+# NEXT_TASKS.md
+
+## Live backlog — reconciled 2026-08-16
+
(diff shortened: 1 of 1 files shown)


realmarcin:

PR #814, opened 2026-08-17
fix: report data staleness, and refuse an unknown transform source

### LINKED ISSUE #812: Transforms report FRESH when their mapping-file inputs have changed — #778/#786 fixes never reached the merged KG
A merged KG was rebuilt on 2026-08-16 believing it carried the corrections from #778 and #786. It did not, and nothing in the tooling said so.

Verified in the shipped `data/merged/merged-kg.tar.gz`:

| defect #778 fixed | still in the graph | corrected id |
|---|---|---|
| `Blood-culture` -> `NCIT:C16403` "Cell Line" | **75 edges** | 0 edges -> `UBERON:0000178` |
| `Periodontal-pocket` -> `mesh:D010520` "Aggressive Periodontitis" | **59 edges** | 0 -> `mesh:D010514` |
| `Tick` -> `NCBITaxon:6939` Ixodidae | **28 edges** | 0 -> `NCBITaxon:6935` Ixodida |

## Root cause

`kgm-freshness-check` compares each transform's **code directory** mtime against its output mtime. It has no model of the **data** a transform reads. So:

    data/transformed/bacdive/edges.tsv   Aug 12   <- mappings/isolation_source_to_ontology.tsv changed Aug 15
    data/transformed/mediadive/edges.tsv Aug  5   <- mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz changed Aug 15
    data/transformed/madin_etal/...      Aug  1
    data/transformed/metatraits/...      Aug  1

and every one of them reports **FRESH**, with `0 STALE_VS_CODE` in the summary. The merge then faithfully combined stale inputs and produced a graph that validates perfectly — 0 errors, 0 dangling edges, all archetypes clean — while asserting things a merged PR had already corrected.

That is the dangerous shape: not a loud failure, but a clean bill of health on the wrong inputs.

## Consumers, measured

`mappings/isolation_source_to_ontology.tsv`: `bacdive`, `ontologies_stubs`.

`mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz` (via `chemical_mapping_utils`): `bacdive`, `ctd`, `madin_etal`, `mediadive`, `metatraits`, `metatraits_gtdb`, `microbedecoder` — six of them in `merge.yaml`.

## Why this matters right now

An updated SSSOM is expected from MediaIngredientMech. When it lands it will make all six consumers stale again, and the freshness check will keep reporting FRESH. Whoever re-merges next will hit exactly this.

## Suggested fix

Give each transform a declared set of data inputs — a `DATA_INPUTS: tuple[Path, ...]` class attribute on `Transform`, defaulted empty — and have `kgm_freshness_check` treat an input newer than the output as `STALE_VS_DATA`, listing which input. The mapping files are tracked, so their git commit time is the right signal (see #797: mtimes flip on `git checkout`).

A cheaper interim: hard-code the two mapping paths and their known consumers in the skill, and report `STALE_VS_DATA` for those. That covers the actual failure without a `Transform` API change.

## Related

#797 (mtime instability in the same skill) and #810 (sampled review reported 0 warnings at 20% edge coverage). All three are the same underlying shape — a gate that reads as passing when it has not really checked.

### LINKED ISSUE #813: kg transform -s <unknown> exits 0 silently; a no-op run is indistinguishable from a successful one
`kg_microbe/transform.py:114`:

    for source in sources:
        if source in DATA_SOURCES:
            logging.info(f"Parsing {source}")
            t = DATA_SOURCES[source](input_dir, output_dir)
            ...

There is no `else`. An unrecognised `-s` value is skipped in silence: no warning, no non-zero exit, no output. `poetry run kg transform -s bacdiv` (typo) and `poetry run kg transform -s bacdive` on a machine where the transform dies early are indistinguishable from the outside — both print nothing and exit 0.

This is not hypothetical. On 2026-08-16 a backgrounded `kg transform -s bacdive` exited 0 in seconds with a two-line log and left `data/transformed/bacdive/edges.tsv` untouched at its previous date. `bacdive` **is** registered, so the silent-skip path was not the cause — but the absence of any "starting / finished / wrote N rows" signal is precisely why the run could not be distinguished from a successful no-op, and the diagnosis is still open.

Note `logging.info` does not print under the CLI's default configuration, so even the registered path emits nothing at INFO.

Suggested:

1. **Fail loudly on an unknown source.** Raise (or exit non-zero) listing the valid keys. A typo should not look like success.
2. **Print a per-source start/finish line at a level that actually shows**, with the output paths and row counts written. `ontologies_stubs` already does this well (`[NCIT] wrote 231 nodes / 238 edges to ...`); most transforms print nothing comparable.
3. **Exit non-zero when a requested source produced no output file**, so a batch script can tell.

Related: #685 (one transform's hard failure aborts all remaining sources) is the opposite failure mode on the same loop — together they mean the dispatch is both too fragile and too quiet.

realmarcin:
Two failures with one shape: tooling that reads as passing when it has not
actually checked. Both were found by the merge on 2026-08-16, which validated
perfectly — 0 errors, 0 dangling edges, every path archetype clean — while
still asserting groundings two merged PRs had corrected.

**#812 — a transform is stale against its data, not only its code.**

`kgm-freshness-check` compared each transform's code-directory commit time
against its output mtime. Nothing modelled the curation files a transform
*reads*, so:

    bacdive/edges.tsv    Aug 12   <- isolation_source_to_ontology.tsv changed Aug 15  -> FRESH
    mediadive/edges.tsv  Aug  5   <- unified chemical SSSOM changed Aug 15            -> FRESH
    madin_etal, metatraits  Aug 1                                                     -> FRESH

Consequence, measured in the shipped graph: 75 edges still asserted organisms
isolated from `NCIT:C16403` "Cell Line", 59 from `mesh:D010520` "Aggressive
Periodontitis", 28 from the wrong tick family — every defect #778 fixed, in a
graph built after #778 merged, with the freshness check reporting `0
STALE_VS_CODE`.

`Transform` now carries `DATA_INPUTS`, a tuple of repo-relative curation files,
declared on the eight consumers (`metatraits_gtdb` inherits from
`metatraits`). The skill derives its view by importing the transform classes
rather than keeping a second list — a table in the skill would drift from the
code it describes, and a stale declaration is indistinguishable from a fresh
one. Two new statuses: `STALE_VS_DATA`, and `STALE_VS_CODE_AND_DATA` so fixing
the one you were told about cannot hide the other.

Commit time, not mtime: these files are tracked, and `git checkout` rewrites
mtimes without changing content (#797).

Verified live — `metatraits_gtdb` now reports

    STALE_VS_DATA ... data input advanced 14.0 days after last output build
    (mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz @ 678a3ee)

The summary block also now prints any status it sees, including ones added
later. It had a hard-coded list of five, so a new category would have been
counted and then omitted from the report — the same class of bug again.

**#813 — an unknown source was skipped in silence.**

`for source in sources: if source in DATA_SOURCES:` had no `else`. A typo'd
`-s` value produced exit 0 and no output, indistinguishable from a successful
run *and* from a transform that died early. That ambiguity blocked a diagnosis:
`kg transform -s bacdive` exited 0 in seconds having written nothing, and
whether dispatch or the transform was at fault could not be told from outside.

Now: refuse the whole batch before running anything, listing valid sources; and
print a start line and a completion line carrying actual row counts, via
`print` rather than `logging.info`, which the CLI never configured a handler
for. A run that produces nothing now says so.

Closes #812
Closes #813

### DIFF
--- a/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
+++ b/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
@@ -48,6 +48,48 @@ from typing import Iterable, Optional
 
 REPO = Path(__file__).resolve().parents[3]
 TRANSFORM_ROOT = REPO / "kg_microbe" / "transform_utils"
+
+
+def _declared_data_inputs(source: str) -> tuple:
+    """
+    Repo-relative curation files a transform reads, from its ``DATA_INPUTS``.
+
+    Read by importing the transform classes rather than by keeping a second
+    list here: a hard-coded table in this skill would drift from the code it
+    describes, and the whole point of #812 is that a stale declaration is
+    indistinguishable from a fresh one. Import failure degrades to "no declared
+    inputs" so the skill still runs in an environment without kg_microbe
+    importable.
+
+    :param source: Transform source name.
+    :return: Tuple of repo-relative paths, empty when none are declared.
+    """
+    try:
+        from kg_microbe.transform import DATA_SOURCES  # noqa: PLC0415 - optional, see docstring
+    except Exception:
+        return ()
+    cls = DATA_SOURCES.get(source)
+    return tuple(getattr(cls, "DATA_INPUTS", ()) or ()) if cls is not None else ()
+
+
+def _latest_data_input_commit(source: str, ref: str) -> tuple[Optional[int], Optional[str]]:
+    """
+    Latest commit time across a source's declared data inputs.
+
+    Uses commit time, not mtime: these files are tracked, and ``git checkout``
+    rewrites mtimes without changing content (#797).
+
+    :param source: Transform source name.
+    :param ref: Git ref to measure against.
+    :return: ``(timestamp, "path @ sha")``, or ``(None, None)``.
+    """
+    newest_ts: Optional[int] = None
+    newest_desc: Optional[str] = None
+    for rel in _declared_data_inputs(source):
+        ts, sha = _latest_commit(REPO / rel, ref)
+        if ts is not None and (newest_ts is None or ts > newest_ts):
+            newest_ts, newest_desc = ts, f"{rel} @ {sha}"
+    return newest_ts, newest_desc
 TRANSFORMED_DIR = REPO / "data" / "transformed"
 MERGE_UTILS_DIR = REPO / "kg_microbe" / "merge_utils"
 DEFAULT_REF = "origin/master"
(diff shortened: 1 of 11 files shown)


realmarcin:

Issue #790, opened 2026-08-15
Three isolation-source mappings are over-narrow: the target excludes part of what the subject names


realmarcin:
Reported from HabitatMech, which consumes `mappings/isolation_source_to_ontology.tsv` and screens it for mappings whose target and label agree with each other while the mapping itself loses part of the subject.

Three rows name a target that is strictly narrower than the source, so an organism isolated from the excluded part is filed under a taxon it does not belong to:

| subject | target | what it excludes |
|---|---|---|
| `Muridae-Mouse/Rat` | `NCBITaxon:10114` **Rattus** | the mouse — Muridae covers both, and *Mus* is not *Rattus* |
| `Reptilia` | `NCBITaxon:8504` **Lepidosauria** | turtles and crocodilians, which are reptiles but not lepidosaurs |
| `Decomposing-algae` | `NCBITaxon:2836` **Bacillariophyta** | every alga that is not a diatom |

These pass the id/label consistency check that caught the earlier batch (#777): each id really is labelled what the row claims. The defect is between the *subject* and the target, which that check cannot see by construction.

Suggested targets, if a broader one is acceptable: `NCBITaxon:10066` Muridae, `NCBITaxon:8504` → `NCBITaxon:8782`-style reptile clade or leaving it unmapped, and a broader algal grouping or unmapped for the third.

Low urgency from our side — we keep all three as xrefs only and never adopt them as a record identity, precisely because they are taxa. Reporting because the same rows presumably feed the graph directly, where the narrowing would be an assertion rather than a note.

realmarcin:
Investigated before fixing, and the situation is different from — and larger than — what this issue describes.

## Only one of the three rows is live

    Muridae-Mouse/Rat  -> NCBITaxon:10114  skos:closeMatch / medium   -> loader drops it
    Decomposing-algae  -> NCBITaxon:2836   skos:closeMatch / medium   -> loader drops it
    Reptilia           -> NCBITaxon:8504   skos:exactMatch / high     -> LIVE

So two of the three never reach the graph at all. Good news for impact; worth knowing before anyone spends time retargeting them.

## The live one is a symptom of something broader

`NCBITaxon:8504` is in the merged graph as `biolink:NamedThing` with **no label**. It is not alone. Of the **22** trusted `NCBITaxon:` targets in `isolation_source_to_ontology.tsv`, **8 are untyped phantoms**:

[...]

PR #815, opened 2026-08-17
fix(bacdive): emit typed stubs for trusted NCBITaxon targets the trim drops

realmarcin:
`_verify_isolation_source_prefixes` permits `NCBITaxon` because the ontologies transform loads it — but that transform emits a **trimmed** taxonomy, so permission at prefix granularity does not imply the specific id is present. 8 of the 22 trusted NCBITaxon targets are not:

    NCBITaxon:6231 Nematoda    NCBITaxon:7898 Fishes
    NCBITaxon:6340 Annelida    NCBITaxon:8015 Salmonidae
    NCBITaxon:6657 Crustacea   NCBITaxon:8292 Amphibia
    NCBITaxon:7712 Tunicata    NCBITaxon:8504 Reptilia

They reach the merged graph as `biolink:NamedThing` with no label. The dangling-edge check cannot see this — KGX materialises a row for any referenced endpoint, so the signal is the missing *category*, not a missing node.

**The ones that work, work by accident.** The trimmed extract contains none of these host taxa — not Mammalia, not Aves, not `Homo sapiens`. The typed ones are typed because ENVO's import closure happens to carry 193 NCBITaxon nodes, `NCBITaxon:40674 Mammalia` among them. Which host taxa resolve should not depend on an unrelated ontology's imports.

Emits `biolink:OrganismTaxon`, not the `OntologyClass` used for mesh/NCIT/BTO stubs — these are taxa, and typing them as an ontology class is what fails the Biolink domain check on `location_of`.

### On #790

Investigating first changed the picture. Only **one** of the three rows that issue reports is live: the Muridae and algae rows are `closeMatch/medium` and already dropped by the trust policy. And retargeting `Reptilia` to a broader reptile clade would not have helped, because that clade is not in the extract either — which is why this is a stub fix rather than a curation fix.

### Testing

The end-state test **skips with a reason** while `data/transformed/bacdive/` is older than either the mapping or `bacdive.py`, and re-arms on the next rebuild. That beats xfail, which needs someone to remember to flip it back. A second test pins the premise, so if the trim ever starts carrying these taxa the stub path gets flagged as dead rather than lingering.

Deliberately **not** triggering a bacdive run now: the pending MediaIngredientMech SSSOM will require a rebuild anyway, and this should ride along rather than cost a second two-hour run.

Refs #790

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -3264,6 +3264,30 @@ class BacDiveTransform(Transform):
                                         mapped_label or isol_source,
                                     )
                                 )
+                            elif stub_prefix == "NCBITaxon" and subject_id not in self.ncbitaxon_labels:
+                                # The prefix guard above permits NCBITaxon because the
+                                # ontologies transform loads it — but that transform
+                                # emits a *trimmed* taxonomy, so permission at prefix
+                                # granularity does not imply the specific id is there.
+                                # 8 of the 22 trusted NCBITaxon targets are absent
+                                # (Nematoda, Annelida, Crustacea, Tunicata, Fishes,
+                                # Salmonidae, Amphibia, Reptilia) and reached the merged
+                                # graph as `biolink:NamedThing` with no label (#790).
+                                #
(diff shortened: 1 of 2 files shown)


realmarcin:

PR #816, opened 2026-08-17
fix(madin_etal): correct two upstream environment ids naming the wrong term

### LINKED ISSUE #796: environments.csv: two rows carry parameter bands to the wrong term (taxonomy root, and a whole organ system)
Reported from HabitatMech, which consumes `data/raw/environments.csv` and attaches each row's physicochemical bands to the term in its `ENVO_ids` column.

Two rows send those bands somewhere they do not describe.

**1. `host_plant_root-associated` → `NCBITaxon:1`**

```
host_plant_root-associated    term_ids: NCBITaxon:1    labels: root
```

`NCBITaxon:1` is the root of the **taxonomy** — the top of the tree of life — matched because its label happens to be "root". The row is about a plant root. `PO:0009005 root` is the term meant.

Downstream this is currently *dropped* rather than misattributed, because we refuse to treat a taxon as a habitat, so the effect is that a real habitat's parameter bands are silently lost. Anything that does not apply that filter would attach root-zone gradients and nutrient levels to the root of all life.

**2. `host_animal_endotherm_nasopharyngeal` → `UBERON:0000065`**

```
host_animal_endotherm_nasopharyngeal    term_ids: UBERON:0000065    labels: respiratory tract
```

The row is specifically nasopharyngeal; `UBERON:0000065` is the entire respiratory tract. So bands measured in the nasopharynx — `Gradients high`, `Structural high` — are asserted about the trachea, bronchi and alveoli too. `UBERON:0001728 nasopharynx` exists and is what the row describes.

Both are the same shape as the id/label problems in #777 and #790, but in a different file: the id and label agree with *each other* here, and neither matches the row's own key.

Not urgent for us — both are corrected locally by curation decisions — but the same rows presumably feed the graph, where the misattribution would be an assertion rather than a note.

realmarcin:
Reported from HabitatMech (#796). Two rows of Madin et al's `environments.csv` carry an id/label pair that agrees with itself while matching neither the row's own key:

    host_plant_root-associated            NCBITaxon:1     "root"
    host_animal_endotherm_nasopharyngeal  UBERON:0000065  "respiratory tract"

`NCBITaxon:1` is the root of the **taxonomy**, matched because its label happens to be "root". The row describes a plant root.

### This is worse here than the reporter assumed

They note the effect is "dropped rather than misattributed", because they refuse to treat a taxon as a habitat. kg-microbe applies no such filter, so it asserted it: **628 edges in the merged graph of the form `NCBITaxon:1 biolink:location_of <organism>`** — the root of the tree of life as a habitat. The nasopharyngeal row contributes 735 edges attributing nasopharyngeal measurements to trachea, bronchi and alveoli.

### Two design points, both nearly got wrong

**Why a TSV rather than a dict in `madin_etal.py`.** `environments.csv` is downloaded, not tracked, so editing `data/raw/` is erased by the next `kg download`. But the correction cannot live in Python either: `stub_curie_collection.py` scans *mapping TSVs* to decide which CURIEs `ontologies_stubs` emits nodes for, and `PO:0009005` is carried by no loaded ontology. A code-only correction would have swapped the taxonomy-root phantom for a PO phantom — the defect just fixed for NCBITaxon in #815. Registering the file in `DEFAULT_MAPPING_PATHS` is what makes the target real, and a test asserts that end to end.

**Why corrections match the wrong value, not just the row key.** If upstream fixes a row, a key-only match would overwrite the *newer* correct value with our stale one. Non-matching rows are reported so the table cannot rot unnoticed.

### Note for whoever rebuilds

`tests/test_ontologies_stubs.py` fails **locally** until `poetry run kg transform -s ontologies_stubs` regenerates `po_nodes.tsv` with `PO:0009005`. It skips in CI, where `data/transformed/` does not exist, so this PR is green. Not run here because transform runs are on hold pending the MediaIngredientMech SSSOM.

Both rows should also be reported upstream to `bacteria-archaea-traits/bacteria-archaea-traits`.

Closes #796

### DIFF
--- a/kg_microbe/transform_utils/madin_etal/madin_etal.py
+++ b/kg_microbe/transform_utils/madin_etal/madin_etal.py
@@ -93,6 +93,65 @@ PARENT_DIR = Path(__file__).resolve().parent
 # quality is not a location.
 _QUALITY_CURIE_PREFIXES = ("PATO:",)
 
+#: Corrections to environments.csv, which is downloaded rather than tracked, so
+#: a fix to data/raw/ is erased by the next `kg download`.
+ENVIRONMENT_ID_CORRECTIONS_FILE = (
+    Path(__file__).resolve().parents[3] / "mappings" / "madin_environment_id_corrections.tsv"
+)
+
+
+def _apply_environment_id_corrections(envo_df: "pd.DataFrame") -> "pd.DataFrame":
+    """
+    Replace upstream ENVO_ids values that name the wrong term.
+
+    Two rows in Madin et al's environments.csv carry an id/label pair that agrees
+    with itself while matching neither the row's own key (#796):
+
+    * ``host_plant_root-associated`` -> ``NCBITaxon:1``, the root of the
+      *taxonomy*, matched on the label "root". The row describes a plant root.
+      This produced 628 merged edges asserting the root of the tree of life is
+      the ``location_of`` an organism.
+    * ``host_animal_endotherm_nasopharyngeal`` -> ``UBERON:0000065``, the whole
+      respiratory tract, so nasopharyngeal measurements were asserted about
+      trachea, bronchi and alveoli too.
+
+    Corrections are matched on the exact wrong value, not applied blindly: if
+    upstream fixes a row, the correction stops firing instead of overwriting the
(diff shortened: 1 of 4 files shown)


realmarcin:

Issue #811, opened 2026-08-16
microbedecoder references 5,242 LPSN ids the lpsn transform does not supply, creating 6,586 untyped phantom organisms


realmarcin:
Surfaced by the full-coverage `kg-model-review` on the 2026-08-16 merge (the violations were invisible at the previous sampling — see #810).

`kg-model-review` reports **21,196 `biolink:capable_of` edges whose subject category is `biolink:NamedThing`** — 75% of all domain/range violations in the graph. Example: `lpsn:20002 -> kgmicrobe.pathway:phototrophy`.

Traced it:

* `microbedecoder` emits **61,459** `lpsn:* --biolink:capable_of--> kgmicrobe.pathway:*` edges.
* It references **25,770** distinct `lpsn:` subjects.
* The `lpsn` transform supplies **34,301** `lpsn:` node rows.
* **5,242 of microbedecoder's subjects are not among them.** No transform emits a node row for `lpsn:20002`; the merged graph materialises it as a bare endpoint.

Result in the merged KG:

    biolink:OrganismTaxon   34,301   properly typed, from the lpsn transform
    biolink:NamedThing       6,586   materialised from edge endpoints only

Those 6,586 are organisms in name only: no label, no taxon typing, no lineage. They are why `capable_of` fails its Biolink domain check — the predicate expects a subject that is some kind of biological entity, and `NamedThing` is not.

**This contradicts the transform's stated contract.** `microbedecoder.py:25` says its first design rule is "attach every edge to the *existing* `lpsn:<LPSN_ID>` node", and CLAUDE.md repeats it: "Emits edges only against existing `lpsn:<LPSN_ID>` nodes — needs the `lpsn` transform present alongside for organism identity." Nothing in the transform reads `data/transformed/lpsn/nodes.tsv`, so the guarantee is documentation, not code.

Worth noting the dangling-edge check does **not** catch this: KGX materialises a node row for every referenced endpoint, so `0 dangling edges` is true and the phantoms are invisible to it. The signal is the category, not the absence.

Two things to decide:

1. **Whether the 5,242 are real LPSN records the lpsn transform is missing, or stale/withdrawn ids in the MicrobeDecoder crosswalk.** That determines whether the fix is to widen `lpsn` or to drop the edges. `lpsn:1074` and `lpsn:11000` are low-numbered ids while supplied ones run to `lpsn:797965`, which hints at an older id series rather than a coverage gap.
[...]

PR #817, opened 2026-08-17
fix(microbedecoder): stub the LPSN taxa lpsn_gss.csv cannot supply

realmarcin:
5,242 `lpsn:` subjects reach the merged graph as untyped `biolink:NamedThing` with no label, producing **21,196 `capable_of` domain violations — 75% of every domain/range violation in the graph** (#811).

### Not a transform bug, and not stale ids

`lpsn_gss.csv` holds exactly **34,301** `record_no` values and the lpsn transform emits all 34,301 — full coverage, no filtering. The missing ids are absent from the export entirely.

### They are all cyanobacteria, and the reason is nomenclatural

    MISSING by phylum:  5242 Cyanobacteriota   (100%, no exceptions)
    MISSING by rank:    4020 Species | 795 Variety | 423 Form | 4 Subspecies

Every id lpsn supplies carries status `VP;` or `VL;` — validly published under the **bacteriological** code. Every missing id carries `Species;` / `Variety;` / `Form;`, which are **botanical**-code ranks. `Variety` and `Form` are not bacteriological ranks at all.

Cyanobacteria are historically named under the ICN rather than the ICNP. LPSN carries both; our GSS export is scoped to the ICNP side, while MicrobeDecoder's crosswalk spans both.

**So dropping the edges — the other option on the table — would have deleted every cyanobacterium this source describes.** A systematic taxonomic hole, not a cleanup.

### The fix keeps the existing rule, conditionally

The documented rule against stubbing cross-referenced entities is preserved: a subject lpsn supplies is still not stubbed. That rule assumes the entity exists in KG-Microbe, and for these 5,242 it does not.

All carry genus and species in the crosswalk (5,242 of 5,242), so the stubs are labelled `biolink:OrganismTaxon` nodes, not bare placeholders.

Fails safe: a missing `data/transformed/lpsn/nodes.tsv` emits **no** stubs and warns, rather than minting 34,301 duplicates. `NA` is filtered as the empty marker. Each id is stubbed once however many strain rows reference it.

### Note

`tests/test_ontologies_stubs.py` fails locally on `PO:0009005` — inherited from #816, cleared by regenerating `ontologies_stubs`. It skips in CI, so this PR is green.

Refs #811

### DIFF
--- a/kg_microbe/transform_utils/microbedecoder/microbedecoder.py
+++ b/kg_microbe/transform_utils/microbedecoder/microbedecoder.py
@@ -308,13 +308,18 @@ class MicrobeDecoderTransform(Transform):
         the add-transform skill's rule on cross-referenced entities.
 
         It *does* emit one for a subject ``lpsn`` cannot supply. The rule
-        assumes the entity exists in KG-Microbe, and for 5,242 of these it does
-        not: ``lpsn_gss.csv`` carries only names validly published under the
+        assumes the entity exists in KG-Microbe, and for **5,763** of these it
+        does not: ``lpsn_gss.csv`` carries only names validly published under the
         **bacteriological** code (every record is ``VP;`` or ``VL;``), while
         MicrobeDecoder's crosswalk also spans the **botanical** code —
-        ``Species;`` / ``Variety;`` / ``Form;``. All 5,242 such ids are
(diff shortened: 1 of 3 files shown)


realmarcin:

PR #818, opened 2026-08-17
feat(gold): drop samples and studies, keep organism→environment, remap retired taxids

realmarcin:
KG-Microbe does not want GOLD's samples or studies. It wants the organism→environment link. This emits that and little else, and fixes the taxid rot that was silently discarding bacteria.

### Samples and studies dropped

`MaterialSample` was 279,670 nodes with 242 incident edges — 279,428 disconnected. `Study` was reachable only via `IndividualOrganism -related_to-> Study`, a predicate asserting nothing usable, at 27.5% of the export. Both node kinds and every edge touching them now go.

`biolink:occurs_in` — the edge the ingest exists for — is untouched, as is the ecosystem `subclass_of` hierarchy.

Measured on the real payload:

    nodes   975,839 -> 637,286
    edges 1,110,984 -> 823,250
    MaterialSample / Study            -> 0
    occurs_in                          286,725  (unchanged)
    orphan warning   279,618 nodes    -> 5

### Retired NCBI taxids remapped

GOLD's export carries taxa NCBI has since merged. Judging them on the retired id makes the trim drop them *and* the organisms typed to them — the opposite of what the trim is for. Applied **before** the trim, since that is the point.

Measured: 100,509 merges loaded; the trim now drops 23,084 taxa rather than 23,695, and 74,561 organisms rather than 76,034 — so **611 taxa and 1,473 organisms are retained** that were previously discarded. 534 nodes collapse onto an existing id where a retired taxon and its replacement were both present upstream.

All three reported cases resolve: `262981→4914` *Lachancea waltii*, `2049935→492670` *Bacillus velezensis*, `2989541→2763025` *Bacteroides parvus*.

NCBI publishes no standalone `merged.dmp`, so the taxdump is downloaded and the single member is read from the tarball without extracting. Absence is non-fatal — the remap becomes a no-op and the affected taxa are dropped as before, which is the pre-existing behaviour rather than a new failure mode.

### Tests changed rather than repaired

Two assertions in `test_gold_transform.py` were **removed**, because the policy they encoded is gone: "a study that still has a microbe stays" and "an upstream orphan sample is reported, not dropped". Neither node kind is ingested now. Flagging explicitly so it is not mistaken for a test being bent to fit.

### Still upstream

Versioned filenames or a published checksum, and the six `Saccharomyces` rows that lose their hybrid `x` marker. Unaffected by this change.

### Note

`tests/test_ontologies_stubs.py` fails locally on `PO:0009005` — inherited from #816, cleared by regenerating `ontologies_stubs`. It skips in CI.

Also worth deciding separately: **`gold` is not in `merge.yaml`**, so none of this reaches the merged graph yet.

### DIFF
--- a/download.yaml
+++ b/download.yaml
@@ -706,3 +706,20 @@
   url: gdrive:1XYVjpPMW-CiuXm8274Ya5PmxZzRSMNxV
   local_name: gold/GOLD_edges.tsv
   tag: gold
+
+#
+# NCBI taxonomy dump — used only for `merged.dmp`, the retired-taxid to
+# current-taxid map. GOLD's export carries ~950 taxa NCBI has since merged
+# (262981 -> 4914 Lachancea waltii, 2049935 -> 492670 Bacillus velezensis,
+# 2989541 -> 2763025 Bacteroides parvus). Without the remap those organisms
+# point at ids no longer in NCBITaxon, so the trim drops them — losing
+# bacteria we would otherwise keep.
+#
+# The whole taxdump is fetched because NCBI publishes no standalone
+# merged.dmp; the transform reads the one member out of the tarball without
+# extracting it. Tagged `gold` so it travels with the source that needs it.
+#
+-
+  url: https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxdump.tar.gz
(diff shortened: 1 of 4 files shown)


realmarcin:

PR #821, opened 2026-08-19
feat(gold): resolve uninformative ecosystems upward, drop root-only edges

realmarcin:
The organism→environment edges are the reason to ingest GOLD, and they were not usable as they stood: **63.5% pointed at a node labelled "Unclassified"**, across **1,663 distinct such nodes** — not a bucket you could filter, but 1,663 semantically empty ids.

### The meaning is in the hierarchy, not the node

The single largest target, `gold.ecosystem:4138`, carries **39,446 organisms** and sits three "Unclassified" hops below something real:

    Unclassified [4138] <- Unclassified <- Unclassified <- Mammals: Human <- Host-associated <- root

So those organisms were found in a **human host**, and the graph said only "Unclassified". A one-hop `occurs_in` query got nothing.

Each such edge now names the nearest informative ancestor and keeps the original target in Biolink's **`original_object`** — the slot meant for what a source said before transformation — so the collapse is auditable and reversible rather than a silent rewrite.

### Root-only edges dropped

58,340 edges resolved only to the hierarchy root, each asserting an organism *lives somewhere*. There is nothing above to recover.

`root` here is the **hierarchy** root, not a plant root. That exact label put physicochemical bands on `NCBITaxon:1` in #796, so it is named in an explicit exclusion set with a test guarding it.

### Measured on the real payload

    occurs_in edges     287,706 -> 229,366
    resolved upward             182,762  (carry original_object)
    dropped as root-only         58,340
    still "Unclassified"              0

    largest targets now: Mammals: Human 39,472 | Fecal 19,721 | Soil 17,984
                         Plants 10,825 | Blood 7,555 | Marine 5,998

### Scope

Only `occurs_in` is resolved. Collapsing `subclass_of` would destroy the structure the resolution depends on — and that hierarchy is what an ENVO crosswalk will map.

### Still open

There is no `gold.ecosystem:` → ENVO crosswalk, so these edges remain an **island**: KG-Microbe models environments in ENVO everywhere else (BacDive isolation sources, Madin habitats, PREGO), and a GOLD environment and a BacDive environment for the same organism never meet. That is the remaining reason not to wire `gold` into `merge.yaml`.

Resolution shrinks the target vocabulary from 4,226 nodes to the named ones, which is what makes the crosswalk tractable.

### Note

`test_schema_is_conformed` now asserts the standard edge header **plus** `original_object`, spelled out rather than loosened to a prefix match, so a future stray column still fails it.

`tests/test_ontologies_stubs.py` fails locally on `PO:0009005` — inherited from #816, cleared by regenerating `ontologies_stubs`. Skips in CI.

### DIFF
--- a/download.yaml
+++ b/download.yaml
@@ -723,3 +723,19 @@
   url: https://ftp.ncbi.nlm.nih.gov/pub/taxonomy/taxdump.tar.gz
   local_name: taxdump.tar.gz
   tag: gold
+
+#
+# GOLD ontology — the JGI ecosystem classification as OWL, with curated ENVO
+# mappings (`skos:exactMatch` plus the MIxS env_broad / env_local / env_medium
+# triad). Sourced from the canonical GitHub repo rather than BioPortal, which
+# gates its API behind a key; OLS4 lists this same file as the ontology's
+# location.
+#
+# Without it GOLD's `gold.ecosystem:` vocabulary is an island: KG-Microbe models
+# environments in ENVO everywhere else, so a GOLD environment and a BacDive
+# environment for the same organism never meet.
+#
+-
+  url: https://raw.githubusercontent.com/cmungall/gold-ontology/refs/heads/main/gold.owl
+  local_name: gold_ontology.owl
+  tag: gold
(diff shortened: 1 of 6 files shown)


realmarcin:

Issue #860, opened 2026-08-21
Split ontology_utils into explicit loading, resolution, and validation components


realmarcin:
**Priority: P3 — staged maintainability work after current correctness blockers.**

## Problem

`kg_microbe/utils/ontology_utils.py` is approximately 2,981 lines and combines ontology loading, lookup/index management, identifier and label resolution, mapping behavior, validation, and cache concerns. The broad utility surface encourages unrelated transforms to depend on internal details.

## Proposed change

Separate ontology access, index/cache management, resolution policy, mapping helpers, and validation behind small typed interfaces. Keep compatibility shims only long enough to migrate callers deliberately.

## Acceptance criteria

- [ ] A dependency/caller inventory and design note define the new boundaries before code moves.
- [ ] Loading/cache code is separate from resolution policy and transformation-specific mapping logic.
- [ ] Resolution behavior has focused tests for exact labels, synonyms, obsolete terms, ambiguity, and missing terms.
- [ ] Existing callers migrate incrementally with no circular imports.
- [ ] Golden/snapshot comparison shows no unintended graph changes.
- [ ] Network and cache behavior are explicit and injectable in tests.

Issue #859, opened 2026-08-21
Split the BacDive transform into tested responsibility-focused modules


realmarcin:
**Priority: P3 — staged maintainability work after current correctness blockers.**

## Problem

`kg_microbe/transform_utils/bacdive/bacdive.py` is approximately 3,448 lines and combines API/input parsing, normalization, ontology resolution, domain-specific edge construction, writing, and orchestration. The size and mixed responsibilities make BacDive correctness fixes hard to isolate.

## Proposed change

Split the implementation around stable responsibilities such as record parsing, strain/taxon normalization, ontology and mapping resolution, assay/trait modeling, KG emission, and orchestration.

## Acceptance criteria

- [ ] A short design note defines module boundaries and allowed dependencies before the mechanical split.
- [ ] Parsing and normalization can be tested without network access or a full BacDive download.
- [ ] Domain-specific emitters have focused fixture-based tests.
- [ ] The existing public transform interface remains compatible or has a documented migration.
- [ ] Golden/snapshot comparison shows no unintended node or edge changes.
- [ ] The original module becomes a small orchestration/compatibility layer rather than a second implementation.

Issue #858, opened 2026-08-21
Split the MetaTraits transform into tested responsibility-focused modules


realmarcin:
**Priority: P3 — staged maintainability work after current correctness blockers.**

## Problem

`kg_microbe/transform_utils/metatraits/metatraits.py` is approximately 3,732 lines and combines parsing, mapping/ontology resolution, graph construction, file writing, and orchestration. This makes focused tests and safe changes difficult.

## Proposed change

Split the implementation around stable responsibilities while preserving the public transform entry point and output contract. Candidate boundaries include input parsing, normalized records, ontology/mapping resolution, node/edge emission, validation, and orchestration.

## Acceptance criteria

- [ ] A short design note defines module boundaries and allowed dependencies before the mechanical split.
- [ ] Parsing, resolution, and emission logic have direct unit tests independent of a full transform run.
- [ ] The existing public transform interface remains compatible or has a documented migration.
- [ ] Golden/snapshot comparison shows no unintended node or edge changes.
- [ ] No new import-time data loading or network access is introduced.
- [ ] The original module becomes a small orchestration/compatibility layer rather than a second implementation.

Issue #839, opened 2026-08-20
ontologies_stubs declares 1 of the 11 mapping files it reads — the #812 blind spot, still open


realmarcin:
Found while auditing which transforms need re-running.

`OntologiesStubsTransform.DATA_INPUTS` declares one file:

```python
DATA_INPUTS = ("mappings/isolation_source_to_ontology.tsv",)
```

It reads eleven, via `stub_curie_collection.DEFAULT_MAPPING_PATHS`:

```
kgmicrobe_unified_entity_mappings.sssom.tsv.gz
isolation_source_to_ontology.tsv          <- the only declared one
madin_environment_id_corrections.tsv
ingredient_mappings.sssom.tsv
canonical/chemical_mappings.tsv
canonical/enzyme_mappings.tsv
canonical/enzyme_name_to_go.tsv
canonical/metpo_alias_mappings.tsv
canonical/pathway_mappings.tsv
canonical/phenotype_mappings.tsv
canonical/special_chemical_mappings.tsv
```

`kgm_freshness_check` reads `DATA_INPUTS` and nothing else, so a change to any of the other ten leaves the transform reported as fresh when it is not.

## This is the defect #812 was filed for, in the one transform that still has it

A sweep of every transform declaring `DATA_INPUTS` finds this is now the only one. The others (bacdive, mediadive, madin_etal, metatraits, microbedecoder, ctd) declare what they read.

It is not hypothetical here. The currently-failing `test_every_referenced_curie_has_stub_node` is about `PO:0009005`, which enters through `madin_environment_id_corrections.tsv` — an undeclared input. The freshness check reports `ontologies_stubs` as `STALE_VS_CODE`, attributing it entirely to code, when a data input moved too.

It also changes the answer to "what can I re-run independently of the incoming MIM SSSOM". `DEFAULT_MAPPING_PATHS[0]` **is** the unified SSSOM, so `ontologies_stubs` is SSSOM-dependent and will need re-running again after that lands — which is not what its declaration implies.

## Fix

Declare the same list, from the same constant, rather than restating it — `DATA_INPUTS` should be derived from `DEFAULT_MAPPING_PATHS` so the two cannot drift. A test that the declared set covers everything `collect_stub_curies` opens would keep it honest.

PR #862, opened 2026-08-21
Remediate release, offline CLI, test, and maintainability findings

### LINKED ISSUE #534: Clean up .gitignore and set a policy for generated/downloaded data
## Problem

`.gitignore` is 30,288 lines. ~30,000 of those are individual file paths rather than patterns. There are duplicates, mismatches (e.g., `mediadive_cache.sqlite` ignored but `mediadive_bulk_cache.sqlite` was not), and no clear policy for what belongs.

Only 39 lines use glob patterns. The rest appear to have accumulated from `git add`-ing specific paths over time.

### Ignore by pattern, not by path
Replace the ~30,000 specific file paths with directory-level ignores:

```gitignore
# Downloaded data — regenerable via `kg download`
data/raw/

# Transformed output — regenerable via `kg transform`
data/transformed/

# Merged output — regenerable via `kg merge`
data/merged/

# SQLite caches (requests_cache, oaklib, etc.)
*.sqlite
*.sqlite-journal

# Legacy/snapshot directories
data/raw_last_local/
data/raw_pre*/
data/transformed_*/
```

### Keep tracked
- `download.yaml`, `merge.yaml`, `merge.minimal.yaml` — pipeline config
- `tests/resources/` — test fixtures
- Source code, templates, mappings TSVs that are hand-curated

### Transition plan
1. Verify no checked-in data files are referenced by imports or tests (vs. downloaded at runtime)
2. Replace the 30k specific paths with the pattern-based policy above
3. Run `git rm --cached` on any tracked files that should be ignored
4. Document the policy at the top of `.gitignore`

## Context

- Current file: 30,288 lines, ~30,000 are specific `data/` paths
- Only 39 lines use glob patterns
- 2 known duplicates (`data/raw/.keep`, `mediadive_cache.sqlite`)
- Fixed one mismatch in https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/527 (`.gitignore` had `mediadive_cache.sqlite` but code creates `mediadive_bulk_cache.sqlite`)
- Related: https://github.com/Knowledge-Graph-Hub/kg-microbe/issues/533 (download ergonomics)

### LINKED ISSUE #543: Add tests for DuckDB query stack and kg query-organism CLI
## Context

Surfaced in Copilot's 2026-04-16 review of #531.

PR #531 added a new `kg query-organism` CLI subcommand and a `kg_microbe/query_utils/` module (DuckDB loader, organism queries). None of the new functionality has automated tests. This is a gap that (a) lets correctness bugs sneak in unnoticed — see the predicate/relation inconsistency and CRLF bug filed alongside this issue — and (b) makes later refactors risky.

## Suggested minimum coverage

1. **Name resolution** (`find_organism_by_name`):
   - 0 matches → documented behavior (raise or return a sentinel)
   - 1 match → returns the row
   - many matches → documented behavior (raise with candidate list, or return selected + alternatives — see the contract-mismatch issue)

2. **Media preference retrieval** (`get_media_preferences`): confirm the chosen predicate/relation encoding actually returns rows on the expected fixtures. Ideally blocked on resolving the predicate/relation inconsistency issue first so the test isn't written against the wrong schema.

3. **Media composition retrieval** (`get_media_composition`): assert correct rows for a list of medium IDs; use this as the test that pins the parameterized-query fix.

4. **DuckDB loader** (`duckdb_loader.py`): round-trip a small CRLF TSV fixture and assert no trailing `\r` in the last column (pins the CRLF fix).

5. **CLI smoke test**: invoke `kg query-organism` on a tiny fixture and assert the exit code + key output lines.

Use a small in-memory DuckDB database or a tiny on-disk fixture so tests are hermetic and fast.

## Files involved

- `kg_microbe/query_utils/` (new module)
- `kg_microbe/run.py` (`query-organism` CLI entry)
- `tests/` (new test files)

## References

- [PR #531](https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/531)
- Copilot review at commit `1de973d`, 2026-04-16T23:15Z

### LINKED ISSUE #553: Release asset contents and download paths are undocumented
## What's missing

The release assets (e.g. on the `2025-03-07` release) include multiple tarballs:

- `kg-microbe-core.tar.gz` (112 MB)
- `kg-microbe-biomedical.tar.gz` (185 MB)
- `madin_etal.tar.gz` (1 MB)
- `ontologies.tar.gz` (75 MB)
- and others

The following information cannot be found in the repo README, release notes, or documentation:

1. What each tarball contains (which nodes/edges, which sources are merged)
2. What predicates connect taxa to phenotypes
3. A changelog between releases
4. A recommended stable URL pattern for programmatic download

## Why this matters

Downstream consumers need to know which tarball to download and what's in it. Without documentation, the only option is to download all of them and inspect the contents.

### LINKED ISSUE #847: All six merge configs write to the same merged-kg.tar.gz, so the last run silently wins
Found reviewing #846, while checking that the `merge.no_metatraits.yaml` rebuild had not changed its destination. It had not — but every config shares one:

| config | writes |
|---|---|
| `merge.yaml` | `data/merged/merged-kg.tar.gz` |
| `merge.minimal.yaml` | `data/merged/merged-kg.tar.gz` |
| `merge.no_metatraits.yaml` | `data/merged/merged-kg.tar.gz` |
| `merge.noprego.yaml` | `data/merged/merged-kg.tar.gz` |
| `merge.prego-full.yaml` | `data/merged/merged-kg.tar.gz` |
| `merge_bakta.yaml` | `data/merged/merged-kg.tar.gz` |

Running `kg merge -y merge.minimal.yaml` — a deliberately tiny 3-source graph — overwrites the full 2.85M-node graph in place, and **nothing on disk records which config produced the artifact**. A 3-source graph and a 25-source graph are the same filename, in the same directory, distinguished only by size.

## Why this is the #828 shape, sharpened

#828 was about a stale artifact that looked current. This is worse: an artifact that looks like *the standard graph* and is a variant. `merged_graph_stats.yaml` is likewise shared, so the stats file describes whichever ran last.

It is also a plausible explanation for artifacts like the `merged-kg_default_*` TSVs in #828 — someone's variant run, left behind.

The warning added in #846 helps only when the filenames differ; a same-name overwrite leaves nothing to warn about.

## Fix

Give each config a distinct `filename` (`merged-kg-no-metatraits`, `merged-kg-minimal`, …), or write the config name and its source list into the output directory alongside the archive so an artifact can be identified after the fact. The second is cheap and also answers "what is this file" for artifacts that already exist.

Not folded into #846: that PR is about config drift and stale siblings, and renaming release artifacts is a change with downstream consequences — anything scripted against `merged-kg.tar.gz` would need updating, which is a decision rather than a fix.

### LINKED ISSUE #849: Stop release tooling from exposing GH_TOKEN or deleting unrelated archives
**Priority: P0 — complete before the next release.**

## Problem

`kg-microbe.Makefile` defines a `git_remote` target that embeds `GH_TOKEN` in the Git remote URL before release commands run. The expanded command can appear in Make output, and the credential remains in `.git/config` after the command completes.

The same release path uses broad `rm -f *.tar.gz` cleanup, which can remove archives that were not created by the current run.

## Proposed change

- Remove the `git_remote` target and its dependency from release targets.
- Let `gh` authenticate through its normal `GH_TOKEN` environment handling without modifying `origin`.
- Track the exact archives created by the current run and remove only those files when cleanup is required.
- Keep release logs free of credential-bearing URLs and token values.

## Acceptance criteria

- [ ] Running the release or prerelease target never changes `git remote get-url origin`.
- [ ] No token is written to `.git/config`, the package tree, command output, or generated artifacts.
- [ ] Release publication works with `GH_TOKEN` supplied only through the process environment.
- [ ] Cleanup names an explicit set of generated archives and preserves unrelated `*.tar.gz` files.
- [ ] A regression test or safe dry-run check covers both remote preservation and cleanup scope.

## Files

- `kg-microbe.Makefile`
- Any release tests or scripts introduced by the fix

### LINKED ISSUE #850: Make CLI startup and command discovery offline-safe
**Priority: P0 — this blocks reliable CLI use and hermetic tests.**

## Problem

`kg_microbe/run.py` eagerly imports KGX merge code before Click dispatches a command. KGX initializes BMT, which can attempt to fetch a Biolink schema. With network access disabled, even `kg --help` and unrelated subcommands can fail during import.

The repository already downloads a pinned schema to `data/raw/biolink-model.yaml`, but startup does not consistently use it.

## Proposed change

- Move KGX/BMT-dependent imports into the commands that need them.
- Construct the transform registry lazily instead of importing every transform during CLI startup.
- Configure KGX/BMT to use the repository's pinned `data/raw/biolink-model.yaml`.
- Ensure command discovery and test collection do not initiate network requests.

## Acceptance criteria

- [ ] With network access blocked, `kg --help` exits 0.
- [ ] With network access blocked, help for commands unrelated to KGX/BMT exits 0.
- [ ] Importing `kg_microbe.run` does not fetch a schema or initialize remote-backed services.
- [ ] Merge/transform paths use the pinned local Biolink schema, with a clear error if it is missing.
- [ ] Regression tests fail if CLI startup or unit-test collection attempts network access.

## Files

- `kg_microbe/run.py`
- `kg_microbe/merge_kg.py`
- Transform registry/import code
- CLI tests

### LINKED ISSUE #851: Implement kg holdouts or remove the silent no-op command
**Priority: P1 — resolve the silent-success behavior before advertising this command.**

## Problem

`kg holdouts` is registered in `kg_microbe/run.py`, but its implementation call is commented out. The command returns success without producing output. `CLAUDE.md` currently presents it as functional.

A successful no-op is especially risky in a pipeline because downstream work can proceed using missing or stale holdout artifacts.

## Decision required

Choose one supported outcome:

1. Implement the command and restore tested holdout generation; or
2. Remove the command and its documentation until there is a maintained implementation.

## Acceptance criteria

- [ ] The CLI can no longer exit 0 while producing no expected holdout output.
- [ ] If implemented, the command validates inputs, creates documented artifacts, and fails nonzero on generation errors.
- [ ] If retired, the Click command and every documented invocation are removed together.
- [ ] A regression test specifically covers the current success-with-no-output failure mode.
- [ ] `CLAUDE.md`, `README.md`, and CLI help agree on whether holdouts are supported.

## Files

- `kg_microbe/run.py`
- Holdout implementation, if retained
- CLI tests
- User and agent documentation

### LINKED ISSUE #852: Load graph TSVs directly into DuckDB with atomic rebuilds and freshness checks
**Priority: P1 — required for query support at advertised graph sizes.**

## Problem

`kg_microbe/query_utils/duckdb_loader.py` reads complete node and edge TSV files into Pandas before copying them into DuckDB. This places the full source graph in Python memory, so DuckDB's configured memory limit does not protect the process. The approach is unsuitable for graphs with tens or hundreds of millions of records.

The loader also reuses any existing database without checking whether its source TSV files changed, and it writes the final database in place.

## Proposed change

- Ingest TSVs directly with DuckDB `read_csv`/`COPY` using explicit column schemas.
- Build a new database at a temporary sibling path and atomically rename it after successful validation.
- Store and compare source fingerprints before reuse. At minimum include path, size, and modification time; use content hashes where correctness requires them.
- Leave the last valid database intact when ingestion fails.

## Acceptance criteria

- [ ] Node and edge ingestion does not materialize the complete TSVs in Pandas.
- [ ] Schemas are explicit and preserve KGX list/scalar conventions and null behavior.
- [ ] A changed source TSV invalidates or rebuilds the database.
- [ ] An unchanged source set reuses the database without reloading it.
- [ ] A failed build cannot leave a partial database at the advertised output path.
- [ ] Tests cover freshness, atomic replacement, malformed input, and a bounded-memory ingestion path.

## Related issues

- #543 — DuckDB/query CLI test coverage
- #541 — CRLF parsing correctness

### LINKED ISSUE #853: Repair release and documentation workflows around named build artifacts
**Priority: P1 — release and documentation automation are not currently dependable.**

## Problem

The release workflow archives `data/transformed`, but the workflow neither builds that directory nor downloads a build artifact. Its publication step only runs for tag refs even though the workflow is normally manually dispatched.

The documentation workflow listens on `main`, while this repository's default branch is `master`. It also runs `poetry add` during CI and writes `GH_TOKEN` into the package tree.

## Proposed change

- Make the build pipeline publish a named, versioned graph artifact.
- Make the release workflow consume that artifact rather than assuming local transformed files exist.
- Define coherent tag and manual-dispatch behavior, including an explicit version/release target for manual runs.
- Trigger documentation deployment from `master`.
- Install the committed dependency set; never mutate `pyproject.toml` or the lockfile during CI.
- Pass credentials only to the steps that require them and never write them into the checkout.
- Use least-privilege workflow permissions.

## Acceptance criteria

- [ ] A release job fails clearly when the named build artifact is absent or invalid.
- [ ] Tag and manual-dispatch paths are both tested or reduced to one documented supported path.
- [ ] The released archives are traceable to a specific build/configuration.
- [ ] Documentation deployment runs from `master` with locked dependencies.
- [ ] Workflow execution creates no token file and changes no dependency metadata.
- [ ] Workflow syntax/static validation is part of CI.

## Related issues

- #553 — document release asset contents and download paths
- #847 — distinguish merge variants and their outputs

### LINKED ISSUE #854: Fix Poetry metadata and align CI with supported Python versions
**Priority: P1 — packaging metadata should be valid before dependency or release work.**

## Problem

`poetry check` fails because the `docs` extra references packages that are defined only in the optional `docs` dependency group. The extra is obsolete under the current Poetry group layout.

The project declares Python `^3.10`, but CI does not establish support for the full declared range, including Python 3.13 and later.

## Proposed change

- Keep the optional `docs` dependency group and remove the invalid/obsolete extra.
- Standardize documentation setup on `poetry install --with docs`.
- Add `poetry check` to CI.
- Either test currently declared Python versions or constrain the declaration to the versions the project actually supports.

## Acceptance criteria

- [ ] `poetry check` passes from a clean checkout.
- [ ] Documentation dependencies install with `poetry install --with docs`.
- [ ] No CI job mutates dependency metadata to make a build pass.
- [ ] The CI Python matrix and `requires-python`/Poetry declaration describe the same supported range.
- [ ] The support decision is noted in contributor documentation.

## Related issues

- #523 — remove stale Black configuration after the Ruff migration
- #524 — align pre-commit and development dependency versions

### LINKED ISSUE #855: Make unit tests hermetic and align Biolink schema versions
**Priority: P1 — required before the suite can be trusted as a release gate.**

## Problem

Unit-test collection can trigger KGX/BMT schema downloads. In addition, `tests/conftest.py` downloads mutable content from the `metpo/main` branch into `data/raw` during test setup. This makes results dependent on the network and upstream branch state, and lets tests modify repository data.

Biolink versions are also inconsistent: `download.yaml` pins schema 4.3.6, the locked package has used 4.4.2, and BMT may use its remote default.

## Proposed change

- Commit small immutable fixtures needed by unit tests, or pin fixture downloads by immutable version and reserve them for integration tests.
- Run unit tests in temporary directories without writes to `data/raw`.
- Disable network access in the unit-test job.
- Mark network/large-data tests explicitly and run them in a separate integration job.
- Select one supported Biolink version and configure schema consumers consistently.

## Acceptance criteria

- [ ] Unit-test collection and execution pass with network access disabled.
- [ ] Unit tests do not write into repository data directories.
- [ ] No unit fixture follows an upstream `main`/`master` branch.
- [ ] Integration tests are explicitly marked and excluded from the default unit job.
- [ ] `download.yaml`, locked dependencies, KGX, and BMT use one documented Biolink version.
- [ ] CI has a guard that detects unexpected outbound network access in the unit job.

## Related work

Coordinate with the offline-safe CLI startup issue so import-time network access is removed rather than hidden by fixtures.

### LINKED ISSUE #856: Reduce CLAUDE.md to durable guidance and add a tool-neutral AGENTS.md
**Priority: P2 — do after CLI/workflow behavior is settled so documentation records stable contracts.**

## Problem

`CLAUDE.md` is 457 lines and mixes durable agent rules with volatile measurements, historical debugging notes, and duplicated environment configuration. Several statements are stale or unverifiable:

- `kg holdouts` is documented although it is currently a no-op.
- The SPARQL example points to missing `queries/sparql/example_query.yaml`.
- The test inventory names files that no longer exist.
- The standard edge schema omits `knowledge_level` and `agent_type`.
- Formatting guidance says Black although the repository uses Ruff formatting.
- Type hints are described as required but CI does not enforce that rule.
- `make run-summary` expects loose TSVs while the default merge produces an archive and removes extracted files.

## Proposed change

- Reduce `CLAUDE.md` to roughly 100–150 lines of commands, invariants, safety rules, and links.
- Add a concise, tool-neutral `AGENTS.md` containing critical repository conventions.
- Move PREGO measurements and ontology failure history into versioned ADR/runbook documents.
- Put supported environment variables, defaults, and warnings in `.env.example`; link to it instead of duplicating the list.
- Validate every referenced command, path, test file, and schema field.

## Acceptance criteria

- [ ] `CLAUDE.md` contains only durable operational guidance and is approximately 100–150 lines.
- [ ] `AGENTS.md` exposes the critical conventions without Claude-specific assumptions.
- [ ] Detailed historical/measurement material remains available in linked ADRs or runbooks.
- [ ] `.env.example` is the canonical inventory of supported environment variables and contains no secrets.
- [ ] Holdout, SPARQL, test inventory, edge schema, formatter, typing, and run-summary statements match repository behavior.
- [ ] Documentation link/path checks cover the files referenced by agent guidance.

## Related issues

- #523 — remove obsolete Black configuration
- #553 — release artifacts and download documentation

### LINKED ISSUE #857: Generate merge YAML variants from one canonical specification
**Priority: P2 — reduce configuration drift after output naming is resolved.**

## Problem

The six closely related `merge*.yaml` files are maintained as near-copies. Recent drift fixes show that source additions and option changes can be applied to one variant while silently missing others.

## Proposed change

- Define one canonical merge specification containing the shared source set and settings.
- Express each variant as a small, reviewed set of additions/removals/overrides.
- Generate the committed runtime YAML files deterministically.
- Add a CI check that regeneration produces no diff.
- Coordinate generated filenames and provenance metadata with #847 so variants cannot overwrite each other.

## Acceptance criteria

- [ ] Shared merge configuration is defined once.
- [ ] Variant differences are explicit and reviewable without diffing full near-copies.
- [ ] Generation is deterministic and documented.
- [ ] CI fails when generated YAML is stale.
- [ ] Existing merge variants retain their intended source sets, with regression tests for key inclusions/exclusions.
- [ ] Output names/provenance are compatible with the resolution of #847.

## Related issues

- #847 — all merge variants currently overwrite `merged-kg.tar.gz`
- #827 — prior drift in `merge.no_metatraits.yaml` (closed)

### LINKED ISSUE #865: PR #862 removes 336 lines of CLAUDE.md operational knowledge that is not relocated anywhere
Adversarial review of #862.

`CLAUDE.md` goes from **457 lines to 121**, and every existing heading is replaced. The PR summary describes this as "refreshes README, CLAUDE.md, `.env.example`, and adds tool-neutral AGENTS.md and runbooks", which reads like relocation. I checked whether the content moved, and largely it did not.

Searched every `.md` in the PR tree:

| documented thing | before | after | found elsewhere in the tree? |
|---|---:|---:|---|
| `PREGO_SHAPES` | 2 | 0 | **nowhere** |
| `PREGO_MIN_CONFIDENCE` | 2 | 0 | **nowhere** |
| "sits under this taxon" convention | 1 | 0 | **nowhere** |
| `GOLD_APPLY_TAXON_TRIM` | 1 | 0 | only `docs/GOLD_TRANSFORM_REVIEW.md`, a dated point-in-time review |
| `KG_MEDIADIVE_ALLOW_STALE_CACHE` | 1 | 0 | only a `TODO/` review prompt |
| `KG_SEMSQL_BUILD` | 2 | 0 | partially — `docs/runbooks/ontology-caches.md` |
| `merge.noprego` guidance | 3 | 0 | — |
| `atomic_write` guidance | 1 | 0 | — |

Sections removed outright include **"Ontology failures abort; they do not degrade"** (why `FatalOntologyError` derives from `BaseException`, why you must not wrap adapter use in `except Exception`, and why it must never be resolved inside a `multiprocessing.Pool` worker) and **"Adding a New Transform"**.

## Why this is more than a docs tidy

Several of these entries exist *because* something went wrong, and they encode the trap rather than the feature:

- **`PREGO_SHAPES`** documented that `merge.yaml` reads `data/transformed/prego_habitat/` and the transform default must agree, because **KGX silently skips a missing input** — a default transform writing `prego/` against a default merge reading `prego_habitat/` produced a graph with no PREGO and no warning. That incident's lesson is now nowhere.
- **The taxon convention** records a deliberate, graph-wide deviation from Biolink's letter across ~1.65M edges (`OrganismTaxon` + `subclass_of`), added yesterday in #832 precisely so the next person reaching for `in_taxon` finds the reasoning instead of rediscovering the split.
- **The ontology-abort section** tells you why an apparently reasonable `except Exception` produces a systematically wrong graph with a zero exit code.

`CLAUDE.md` is also loaded into context automatically, so its contents are not equivalent to the same text in `docs/` — moving guidance there changes whether it is seen at all.

## Suggested resolution

Not "revert the rewrite" — the new structure and AGENTS.md are reasonable. But the operational invariants above should survive somewhere loaded by default, and the ones tied to a past incident should keep the incident with them.

### LINKED ISSUE #866: PR #862's DuckDB cache keys on mtime+size, reintroducing the staleness class #844 just closed
Adversarial review of #862.

`kg_microbe/query_utils/duckdb_loader.py` decides whether to reuse an existing DuckDB or rebuild it:

```python
def _source_fingerprints(nodes_path, edges_path):
    """Return stable-enough freshness metadata without scanning multi-GB files."""
    for kind, path in (("nodes", nodes_path), ("edges", edges_path)):
        stat = path.stat()
        result[kind] = (str(path), stat.st_size, stat.st_mtime_ns)
```

This is the timestamp class #844 closed yesterday, in a new place. It has zero references to `kg_microbe/utils/transform_fingerprint.py`, the shared content-fingerprint utility that exists for exactly this decision.

## Two failure directions, one of which is silent

**Wasteful but safe:** `git checkout` rewrites the merged TSVs' mtime with no content change (#797), forcing a full rebuild of multi-GB files. Annoying, not wrong.

**Silent and wrong:** size plus mtime can both be preserved across a genuine content change. `cp -p`, `rsync --times`, a restore from backup, or an artifact re-extracted from a tarball can all land different bytes with the recorded mtime and an identical size — merged TSVs of similar vintage are very close in size. The database is then judged current and **serves stale query results with no indication**. A wrong answer from `kg query` is materially worse than a slow one.

## The tradeoff is real and I am not asking to hash gigabytes

The docstring is candid — "stable-enough … without scanning multi-GB files" — and hashing the merged TSVs on every CLI startup would be unacceptable. But there are cheaper options that do not key on the clock:

1. Have the **merge** write a fingerprint beside its output, the way transforms now do (#844), and have DuckDB record *that*. The merge computes it once; the query path only compares strings.
2. Hash a bounded prefix plus the size and the row count. Still cheap, and defeats the same-size collision.
3. If mtime is kept, record it **alongside** a content signal rather than as the sole one.

Related: #845 (nothing detects staleness against another transform's output) is the same family. Worth deciding once where the repo's "has this input changed" answer lives, rather than a third mechanism appearing next.

### LINKED ISSUE #867: PR #862 removes the holdouts CLI command but leaves three references to it
Adversarial review of #862.

Removing `kg holdouts` is right: the body was already commented out (`# make_holdouts(*args, **kwargs)`), so the CLI advertised a command that did nothing. That is #851's point and the PR fixes it.

But three places still tell someone to run it:

```
.claude/skills/codex-review-kg-microbe/SKILL.md
.claude/skills/kg-postprocess-report/SKILL.md
.claude/skills/kg-postprocess-report/kg_postprocess_report.py
tests/test_duckdb_query_cli.py
```

The two `SKILL.md` files are instructions an agent follows. After this PR they direct it to invoke a command that no longer exists, which fails at a point far from the cause. `kg_postprocess_report.py` is executable code referencing it.

Worth checking each: some may be describing `data/holdouts/` as a directory rather than the command, which is fine. The ones that name the *command* need updating or removing in the same change — a removal that leaves callers behind converts a documented no-op into an undocumented crash.

Also worth a line in the PR body or a release note: this is a user-visible CLI removal, even if what it removed did nothing.

### LINKED ISSUE #868: PR #862 pins Biolink to 4.4.2 while docs and the model-review skill still say 4.2.2
Adversarial review of #862.

The PR pins `biolink-model = "4.4.2"` and points the fetched YAML at `v4.4.2`, which is a good change — it makes BMT initialization offline-safe and deterministic. Two loose ends.

## 1. Three places still say 4.2.2

```
.claude/skills/kg-model-review/SKILL.md:53   "Biolink Model 4.2.2 loaded (154 categories, 469 predicates)"
docs/GOLD_TRANSFORM_REVIEW.md:5              Method: ... Biolink Model 4.2.2 via bmt
docs/GOLD_UPSTREAM_REQUESTS.md:14            Every category and predicate is a real Biolink 4.2.2 element
```

The two `docs/` files are dated point-in-time reviews and arguably should keep the version they were run under — amending them would falsify the record. The `SKILL.md` line is different: it is an example of expected output for a skill someone runs *now*, and it will no longer match.

## 2. The verdicts that depend on the version are unverified

Some load-bearing conclusions in this repo were derived from `bmt` under 4.2.2 and are not re-checked by this PR:

- **`biolink:OrganismTaxon` is not an `ontology class`.** #832 conformed 471,702 GOLD organism edges to `subclass_of` on the reasoning that the domain/range objection applies equally to NCBITaxon's own 925,219 hierarchy edges, so consistency beats conformance. If 4.4.2 changed `OrganismTaxon`'s ancestry or `subclass_of`'s domain, that argument's premise moves.
- **`biolink:located_in` vs `occurs_in`** — #821 re-predicated 229,366 GOLD edges on the 4.2.2 definitions.
- `docs/GOLD_TRANSFORM_REVIEW.md` records a domain/range audit whose "VIOLATION" and "OK" verdicts were computed under 4.2.2.

Not asserting any of these changed — I have not run 4.4.2. The ask is that the bump include a re-run of `kg-model-review` (or a short `bmt` check of those specific elements) so the version change is known to be behaviour-preserving rather than assumed to be. A pin that silently alters domain/range verdicts is the kind of thing that shows up much later as an unexplained diff.

### LINKED ISSUE #869: PR #862 release workflow interpolates workflow_dispatch inputs into shell, including inside the step meant to validate them
Second adversarial review pass on #862. Complements #865–#868; this one is in `.github/workflows/release.yml`.

The rewritten release workflow is a clear improvement overall — no secrets on command lines, artifact cleanup narrowed from `rm -f *.tar.gz` (which would delete any tarball in the directory, including one a user put there) to a manifest-driven removal. One pattern is worth changing before it lands.

## The issue

`${{ inputs.* }}` is interpolated into `run:` blocks:

```yaml
- name: Validate inputs
  shell: bash
  run: |
    set -euo pipefail
    [[ "${{ inputs.build_date }}" =~ ^[0-9]{8}$ ]]
    [[ "${{ inputs.release_tag }}" =~ ^[A-Za-z0-9._-]+$ ]]
```

and again at the provenance step:

```yaml
    printf 'build_date=%s\n' "${{ inputs.build_date }}"
```

GitHub expands `${{ }}` **before** the shell parses the script, so the input becomes part of the program text rather than a value. A `build_date` of

```
x"; whoami; :"
```

renders the guard as `[[ "x"; whoami; :"" =~ ^[0-9]{8}$ ]]` and executes `whoami`.

**The regex never runs.** The step that exists to reject malformed input is itself the first place the input is executed, so validation cannot protect anything downstream of it — including its own step.

## Severity

Low likelihood, and I do not want to overstate it: `workflow_dispatch` requires write access to the repository, so anyone who can trigger this can already push. `permissions: contents: write` bounds the blast radius to this repo.

It is worth fixing anyway because it is the exact pattern GitHub's own hardening guidance names, it is free to fix, and the misleading part — a validation step that can be bypassed by the thing it validates — is the kind of construct that gets copied into a workflow where the trigger *is* reachable by an outside contributor.

## Fix

Bind inputs to `env:` and reference the shell variable, so the value is never program text:

```yaml
- name: Validate inputs
  shell: bash
  env:
    BUILD_DATE: ${{ inputs.build_date }}
    RELEASE_TAG: ${{ inputs.release_tag }}
  run: |
    set -euo pipefail
    [[ "$BUILD_DATE" =~ ^[0-9]{8}$ ]]
    [[ "$RELEASE_TAG" =~ ^[A-Za-z0-9._-]+$ ]]
```

The job-level `env:` entries that already build `BUILD_BASE_URL` and `ARTIFACT_NAME` from `inputs.build_date` are fine as they are — env assignment is not shell-parsed. Only the `run:` occurrences need moving.

`qc.yml` and `deploy-docs.yml` were checked and have no `${{ inputs.* }}` or `${{ github.event.* }}` interpolation in `run:` blocks.

### LINKED ISSUE #870: PR #862's LazyTransform moves import failure past the guard that was meant to absorb it
Second adversarial review pass on #862.

`transform.py` replaces eager transform imports with a `LazyTransform` wrapper, which is the right shape for offline-safe CLI startup and correctly proxies class attributes:

```python
def __getattr__(self, name: str) -> Any:
    """Expose class attributes without changing the registry API."""
    return getattr(self.transform_class, name)
```

`DATA_INPUTS` still resolves through it, so the #839/#812 data-input freshness detection keeps working in the normal case. The problem is the failure case.

## The degradation no longer degrades

`kgm_freshness_check._declared_data_inputs` is written to survive a broken environment, and says so:

> Import failure degrades to "no declared inputs" so the skill still runs in an environment without kg_microbe importable.

```python
try:
    from kg_microbe.transform import DATA_SOURCES
except Exception:
    return ()
cls = DATA_SOURCES.get(source)
return tuple(getattr(cls, "DATA_INPUTS", ()) or ()) if cls is not None else ()
```

Before this PR, a transform module that could not import made the **registry import** fail, inside the `try`. Now the registry import always succeeds — that is the point of laziness — and the failure is deferred to the `getattr` on the next line, which is **outside** the guard.

`getattr(obj, name, default)` only swallows `AttributeError`. `LazyTransform.__getattr__` raises `ImportError`/`ModuleNotFoundError`, so the default is not used. Verified:

```
getattr(LazyTransform("kg_microbe.transform_utils.nonexistent.Thing"), "DATA_INPUTS", ())
  -> ModuleNotFoundError    (the default is not used)
```

The skill now crashes on one unimportable transform instead of reporting the other seventeen.

## Why this is worth fixing rather than shrugging at

The scenario is not hypothetical — it is the one laziness exists to create. Deferring imports means an environment can now be missing a heavy or optional transform dependency and still start the CLI, which is the stated goal. That is precisely the environment in which the freshness pre-flight is asked to run, and precisely when it will now fail.

## Fix

Either catch it at the use site:

```python
cls = DATA_SOURCES.get(source)
if cls is None:
    return ()
try:
    return tuple(getattr(cls, "DATA_INPUTS", ()) or ())
except ImportError:
    return ()
```

or have `LazyTransform.__getattr__` raise `AttributeError` from an import failure, so `getattr`'s default behaves as every caller expects. The second is the more general fix — anything else in the codebase doing `getattr(cls, ..., default)` against the registry has inherited the same surprise.

Worth grepping for other `getattr(...)`-with-default calls against `DATA_SOURCES` entries before choosing.

### LINKED ISSUE #871: PR #862 advertises Python 3.13 support but locked pyarrow cannot install
Adversarial/CI review of #862.

The PR changes the declared Python range to `>=3.10,<3.14` and adds Python 3.13 to the QC matrix. The first matrix run fails during `poetry install` before tests:

```
pyarrow 15.0.2
Backend subprocess exited when trying to invoke get_requires_for_build_wheel
ModuleNotFoundError: No module named pkg_resources
```

The locked dependency set has no usable Python 3.13 wheel for this path and attempts an incompatible source build. Because the review requirement was to either test newer Python or constrain metadata to versions actually supported, a red 3.13 job means the new metadata overstates support.

## Resolution

Until the dependency chain is upgraded independently:

- constrain Python to `>=3.10,<3.13`;
- keep CI on 3.10, 3.11, and 3.12;
- add a comment linking this issue so 3.13 is intentionally revisited rather than forgotten;
- regenerate and validate `poetry.lock`.

A future dependency-upgrade PR can close this limitation by restoring 3.13 to both metadata and CI once `poetry install` and the full suite pass there.

Found in GitHub Actions run 32448407868, job 96672161350.

realmarcin:
## Summary

This PR addresses the prioritized August 2026 repository review findings:

- removes release credential exposure and limits cleanup to generated artifacts
- makes CLI startup and KGX/BMT initialization offline-safe using the pinned Biolink 4.4.2 model
- removes the advertised no-op `kg holdouts` command (a user-visible CLI removal) and makes query failures exit nonzero
- replaces Pandas-based DuckDB loading with direct, typed DuckDB ingestion, content-sensitive source fingerprints, and atomic rebuilds
- repairs packaging metadata, constrains support to Python 3.10–3.12, and makes unit tests hermetic
- generates all merge variants from `merge.yaml` plus canonical deltas, with distinct artifact and statistics names
- repairs release and documentation workflows around the named Jenkins artifact and checksum
- replaces the 30,461-line ignore list with reviewed policies and regression tests
- refreshes README, CLAUDE.md, `.env.example`, and adds tool-neutral AGENTS.md and runbooks
- starts the staged P3 modularization with a dependency design note and tested MetaTraits I/O, BacDive emission, and ontology resolution extractions
- addresses the full adversarial-review pass: restores operational invariants, removes stale holdouts guidance, revalidates Biolink 4.4.2 assumptions, hardens dispatch input handling, and makes lazy metadata inspection resilient

## Verification

- `KG_MICROBE_TESTS_NO_NETWORK=1 python -m pytest -q`
  - 1,139 passed, 45 skipped, 21 subtests passed
- Ruff and codespell passed
- `poetry check --lock` passed after regenerating the lock
- generated merge configuration check passed
- workflow YAML parsing and `git diff --check` passed

The skipped tests are explicitly marked integration or require large generated data. The four MediaDive local-server cache tests can be enabled with `KG_MICROBE_RUN_INTEGRATION=1`.

## P3 scope

This PR progresses #858, #859, and #860 but does not close them. It establishes module boundaries and moves the first stable, directly tested responsibilities. The remaining orchestration splits and golden full-transform comparisons should land incrementally.

Closes #849
Closes #850
Closes #851
Closes #852
Closes #853
Closes #854
Closes #855
Closes #856
Closes #857
Closes #847
Closes #534
Closes #543
Closes #553
Closes #865
Closes #866
Closes #867
Closes #868
Closes #869
Closes #870
Closes #871

Progresses #858
Progresses #859
Progresses #860


realmarcin:
Adversarial review pass. Four findings filed as issues rather than pushed as changes: **#865, #866, #867, #868**.

Ranked by what I would want addressed before merge:

| # | finding | severity |
|---|---|---|
| #866 | DuckDB cache keys on `mtime + size`, which can serve **stale query results** silently after a `cp -p` / `rsync --times` / backup restore. Also reintroduces the timestamp class #844 closed, without reusing `transform_fingerprint`. | highest — wrong answers, not just slow ones |
| #865 | `CLAUDE.md` 457 → 121 lines; verified that `PREGO_SHAPES`, `PREGO_MIN_CONFIDENCE`, the `OrganismTaxon`/`subclass_of` convention and the "ontology failures abort" section appear **nowhere** in any `.md` in the tree. Several encode a past incident rather than a feature. | high — loses the trap, not just the docs |
| #868 | Biolink pinned to 4.4.2 while three files still say 4.2.2, and the domain/range verdicts derived under 4.2.2 (#832's 471,702 edges, #821's 229,366) are not re-checked. | medium |
| #867 | `kg holdouts` removed — correctly, it was a no-op — but two `SKILL.md` files and a script still invoke it. | medium |

## What I checked and found sound

Recording these so they are not re-reviewed:

- **Nothing merged yesterday is reverted.** Branched from `e463a9f`; all seven of #825/#826/#829/#832/#840/#844/#846 are ancestors. `transform.py` keeps the #844 fingerprint (4 refs), `gold.py` keeps the #832 convention (9 refs), and the table-driven merge drift test from #846 is retained (7 refs).
- **The `.gitignore` rewrite is an improvement, not a risk.** 30,461 → 181 lines, and I tested the real paths against the new rules in a scratch repo: `data/transformed/`, `data/merged/`, `data/raw/gold/` and `data/transformed_pre20260801/` all still ignored. The new `/data/transformed_*/` is strictly better — the old file did **not** cover `transformed_pre20260801`, which is how a `git add -A` swept 17M lines into a commit here recently.
- **The merge generator genuinely closes #847.** All six configs keep their source counts (no_metatraits stays at the 19 that #846 corrected) and now carry distinct artifact names, so a variant run no longer overwrites the standard graph in place.
- **The 45 skips are not a coverage regression.** The `KG_MICROBE_RUN_INTEGRATION` gate is narrow — one file, the four MediaDive local-server tests. The rest are pre-existing data-dependent skips; on a checkout with `data/` built, master reports 1 skip for the same reason.
- **The two new `except Exception` sites are benign** — a DuckDB insert and a gzip probe. Neither wraps ontology adapter use, so the `FatalOntologyError`/`BaseException` design still holds, and `ontology_resolution.py` is a pure extraction that adds no error handling of its own.

realmarcin:
Second review pass, covering the areas the first one did not reach: workflows, packaging, the `transform.py` lazy-import rework, the extractions, and the config generator. Two further findings, both filed: **#869, #870**.

| # | finding | severity |
|---|---|---|
| #870 | `LazyTransform` moves import failure **past** the `try` that was meant to absorb it. `getattr(cls, "DATA_INPUTS", ())` raises `ModuleNotFoundError` rather than returning the default, because `__getattr__` raises `ImportError` and `getattr`'s default only catches `AttributeError`. Verified empirically. The freshness pre-flight now crashes on one unimportable transform instead of reporting the other seventeen — in exactly the environment laziness exists to enable. | medium-high |
| #869 | `release.yml` interpolates `${{ inputs.* }}` into `run:` blocks, **including inside the step that validates them** — so the regex guard is itself the first place the input executes and cannot protect anything, including its own step. Bounded by `workflow_dispatch` requiring write access, so low likelihood; free to fix and it is the pattern GitHub's hardening guide names. | low-medium |

## Checked in this pass and found sound

- **The merge-config generator is drift-proof.** `scripts/generate_merge_configs.py --check` runs in `qc.yml` and there is a dedicated `tests/test_merge_config_generation.py`. My initial concern — that `config/merge_variants.yaml` says files "must be refreshed" manually, which is how #827 happened — does not apply, because the refresh is enforced rather than remembered.
- **The release workflow is otherwise a real improvement.** No secrets on command lines or echoed; inputs are at least *intended* to be validated; and artifact cleanup is narrowed from `rm -f *.tar.gz` — which would delete any tarball in the directory, including one a user put there — to a manifest-driven removal.
- **`LazyTransform` proxies class attributes correctly**, so `DATA_INPUTS` still resolves and the #839/#812 data-input freshness detection keeps working in the normal case. Only the failure path is affected.
- **`qc.yml` and `deploy-docs.yml`** have no `${{ inputs.* }}` or `${{ github.event.* }}` interpolation inside `run:` blocks.
- **The two new `except Exception` sites** (a DuckDB insert, a gzip probe) do not wrap ontology adapter use, so the `FatalOntologyError`/`BaseException` contract is intact; `ontology_resolution.py` is a pure extraction adding no error handling of its own.

Running total across both passes: **#865, #866, #867, #868, #869, #870**. If I had to pick two to block on, they are #866 (stale query results served silently) and #870 (the pre-flight crashing in the environment this PR makes possible).

realmarcin:
Adversarial review follow-up is in 9a3dbd9:

- #865: restored incident-bearing operational invariants and new-transform guidance in CLAUDE.md
- #866: replaced mtime freshness with bounded content fingerprints and a same-size/same-mtime regression
- #867: removed all remaining agent/report recommendations for the deleted `kg holdouts` CLI and added the user-visible removal note
- #868: pinned the model-review skill to local 4.4.2 files and documented the exact taxon/located_in/occurs_in revalidation
- #869: moved dispatch inputs out of shell program text and added a workflow regression test
- #870: preserved getattr defaults for unavailable lazy transforms while retaining execution-time import failures
- #871: constrained Python to 3.10–3.12, regenerated the lock, and updated CI/docs

Local verification: 1,139 passed, 45 skipped, 21 subtests passed; Ruff, codespell, Poetry lock validation, merge-config generation, and diff checks passed.

realmarcin:
CI follow-up: clean installs exposed #872 in locked BMT 1.4.8. Commit 5545dbc loads local predicate-map paths/file objects/mappings without entering BMT 1.4.8s unconditional `requests.get` path. The regression passes with a legacy stand-in and the actual BMT 1.4.8 environment; the replacement matrix is running.

Issue #834, opened 2026-08-20
Record the graph-wide OrganismTaxon / subclass_of deviation from Biolink's letter


realmarcin:
Raised while reviewing #832. Documentation, not a bug — but it should be written down somewhere other than a comment in `gold.py`.

`biolink:subclass_of` declares `domain: ontology class` and `range: ontology class`. `biolink:OrganismTaxon` is **not** an ontology class — its ancestry is `OrganismTaxon → NamedThing → Entity`.

The graph nevertheless uses `subclass_of` between `OrganismTaxon` nodes throughout:

| source | edges |
|---|---|
| NCBITaxon's own hierarchy | 925,219 |
| bacdive strains → taxon | 251,916 |
| gold organisms → taxon (as of #832) | 471,702 |
| metatraits | 186 |

That is ~1.65M edges, including the entire taxonomic backbone. Nothing mechanically rejects them — KGXVal does not enforce domain/range on this slot — so the deviation has been invisible.

## Why we are keeping it

Consistency beats conformance here. `biolink:in_taxon` is the domain-valid alternative and GOLD used to use it, but a source that is uniquely valid is *unreachable*: a query walking `subclass_of` down from a species found bacdive's strains and silently missed all 531k GOLD organisms. #832 resolved that by conforming GOLD to the house convention rather than the spec.

The alternative — moving all four sources to `in_taxon` and retyping 251k bacdive strains plus 471k GOLD organisms to `IndividualOrganism` — would also have to answer what NCBITaxon's own hierarchy becomes, since `in_taxon` between two taxa is meaningless. There is no clean version of that, which is a decent sign the Biolink reading is the wrong tool for a taxonomy.

## What to do

Write it into `CLAUDE.md` next to the other modelling conventions, so the next person to reach for `in_taxon` finds the reasoning instead of rediscovering the split. Worth raising upstream with Biolink separately — `OrganismTaxon` not being an `OntologyClass` while every taxonomy in practice is a class hierarchy looks like a model gap rather than our error.

realmarcin:
Documented in CLAUDE.md under Naming Conventions, via #832. Left the upstream Biolink question — `OrganismTaxon` not being an `OntologyClass` while every taxonomy in practice is a class hierarchy — unfiled here; it is worth raising with the Biolink model separately rather than tracking as a kg-microbe issue.

PR #893, opened 2026-08-30
fix(bacdive): resolve contradictory parents on shared deposit nodes, and link deposits to their records (#892, #894, #897, #898, #902, #903)

### LINKED ISSUE #892: Strain nodes with two taxonomically incompatible NCBITaxon parents (524 in merged 20260815; 2,660 in 20240826)
Found while building organism-level clusters for a train/test split, where a strain and its parent taxon must not straddle the split. Union-find over `biolink:subclass_of` edges produced clusters that merged unrelated organisms, which traced back to strain nodes carrying more than one NCBITaxon parent.

## Current build — `merged/20260815`

```
kgmicrobe.strain: --subclass_of--> NCBITaxon:   251,974 edges over 251,441 strains
  strains with >1 parent:  524  (0.208%)     [515 with two, 9 with three]
  of those, parents in DIFFERENT genera:  368  (70%)
```

Examples with labels:

| strain node | parent 1 | parent 2 |
|---|---|---|
| `kgmicrobe.strain:Agre-8` | *Companilactobacillus farciminis* | *Nocardioides* sp. |
| `kgmicrobe.strain:AJ-1378` | *Prescottella equi* | *Streptosporangium amethystogenes* subsp. *amethystogenes* |
| `kgmicrobe.strain:AS-1` | *Methylophilus methylotrophus* DSM 46235 | *Alteromonas simiduii* |
| `kgmicrobe.strain:ATCC-13722` | *Pseudoclavibacter* sp. | *Gulosibacter faecalis* |

These are not species/subspecies pairs — 368 of 524 are different genera entirely.

## The mechanism looks like non-unique strain designations used as node ids

Every current case is a **bare strain designation** (`AS-1`, `Agre-8`, `ATCC-13722`), not a record-scoped id. Designations like "AS-1" are reused by unrelated organisms across sources, so keying a strain node on the designation collides two real strains into one node, and both parents attach.

`ATCC-13722` is a good illustration: an ATCC catalogue number should be unique, so two different parents suggests two source records disagreeing rather than two genuine strains.

## It used to be worse, and a different bug

For comparison, in `merged/20240826` (the build behind a model we are auditing):

```
strain: --subclass_of--> NCBITaxon:   162,286 edges over 159,606 strains
  strains with >1 parent:  2,660  (1.67%)
    bacdive_NNNN record ids   : 2,230
    bare strain designations  :   430
  parents in different genera : 88% of the resolved-pair cases
```

The 2,230 **record-id** cases are the notable difference. A `bacdive_NNNNN` id is record-scoped and should map to exactly one organism, so those could not be designation collisions. Examples from that build:

| strain node | parent 1 | parent 2 |
|---|---|---|
| `strain:bacdive_164638` | *Arthrobacter crusticola* | *Aspergillus flavus* var. *columnaris* (a fungus) |
| `strain:bacdive_163735` | sulfur-oxidizing endosymbiont of *Inanidrilus leukodermatus* | *Chloracidobacterium thermophilum* |
| `strain:bacdive_162123` | *Hyunsoonleella pacifica* | *Paraburkholderia xenovorans* |

**That class appears fixed**: `merged/20260815` has **zero** multi-parent `bacdive_` record ids, out of 99,386 record-id strains. Whatever changed between the two builds resolved it, and this issue is not asking for that to be revisited — it is recorded so the remaining 524 are not mistaken for the same thing.

In 20240826 a further 1,652 strains had one resolved `biolink:OrganismTaxon` parent plus one parent typed `biolink:NamedThing` with no label — a dangling NCBITaxon id. Also worth a look if those ids are still being emitted.

## Why it matters downstream

Anything that walks `strain → NCBITaxon` to group or resolve organisms inherits the collision. In our case union-find chained them into one 3,312-node component in the 20240826 build (1.8% of nodes). We checked the impact on our own analysis and it was small, but the failure mode is silent: nothing errors, and a consumer that takes the first parent gets a result decided by file order.

## Suggested

1. Scope strain node ids by source record where one exists, rather than by designation, so `AS-1` from two sources cannot merge.
2. Failing that, emit a warning when a strain node acquires parents in different genera — 70% of the current cases would be caught by that check alone.
3. If a strain genuinely has conflicting taxonomy across sources, keeping both edges is defensible, but consumers need a way to tell that apart from a collision.

Happy to share the exact node lists for either build.

realmarcin:
Closes #892

## The bug

A `kgmicrobe.strain:<code>` node minted from a culture-collection deposit number identifies a **physical deposit, not a BacDive record**, so several records can cite the same one — and they do not always agree on the taxon. The transform wrote one `subclass_of` edge per claiming record, so the deposit node ended up with several NCBITaxon parents and nothing to tell them apart. A consumer that takes "the" parent gets whichever one file order happened to put first; the reporter's union-find over `subclass_of` chained unrelated organisms into one component.

Worked example from the current dump — BacDive 3081 and 7476 both cite `ATCC 13722`:

| record | taxon | culture collection no. |
|---|---|---|
| 3081 | NCBITaxon:272240 *Gulosibacter faecalis* | DSM 20154, NCDO 1718, **ATCC 13722**, CIP 108568, IAM 15030, IFO 15706, TISTR 1514 |
| 7476 | NCBITaxon:1915061 *Pseudoclavibacter* sp. | DSM 17446, IAM 15030, JCM 12866, NBRC 15706, TISTR 1514, **ATCC 13722**, … |

## The fix

Deposit parent claims are buffered while records stream and resolved after the loop. A parent is asserted only when every record citing the deposit agrees on it. Where they disagree, no edge is written and the claim goes to `bacdive_strain_deposit_conflicts.tsv` in the transform's output directory (`strain_id`, `parent_count`, `parents`, `bacdive_ids`), so the suppressed taxonomy stays recoverable and the conflict is visible instead of silent. A summary count is logged.

Record-scoped `kgmicrobe.strain:bacdive_NNN` nodes are untouched — they already have exactly one parent each (99,386 of them, zero multi-parent).

## Correcting the reported mechanism

The issue hypothesises non-unique bare strain designations used as node ids. Measured against the dump, that is only ~6% of the cases:

| colliding token looks like | count |
|---|---|
| a recognised culture-collection acronym (ATCC, DSM, CIP, …) | 494 |
| uppercase acronym shape but a rare prefix | 26 |
| not a collection code at all (`Agre-8`, `Chol-1`, `Sao-Paulo`) | 5 |

So restricting id minting to recognised collections — the obvious first fix, and the one LPSN's `_CULTURE_CODE` allowlist already applies on its side — would not have fixed this. The dominant cause is BacDive citing the same deposit from records with conflicting taxonomy: 222 of the two-record cases share more than one deposit number (so they look like duplicate records for one physical strain), 306 share exactly one.

## Verification

- New unit tests in `tests/test_bacdive_deposit_parents.py` (5 tests) covering single claimant, agreeing claimants, disagreeing claimants, per-deposit isolation, and the report rows.
- Offline replay of the resolution over the real 99,392-record `bacdive_strains.json`: **515 of 150,918 deposits (0.34%)** have disagreeing claimants, so **1,039 of 151,442** deposit edges are suppressed, and every emitted deposit has exactly one parent. The replay cannot cover the ~818 records that reach their taxon through the OAK name fallback, which is why the live output shows 525 rather than 515.
- `poetry run tox`: format, lint, codespell, docstr-coverage all OK. `tests/test_ontologies_stubs.py::test_every_referenced_curie_has_stub_node` fails identically on `master` (stale local `ontologies_stubs` output, `PO:0009005`) — pre-existing, unrelated.

The transform itself has **not** been rerun; that is on hold pending the updated MIM SSSOM. This change makes `bacdive` stale and will be picked up by that rerun.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -91,8 +91,11 @@ always agree on the taxon. Such a node gets the one claimed parent that every
 claimant entails — the claim itself when they agree, the shared ancestor when
 they differ only in depth along one lineage. Never a computed common ancestor
 nobody claimed, and never one of two disjoint claims: those get no parent at all
-and go to the source's deposit-conflict report, because picking one would make
-the answer depend on file order. Every record that cites a deposit links to it
+and go to the source's deposit-claim report, because picking one would make
+the answer depend on file order. That report carries every deposit whose
+claimants disagreed, whichever way it went, so a coarsened taxonomy and a
+suppressed one are both visible and neither is confused with an ontology
+lookup that failed. Every record that cites a deposit links to it
 with `biolink:close_match`, so the deposit's provenance is in the graph and its
 taxonomy stays reachable through the record even when no parent is asserted.
 See issues #892 and #894.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,6 +85,13 @@ traversable; it is not a claim that the isolate is an ontology class. Biolink
 validators will report it; do not silently change the convention. See issue
 #834 and [the 4.4.2 revalidation](docs/BIOLINK_4_4_2_REVALIDATION.md).
 
+A `kgmicrobe.strain:<code>` node minted from a culture-collection deposit number
+is shared: several source records can cite the same deposit. Such a node gets a
+`subclass_of` parent only when every record citing it agrees on the taxon. When
+they disagree, no parent is asserted and the claim is written to the source's
+deposit-conflict report instead — picking one would make the answer depend on
+file order. See issue #892.
+
 ## Operational traps
 
 - PREGO defaults to `PREGO_SHAPES=habitat` and `PREGO_MIN_CONFIDENCE=0`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,11 +86,13 @@ validators will report it; do not silently change the convention. See issue
 #834 and [the 4.4.2 revalidation](docs/BIOLINK_4_4_2_REVALIDATION.md).
 
 A `kgmicrobe.strain:<code>` node minted from a culture-collection deposit number
-is shared: several source records can cite the same deposit. Such a node gets a
-`subclass_of` parent only when every record citing it agrees on the taxon. When
-they disagree, no parent is asserted and the claim is written to the source's
-deposit-conflict report instead — picking one would make the answer depend on
-file order. See issue #892.
+is shared: several source records can cite the same deposit, and they do not
+always agree on the taxon. Such a node gets the one claimed parent that every
+claimant entails — the claim itself when they agree, the shared ancestor when
+they differ only in depth along one lineage. Never a computed common ancestor
+nobody claimed, and never one of two disjoint claims: those get no parent at all
+and go to the source's deposit-conflict report, because picking one would make
+the answer depend on file order. See issue #892.
 
 ## Operational traps
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,7 +92,10 @@ claimant entails — the claim itself when they agree, the shared ancestor when
 they differ only in depth along one lineage. Never a computed common ancestor
 nobody claimed, and never one of two disjoint claims: those get no parent at all
 and go to the source's deposit-conflict report, because picking one would make
-the answer depend on file order. See issue #892.
+the answer depend on file order. Every record that cites a deposit links to it
+with `biolink:close_match`, so the deposit's provenance is in the graph and its
+taxonomy stays reachable through the record even when no parent is asserted.
+See issues #892 and #894.
 
 ## Operational traps
 
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -290,6 +290,7 @@ class BacDiveTransform(Transform):
         self.ncbitaxon_synonyms: Dict[str, frozenset] = {}  # {ncbitaxon_id: frozenset(lowercased synonyms)}
         self.ncbitaxon_fallback_cache: Dict[str, Optional[str]] = {}  # Cache for fallback lookups
         self._ncbitaxon_rank_cache: Dict[str, Optional[str]] = {}  # {ncbitaxon_id: rank_name}
+        self._ncbitaxon_ancestor_cache: Dict[str, frozenset] = {}  # {ncbitaxon_id: proper is_a ancestors}
         self._load_ncbitaxon_labels()
 
         # Load CHEBI categories from ontologies transform output (for category alignment).
@@ -917,6 +918,30 @@ class BacDiveTransform(Transform):
             if n >= self._MIN_SHARED_PREFIX and token[: self._MIN_SHARED_PREFIX] == name[: self._MIN_SHARED_PREFIX]:
                 return True
         return False
+
+    def _ncbitaxon_ancestors(self, ncbitaxon_id: str) -> frozenset:
+        """
+        Return the proper ``is_a`` ancestors of an NCBITaxon CURIE, cached.
+
+        Used to tell a real taxonomic disagreement between two records citing one
+        culture-collection deposit from two claims at different depths of the same
+        lineage (#892). Only a few hundred deposits ever reach this, so the cost is
+        a handful of adapter calls per run.
+        """
+        cached = self._ncbitaxon_ancestor_cache.get(ncbitaxon_id)
+        if cached is not None:
+            return cached
+        try:
+            ancestors = frozenset(self.ncbi_impl.ancestors([ncbitaxon_id], predicates=[RDFS_SUBCLASS_OF]))
+            ancestors -= {ncbitaxon_id}
+        except OntologyDbUnavailableError:
+            # An unusable NCBITaxon DB is not an empty ancestry — surface it rather
+            # than silently declaring every pair of claims disjoint.
+            raise
+        except Exception:
+            ancestors = frozenset()
+        self._ncbitaxon_ancestor_cache[ncbitaxon_id] = ancestors
+        return ancestors
 
     def _parse_genus_from_scientific_name(self, scientific_name: str) -> Optional[str]:
         """
@@ -3367,8 +3392,11 @@ class BacDiveTransform(Transform):
                     progress.update()
 
                 # Resolve the buffered culture-collection deposit claims: assert a
-                # parent only where every record citing the deposit agrees (#892).
-                resolved_deposits, deposit_parent_conflicts = resolve_deposit_parents(deposit_parent_claims)
+                # parent only where every record citing the deposit supports it (#892).
+                resolved_deposits, deposit_parent_conflicts = resolve_deposit_parents(
+                    deposit_parent_claims, ancestors_of=self._ncbitaxon_ancestors
+                )
+                deposits_collapsed = sum(1 for curie, _ in resolved_deposits if len(deposit_parent_claims[curie]) > 1)
                 for strain_curie, deposit_parent_id in resolved_deposits:
                     knowledge_level, agent_type = self._add_edge_metadata(
                         SUBCLASS_PREDICATE, RDFS_SUBCLASS_OF, deposit_parent_id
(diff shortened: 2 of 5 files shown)


realmarcin:
Pushed `0f738eeb` after a question about whether the surviving parent edge is always correct. It exposed a real defect in the first commit.

**76 of the 524 multi-parent deposits are not contradictions at all.** The claiming records differ only in depth along one lineage:

| deposit | claim A | claim B |
|---|---|---|
| `ATCC-11426` | NCBITaxon:41276 *Brevundimonas vesicularis* | NCBITaxon:1349759 *B. vesicularis* NBRC 12165 |
| `ATCC-14158` | NCBITaxon:58340 *Streptomyces lavendulae* subsp. *lavendulae* | NCBITaxon:1303691 …JCM 4055 |
| `ATCC-25544` | NCBITaxon:1833 *Rhodococcus erythropolis* | NCBITaxon:1219014 …NBRC 15567 |

The transform prefers a strain-level taxid over a species-level one when a record offers both, so two records describing the same organism land at different depths and the first commit read that as a conflict and suppressed a correct edge.

A deposit now takes **the one claimed taxon that every claimant entails**: the claim itself when they agree, the shared ancestor when they differ only in depth. Genuinely disjoint claims are still suppressed.

The asserted taxon must be one a record actually claimed. Falling back to a *computed* common ancestor would always succeed — every pair of bacteria shares a phylum — and would put `Bacteria` on the node while reporting the conflict resolved. `tests/test_bacdive_deposit_parents.py::test_no_unclaimed_taxon_is_ever_asserted` pins that.

Ancestry comes from the NCBITaxon adapter already resolved for the run, cached per taxon. Replayed over the real dump:

```
deposits claimed:                 150,918
emitted, single claimant:         150,403
emitted, collapsed to shared anc:      78
suppressed, disjoint lineages:        437
534 ancestor lookups, 0.7s
```

(78/437 rather than 76/448 because the replay uses the adapter's ancestry rather than the extract's edges, and 5 conflicting deposits have a parent the extract does not carry — see #895.)

Tests are 5 → 10. `tox` unchanged: everything green except the pre-existing `test_ontologies_stubs` failure that also fails on master.

## What this rule does and does not guarantee

Worth stating plainly, since the PR title could be read as a stronger claim than it is. It guarantees that **no BacDive-minted deposit node carries two contradictory parents**, so no consumer's answer depends on file order. It does **not** guarantee the surviving edge is right:

- **Agreement is not correctness.** 150,403 of 150,481 emitted edges come from a deposit with a single claimant, passed through verbatim. If BacDive's assignment is wrong there, or the OAK name-search fallback grounded to the wrong taxon, the edge is wrong and nothing here notices. The rule detects disagreement, not error.
- **Claimants can agree and both be wrong.** 222 of the two-record cases share more than one deposit number, i.e. they look like duplicate BacDive records for one physical strain. Two duplicates carrying the same wrong taxid agree.
- **Only BacDive's claims are visible.** A designation reused by a source outside BacDive is not seen — that is #896.
- **Collapsing prefers the coarser claim.** For those 78, strain-level resolution is dropped from the deposit node. It stays on the claiming record's own `kgmicrobe.strain:bacdive_NNN` node, which is untouched.

realmarcin:
Folded #894 in as `eedd9a4e`, so this PR now closes both.

Each BacDive record emits `kgmicrobe.strain:bacdive_<id> --biolink:close_match--> kgmicrobe.strain:<deposit>` for every deposit it cites — the half of the link LPSN was already emitting on its side (`lpsn:<record> close_match kgmicrobe.strain:<code>`, whose docstring says BacDive emits the same CURIEs so the merge reconciles both).

```
records citing at least one deposit:  66,774
distinct deposit nodes:              152,012
close_match edges added (deduped):   152,890   (+3.6% on bacdive's 4.24M)
```

**Uniform rather than targeted.** #894 left that open. Emitting only for the deposits whose claimants disagree would cost ~1,000 edges instead of 152,890, but it leaves a rule that reads as arbitrary to anyone who has not read the issue, and provenance is worth having on every deposit, not only the broken ones.

**This is what makes the #892 suppression acceptable.** The 437 deposits that get no parent now reach their taxonomy in two hops through the claiming record, whose own parent is unaffected and unambiguous (99,386 record nodes, one parent each, zero multi-parent). Before this commit they had no path to a taxon at all.

Two new tests, 10 → 12:

- the deposit block emits `close_match` with the record as subject and the deposit as object, not the reverse — parsed out of the real `writerow` argument list, since `run()` needs the NCBITaxon adapter and cannot be exercised end to end;
- BacDive's normalisation and LPSN's `_extract_strain_curies` produce the same CURIE for the same deposit string. The two halves meet at merge only while that holds, and it is currently two independent copies of one rule.

realmarcin:
Adversarial review pass done. Three findings, all filed; two addressed here as `ac755385`.

**#898 — collapsed deposits were invisible.** The report listed only the deposits that lost their edge. For the 78 collapsed to a shared ancestor, the discarded claim existed in no artifact at all: nothing recorded that `ATCC 11426` had ever been claimed as anything more specific than the species now on it, or that its taxonomy had been coarsened. The PR described the report as "the record that the claim existed and what it was", which was true for 437 of the 515. Every contested deposit is now reported, with a `resolution` column and the parent actually asserted:

```
strain_id                      parent_count  parents                              bacdive_ids   resolution  asserted_parent
kgmicrobe.strain:AJ-1378       2             NCBITaxon:43767|NCBITaxon:698417     10848|16611   suppressed
kgmicrobe.strain:ATCC-11426    2             NCBITaxon:1349759|NCBITaxon:41276    2301|2302     collapsed   NCBITaxon:41276
```

**#897 — a failed ancestry lookup was swallowed.** `_ncbitaxon_ancestors` caught bare `Exception` and returned an empty frozenset, which makes two claims on one lineage look disjoint and suppresses a correct edge. Suppressing is the right precaution, but reporting it as a plain conflict blames BacDive for the ontology's failure, and a partially degraded adapter would have looked exactly like a data problem in the source. Failed taxa are now recorded, those suppressions are marked `suppressed_ancestry_unavailable`, and the run logs how many lookups failed.

**#899 — not addressed here.** The `close_match` edges added in `eedd9a4e` are emitted for every token in `culture collection no.`, including the 4,690 (3.1%) whose acronym is not among the 107 collections used 50+ times. 49 of those are cited by more than one record, so two records citing `Agre 8` — *Lactobacillus farciminis* and *Nocardioides* sp. — are now explicitly two hops apart over `skos:closeMatch`. Before this PR that collision was an artifact of a shared node; now it is an assertion. It is a modeling decision with real data loss on any option (dropping edges, fragmenting 4,690 nodes, or minting fewer), and it wants the acronym list shared with LPSN rather than written a second time — so it belongs in its own change, not tacked onto this one.

Replayed over the real dump after the fix:

```
deposits claimed:            150,918
edges emitted:               150,481
contested (in the report):       515
    suppressed                   437
    collapsed                     78
ancestry lookups that failed:      0
```

Tests 12 → 16. `tox` green apart from the pre-existing `test_ontologies_stubs` failure that also fails on master.

Also checked and clear: nothing globs `data/transformed/<source>/*.tsv` (merge.yaml names both files explicitly, `validate_knowledge_sources.py` looks for `edges.tsv` by name, `microbial_trait_mappings` globs `mappings/`), so the report file cannot be mistaken for a KGX input.

realmarcin:
Second adversarial pass. Two findings, both filed and fixed in `bbeee456`.

**#902 — the report's own documentation described the previous behaviour.** `ac755385` widened the file to carry `collapsed` rows as well as `suppressed` ones, but `constants.py` and the comment above the write still said its rows "get no subclass_of edge". False for the 78 collapsed deposits, which do get one. Anyone trusting either comment would filter the graph on a false premise. The filename carried the same claim, so it is renamed `bacdive_strain_deposit_conflicts.tsv` → **`bacdive_strain_deposit_claims.tsv`** — free right now because nothing consumes it, not free once something does.

**#903 — the report was written non-atomically.** `nodes.tsv` and `edges.tsv` are closed and complete before it is written, so a crash in the gap left outputs that looked finished beside a report that silently understated what had been suppressed. Three states were conflated — absent, header-only, and truncated all read as "nothing was contested" — which is the exact silent failure the file exists to prevent, and what CLAUDE.md's *cache existence must imply completeness* rule covers. Now written through `atomic_write`. `bacdive_media_links.txt` sits immediately above with the same shape and predates this PR; converted in the same pass rather than leaving one of two done.

Tests 16 → 18. `tox` green apart from the pre-existing `test_ontologies_stubs` failure that also fails on master.

Checked and clear this pass:

- **Merge does not collapse on `close_match`.** Worth confirming before shipping 152,890 of them — if the merge treated it as identity it would have fused 66,774 record nodes into deposit nodes. `merge.yaml` is a plain KGX union with no such handling, and nothing in `merge_utils/` touches `close_match`.
- **`close match` has no domain or range in Biolink 4.4.2** (`symmetric: true`, `exact_mappings: [skos:closeMatch, SEMMEDDB:same_as]`), so unlike the `subclass_of` house convention in #834 these edges raise no domain/range violation. The `SEMMEDDB:same_as` mapping is why #899 matters, which is filed.
- **No record lacks a `BacDive-ID`**, so the `key` fallback to list index never fires and the report's `bacdive_ids` column cannot silently carry an index posing as an accession.
- **`run()` has no early return**, so the report is reached on every non-exceptional path.

Issue #897, opened 2026-08-30
BacDive ancestry lookup failures are swallowed, silently suppressing deposit parent edges


realmarcin:
Found in the adversarial review of #893.

`BacDiveTransform._ncbitaxon_ancestors` re-raises `OntologyDbUnavailableError` but catches bare `Exception` and returns an empty frozenset:

```python
except OntologyDbUnavailableError:
    raise
except Exception:
    ancestors = frozenset()
```

The precedent is `_get_ncbitaxon_rank`, which does the same, but the consequence is not the same. There an empty result costs a validation check. Here an empty ancestry makes two claims that are genuinely on one lineage look disjoint, so `resolve_deposit_parents` suppresses a `subclass_of` edge that should have been emitted and files it as a conflict. Nothing counts it and nothing says it happened.

That is the #892 failure mode inverted: instead of a wrong answer decided by file order, a correct answer quietly withheld because the ontology hiccupped. A partially degraded NCBITaxon adapter would inflate the conflict count and look like a data problem in BacDive.

Proposed: count the taxa whose ancestry lookup failed, log the count next to the collapsed/disjoint summary, and mark the affected rows in `bacdive_strain_deposit_conflicts.tsv` so a suppression caused by a lookup failure is distinguishable from one caused by real disagreement.

Issue #894, opened 2026-08-30
Culture-collection deposit nodes have no edge back to the records that assert them


realmarcin:
Found in the review of #893 (fixes #892).

`kgmicrobe.strain:<deposit>` nodes minted by the BacDive transform from `External links → culture collection no.` carry exactly one outgoing edge: `subclass_of` to the record's NCBITaxon. Nothing links the deposit node to the BacDive record that asserted it.

```
$ grep -P "kgmicrobe\.strain:bacdive_3081\t" data/transformed/bacdive/edges.tsv \
    | awk -F'\t' '$3 ~ /^kgmicrobe\.strain:/'
(no output)
```

LPSN already emits the equivalent link on its side — `lpsn:<record> --biolink:close_match--> kgmicrobe.strain:<code>` (`_make_close_match_edge`, `kg_microbe/transform_utils/lpsn/lpsn.py:786`), with a docstring that says BacDive emits the same CURIEs so the merge reconciles both sides. BacDive does not hold up its half.

Two consequences:

1. **No provenance.** A deposit node's claimants are invisible in the graph. That a deposit is cited by two records — the whole mechanism behind #892 — cannot be seen from the KG, only from the transform's conflict report.
2. **After #893, 515 deposits are taxonomically orphaned.** #893 suppresses the `subclass_of` edge where claimants disagree, which is right (picking one makes the answer depend on file order), but with no record link there is no remaining path from those nodes to any taxon. Their only other edges are incoming `location_of` from isolation sources.

Proposed: mirror LPSN and emit `kgmicrobe.strain:bacdive_<id> --biolink:close_match--> kgmicrobe.strain:<deposit>` (`skos:closeMatch`) for every deposit BacDive cites. Taxonomy for a conflicted deposit becomes reachable in two hops through the record node, which already has exactly one parent each (99,386 record nodes, zero multi-parent), and a consumer can see how many records claim a deposit.

Cost: about 153k edges on top of bacdive's current 4.24M (+3.6%), one per distinct (record, deposit) pair.

Open question worth deciding before implementing: emit uniformly, or only for deposits whose claimants disagree (~1,000 edges)? Uniform is easier to explain and gives every deposit provenance; targeted is cheaper but leaves a rule that reads oddly to anyone who did not read this issue.

Issue #895, opened 2026-08-30
113 NCBITaxon parents cited by BacDive strain edges are absent from the pinned NCBITaxon extract


realmarcin:
Raised in #892 and confirmed in the review of #893.

#892 reports that in `merged/20240826`, 1,652 strains had one resolved `biolink:OrganismTaxon` parent plus one parent typed `biolink:NamedThing` with no label — a dangling NCBITaxon id — and asks whether those ids are still being emitted.

They are, at a much smaller scale. Against the current local transform outputs:

```
NCBITaxon nodes in the ontologies extract:            925,220
distinct NCBITaxon parents on strain subclass_of edges: 24,708
of those, absent from the extract:                        113
```

e.g. `NCBITaxon:1037355`, `NCBITaxon:1073081`, `NCBITaxon:1076620`, `NCBITaxon:110539`, `NCBITaxon:1143711`.

These are taxids BacDive records carry that the pinned NCBITaxon release does not contain — retired or merged ids, plus whatever the microbial scope trim drops. Nothing in the transform notices: the edge is written, the node never is, and KGX types the object `biolink:NamedThing` with no label. That is exactly the shape the reporter saw, down from 1,652 but not to zero.

Worth deciding:

- Are these merged taxids that should be forwarded to their replacement, or genuinely out of scope?
- If out of scope, should the edge be dropped rather than left pointing at a node that will never exist?
- Either way the transform should count them and say so, the way #893 now does for conflicting deposit parents.

Reproduce:

```bash
awk -F'\t' 'NR>1 && $1 ~ /^NCBITaxon:/ {print $1}' data/transformed/ontologies/ncbitaxon_nodes.tsv | sort -u > /tmp/ncbi_nodes.txt
awk -F'\t' '$1 ~ /^kgmicrobe\.strain:/ && $3 ~ /^NCBITaxon:/ && $2=="biolink:subclass_of" {print $3}' \
  data/transformed/bacdive/edges.tsv data/transformed/metatraits/edges.tsv | sort -u > /tmp/parents.txt
comm -23 /tmp/parents.txt /tmp/ncbi_nodes.txt | wc -l
```

realmarcin:
Correcting my own count before fixing it. **113 is the right number for "absent from the pinned NCBITaxon extract", but it is not the number that becomes an untyped node.**

Measured against the current transform outputs:

```
strain subclass_of parents absent from the ontologies extract:   113
[...]

PR #906, opened 2026-09-02
fix(bacdive): withhold close_match on deposits proven to be shared designations (#899, #907, #908)

### LINKED ISSUE #899: Deposit nodes minted from in-house designations assert false close_match equivalences
Found in the adversarial review of #893, and it sharpens the first suggestion in #892.

BacDive's `External links → culture collection no.` is free text. Most entries are real deposits (`DSM 20154`, `ATCC 13722`), but some are in-house strain designations with no collection behind them. The transform mints a shared `kgmicrobe.strain:<token>` node for every entry alike.

```
distinct deposit nodes:                    152,012
acronym not among the 107 used >=50 times:   4,690  (3.1%)
  of those, cited by more than one record:      49
e.g. AAB-1, AArc-1, ABE-101, ACA-DC-205, Agre-8, BC-3, CAT-2, Sao-Paulo
```

Before #893 a collision here produced two `subclass_of` edges on one node — bad, but an artifact. #893 adds `kgmicrobe.strain:bacdive_<id> --biolink:close_match--> kgmicrobe.strain:<token>` for every citation, which turns the artifact into an **assertion**: two records citing `Agre 8` (BacDive 100025 *Lactobacillus farciminis*, 104641 *Nocardioides* sp.) are now explicitly two hops apart over `skos:closeMatch`, and a consumer clustering on close_match will merge them.

The `subclass_of` conflict on those nodes is handled — #893 suppresses it — but the close_match edges are emitted unconditionally, so the false equivalence survives the fix that was supposed to remove it.

Worth noting that #893 measured this class as only ~6% of the multi-parent cases, which is why an acronym allowlist was not the fix there. It is a much better fit for *this* problem: LPSN already restricts minting with `_CULTURE_CODE` (`kg_microbe/transform_utils/lpsn/lpsn.py:86`) and BacDive does not.

Options, roughly in order of how much they cost:

1. Emit the `close_match` edge only for tokens with a recognised collection acronym, keeping the node for tokens without one. Cheapest, and removes the assertion without losing a node.
2. Scope unrecognised tokens per record (`kgmicrobe.strain:bacdive_<id>_<token>`), so they can never collide. Loses nothing, but fragments 4,690 nodes that currently have no cross-source value anyway.
3. Do not mint nodes for unrecognised tokens at all.

Whichever is chosen, the acronym list should be shared with LPSN rather than written a second time — the two transforms already keep two independent copies of the deposit-string normalisation rule (pinned by `tests/test_bacdive_deposit_parents.py::test_bacdive_and_lpsn_land_on_the_same_deposit_curie`).

realmarcin:
Closes #899

## The problem

#894 added `kgmicrobe.strain:bacdive_<id> --biolink:close_match--> kgmicrobe.strain:<deposit>` for every deposit a record cites. In Biolink 4.4.2 `close match` is `symmetric: true` with `exact_mappings: [skos:closeMatch, SEMMEDDB:same_as]`, so where two unrelated organisms happen to bear the same in-house designation, the graph now asserts a path between them over a predicate consumers read as identity.

| deposit | record | taxon |
|---|---|---|
| `AS-1` | BacDive 7239 | NCBITaxon:1122236 *Methylophilus methylotrophus* |
| | BacDive 134230 | NCBITaxon:398036 *Alteromonas simiduii* |
| `Agre-8` | BacDive 100025 | NCBITaxon:1612 *Lactobacillus farciminis* |
| | BacDive 104641 | NCBITaxon:35761 *Nocardioides* sp. |

Before #894 that collision was an artifact of a shared node. #894 turned it into an assertion.

## Why the allowlist in #899 is the wrong tool

#899 proposed restricting minting to recognised culture-collection acronyms, mirroring LPSN's `_CULTURE_CODE`. I built one — derived from the dump by acronym frequency, ≥50 deposits, 107 acronyms covering 96.9% of tokens — and then deleted it, because it admits exactly the tokens the issue is about:

| acronym | in a frequency allowlist | deposits |
|---|---|---|
| `AS` | **yes** | 150 |
| `AJ` | **yes** | 62 |
| `CN` | **yes** | 54 |
| `ATCC` | yes | 7,720 |

This is consistent with the measurement in #892: **494 of the 525 known collisions already carry an acronym any such list would accept.** Frequency measures how often a prefix appears across records, not whether it scopes an identifier — and a prefix reused by many labs is frequent *because* it is unscoped.

A curated registry list would be better than a derived one, but it still could not separate `AS 1` at Academia Sinica from `AS 1` in someone's freezer, and it would need maintaining against a registry we do not vendor.

## What this does instead

Use evidence we already compute. #892's resolution says, per deposit, whether its claiming records are taxonomically disjoint:

- **kept a parent** (agreed outright, or collapsed to the ancestor every claimant entails) → one identifier its claimants agree about → link it.
- **kept no parent** (disjoint claims) → a designation unrelated organisms share → link nothing; the citation goes to `bacdive_strain_deposit_claims.tsv`, which is where #892 already puts what it declines to assert.
- **ancestry unreadable** (#897) → same treatment. Not asserting beats asserting what was never checked.

Citations are buffered and emitted after resolution. Emitting inline is what shipped the problem: the edge was written before the deposit's other claimants had been read.

## Effect, replayed over the real 99,392-record dump

```
deposit citations total:              152,900
  close_match emitted:                152,009
  withheld (unsafe deposits):             891   over 437 deposits
  records touched by a withheld link:     685
contested resolutions: 437 suppressed, 78 collapsed
```

Every demonstrable false equivalence is removed; nothing else changes.

## Also verified

Changing how these deposits are treated breaks no cross-source link. 48,132 deposit CURIEs are shared between the BacDive and LPSN transforms today — but of the tokens a frequency allowlist would have rejected, **zero** are referenced by LPSN or MediaDive, at every threshold from 10 to 100. Those nodes also have no incoming edges: only 4,677 outgoing `subclass_of`.

## Tests

18 → 22. The four new ones pin the three resolutions against link-safety, and that emission happens after resolution rather than inside the record loop — the ordering that caused this. The #894 test was rewritten rather than deleted: it still requires the link, now at its new site, and still pins subject/object orientation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -95,10 +95,12 @@ and go to the source's deposit-claim report, because picking one would make
 the answer depend on file order. That report carries every deposit whose
 claimants disagreed, whichever way it went, so a coarsened taxonomy and a
 suppressed one are both visible and neither is confused with an ontology
-lookup that failed. Every record that cites a deposit links to it
-with `biolink:close_match`, so the deposit's provenance is in the graph and its
-taxonomy stays reachable through the record even when no parent is asserted.
-See issues #892 and #894.
+lookup that failed. A record that cites a deposit links to it with
+`biolink:close_match`, so the deposit's provenance is in the graph — but only
+where the deposit kept a parent. `close_match` is symmetric and maps to
+`SEMMEDDB:same_as`, so on a deposit whose claimants proved to be unrelated
+organisms it would assert an equivalence between them; those citations stay in
+the claims report. See issues #892, #894 and #899.
 
 ## Operational traps
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,7 +100,8 @@ lookup that failed. A record that cites a deposit links to it with
 where the deposit kept a parent. `close_match` is symmetric and maps to
 `SEMMEDDB:same_as`, so on a deposit whose claimants proved to be unrelated
 organisms it would assert an equivalence between them; those citations stay in
-the claims report. See issues #892, #894 and #899.
+the claims report, and the node carries a description saying so. See issues
+#892, #894, #899 and #907.
 
 ## Operational traps
 
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -29,6 +29,7 @@ from kg_microbe.transform_utils.bacdive.emission import (
     RESOLUTION_SUPPRESSED,
     RESOLUTION_SUPPRESSED_NO_ANCESTRY,
     deposit_conflict_rows,
+    resolution_asserts_a_parent,
     resolve_deposit_parents,
 )
 from kg_microbe.transform_utils.bacdive.emission import StrainProvenanceWriter as _StrainProvenanceWriter
@@ -1690,6 +1691,10 @@ class BacDiveTransform(Transform):
         # records, so claims are buffered here and resolved after the loop (see #892).
         # Shape: {kgmicrobe.strain:<deposit>: {NCBITaxon:<id>: [bacdive record key, ...]}}
         deposit_parent_claims: Dict[str, Dict[str, List[str]]] = {}
+        # Every (BacDive record, deposit) citation seen, buffered for the same reason:
+        # whether the link is safe to assert depends on what the other records citing
+        # that deposit turn out to say (#899).
+        deposit_citations: List[tuple] = []
 
         COLUMN_NAMES = [
             BACDIVE_ID_COLUMN,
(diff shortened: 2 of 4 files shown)


realmarcin:
Three review rounds. Two findings in rounds 1–2, both filed and fixed; round 3 clean.

## Round 1 — #907: 437 deposits ended up with no edges at all

Withholding the parent (#892) and the link (#899) leaves an organism-typed node carrying a culture number and nothing else — indistinguishable from a term the transform failed to process, which is the failure mode it fixes.

```
deposits left with no edges from bacdive:  437
  also referenced by LPSN:                  55
  also minted by MediaDive:                  0
  genuinely isolated:                      382
```

Dropping the node is not available: removing the row for those 55 leaves LPSN's `close_match` edges dangling for KGX to back-fill with untyped `NamedThing` stubs — exactly #895. So the node stays and explains itself instead, via a `description` naming the taxa claimed and the records that claimed them. Node rows are now buffered and written after resolution, for the same reason citations and parents are.

Verified the fix actually ships: the merged graph preserves node `description` (1,283,891 of 2,852,793 nodes in `20260815` carry one).

## Round 2 — #908: the description overstated

It ended "No parent taxon or record link is asserted" — true of this transform, false of the merged graph, since 55 of the 437 do carry an LPSN link. This is node data, not a code comment, so a reader cannot know it was written from one transform's point of view. Now scoped: *BacDive* asserts none, and other sources may still reference the node.

## Round 3 — clean

Checked and found nothing:

- **Deposit CURIEs are written as a node row in exactly one place** (AST-walked every `_create_node_row` call site). Relocating the write cannot produce a second row that `drop_duplicates` picks over the described one.
- **Label choice is deterministic** — first citation wins, and previously the stable sort kept the first-written row, so behaviour is unchanged.
- **No LPSN behaviour change** — no LPSN file is touched, and LPSN does not reproduce the equivalence: of the 55 contested deposits it links, **zero** have LPSN records that disagree on taxon.
- **The blind spot I went looking for does not exist.** The rule only withholds where there is evidence of collision, so I checked what escapes it: of the 848 deposits cited by more than one record, **zero** have no taxon-bearing claimant. Every multi-cited deposit is checkable; 9 are partially checkable and still judged on the claimants that have a taxon. The 1,094 deposits linked with no taxonomic evidence are all single-cited, so no collision is possible within BacDive.
- **Buffer cost** is ~26 MB for labels and ~34 MB for citations, against an input JSON already fully resident.

Tests 18 → 25. `tox` green apart from the pre-existing `test_ontologies_stubs` failure that also fails on master.

Issue #909, opened 2026-09-02
706,765 shipped edges use METPO:2000511, which upstream obsoleted with no declared replacement


realmarcin:
Found while working #900 — measuring what would change if METPO were pinned to a tag rather than tracked from `main`.

`METPO:2000511` is `BACDIVE_ASSAY_PREDICATE` (`kg_microbe/transform_utils/constants.py:410`), the organism → assay predicate BacDive puts on every assay observation, mapped to `biolink:related_to` in `kg_microbe/utils/metpo_predicates.py:83`.

**Upstream it is deprecated.** In `metpo.json` on `main` today:

```
METPO:2000511  lbl='obsolete has observation'  deprecated=true
    IAO_0000231 = obo/IAO_0000226      (obsolescence reason)
    (no IAO:0100001 — no replacement term declared)
```

It is also gone from `src/templates/metpo-properties.tsv` as of the `2026-06-12` tag, along with 26 other property ids.

## Scale

Scanning every `data/transformed/*/edges.tsv`:

```
METPO-predicate edges:        10,008,745  over 67 predicates
deprecated predicates in use:          1  ->  706,765 edges  (7.1%)
  METPO:2000511  'obsolete has observation'
deprecated node endpoints:             0
```

One predicate, but it carries 7% of every METPO-predicated edge in the graph. Nothing else in the graph touches an obsolete METPO term — this is the single instance, which makes it worth fixing rather than living with.

Two further ids are referenced in `kg_microbe/transform_utils/metatraits/metatraits.py` and are also obsolete upstream, though they currently produce no edges:

| id | upstream label |
|---|---|
| METPO:2000054 | obsolete has growth temperature observation |
| METPO:2000508 | obsolete has growth NaCl observation |

## Why it went unnoticed

`download.yaml` pulls `metpo.owl` / `metpo.json` from `refs/heads/main` (#900), so the obsoletion arrived with a download and changed nothing observable: the constant is hardcoded, the biolink mapping is hardcoded, and no check compares emitted predicates against the ontology. The edges kept being written against a term that had stopped existing as a live property.

## What it needs

A decision, not just a code change. `IAO:0000226` with no `IAO:0100001` means upstream removed it without nominating a successor, so there is no mechanical migration. Options, roughly:

1. Find the live METPO property that now covers organism → assay and repoint the constant.
[...]

PR #910, opened 2026-09-02
fix: pin every METPO artifact to one release, and keep them there (#900, #912)

### LINKED ISSUE #900: METPO ontology is pulled unpinned from main while its own templates are pinned at tag 2026-03-24
Found while auditing the METPO proposal state.

`download.yaml` fetches METPO two different ways:

```yaml
# unpinned — always whatever main holds at download time
  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/heads/main/metpo.owl
  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/heads/main/metpo.json

# pinned at a tag, five months old
  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-03-24/src/templates/metpo_sheet.tsv
  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-03-24/src/templates/metpo-properties.tsv
```

So the ontology and the ROBOT templates it is built from can drift arbitrarily far apart, and which pairing a run gets depends on when `kg download` happened to fetch. CLAUDE.md says `download.yaml` owns pinned upstream versions; this is the one place it does not.

This is not hypothetical drift. Between our cached `metpo.json` (fetched 2026-07-27) and `main` today, the file went from 275 KB / 403 terms to 1,044 KB / 1,619 terms — same 403 active terms, plus 1,216 deprecated ones that upstream began merging into the release (berkeleybop/metpo#600). A re-download changes what every METPO consumer in this repo sees, with nothing recording that it happened.

`kg_microbe/transform_utils/metatraits/metatraits.py::_load_metpo_binned_ranges` reads `data/raw/metpo.json` directly and does not filter `deprecated`. It happens to be unaffected — the obsolete terms carry no `range_min`/`range_max`, so the loader still finds 15 binned classes either way — but that is luck, not a guard.

Suggested:

1. Pin `metpo.owl` and `metpo.json` to a tag, matching how the templates are pinned, and move both to the same tag so ontology and templates agree.
2. Filter `deprecated` in `_load_metpo_binned_ranges` rather than relying on obsolete terms lacking range annotations.
3. Note in `download.yaml` which upstream artifact is canonical once berkeleybop/metpo#615 answers that.

Upstream context: berkeleybop/metpo#615 (releases carry no assets and nothing says what to pin), #600 (deprecated terms in the released OWL), #433 (switch source of truth to the repo OWL).

realmarcin:
Closes #900

## The problem

`download.yaml` fetched METPO two different ways:

```yaml
  url: .../berkeleybop/metpo/refs/heads/main/metpo.owl        # unpinned
  url: .../berkeleybop/metpo/refs/heads/main/metpo.json       # unpinned
  url: .../refs/tags/2026-03-24/src/templates/metpo_sheet.tsv       # three releases old
  url: .../refs/tags/2026-03-24/src/templates/metpo-properties.tsv  # three releases old
```

The ontology and the templates it is built from could drift arbitrarily far apart, and which pairing a run got depended on when `kg download` happened to fire.

**They had drifted, and it cost something concrete.** This is how `METPO:2000511` came to carry **706,765 shipped edges** after upstream obsoleted it — filed separately as #909. The obsoletion arrived with a download and changed nothing observable: the constant is hardcoded, the Biolink mapping is hardcoded, and nothing compares emitted predicates against the ontology.

## The fix

All four artifacts pin to `2026-06-12`, the current release.

**Verified safe before moving.** Every id the templates lose on the way from `2026-03-24` is already deprecated upstream:

| template | rows | dropped | of which deprecated upstream | still live |
|---|---|---|---|---|
| `metpo-properties.tsv` | 138 → 115 | 27 | **27** | 0 |
| `metpo_sheet.tsv` | 282 → 292 | 21 | **21** | 0 |

So the move drops only obsolete terms. It also picks up four properties we should already have had — `METPO:2000064 tolerates`, `METPO:2000065 does not tolerate`, `METPO:2000067 isolated from host with quality`, `METPO:2000068 isolated from environment with quality` — our own proposed terms, adopted upstream after the old pin was set and invisible to us since.

`metpo.owl` at the tag has an identical id set to `main` (1,620 either way; 202 lines of content drift), so pinning the ontology loses no terms.

## One version, one place

`kg_microbe/utils/mapping_file_utils.py` kept its own copy of the tag in two hardcoded URLs — which is how one of them fell three releases behind without the other noticing. Both now derive from `METPO_VERSION` in `transform_utils/constants.py`.

`tests/test_metpo_version_pin.py` (4 tests) makes the drift impossible to reintroduce quietly:

- no METPO URL in `download.yaml` points at a branch;
- every METPO URL carries the same tag, and it is `METPO_VERSION`;
- every URL the code uses is one `download.yaml` actually fetches — `download.yaml` cannot import the constant, so the two are checked against each other;
- `mapping_file_utils` contains no `raw.githubusercontent.com/berkeleybop/metpo` literal.

## Canary

All four pinned URLs resolve (200, expected sizes), and the pinned `metpo.json` run through `_load_metpo_binned_ranges` yields exactly the same 15 binned optimum classes (7 temperature, 4 pH, 4 NaCl) as the current cache — so the MetaTraits loader is unaffected by the 1,216 deprecated terms that arrive with the newer release. Done against a scratch copy; `data/raw` was not touched.

## Not fixed here

#909 — the 706,765 edges on the obsoleted predicate. That needs a decision about which live property replaces `has observation`, and upstream declared no successor (`IAO:0000226`, no `IAO:0100001`), so it is not a mechanical migration.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,6 +165,14 @@ KGX/BMT must use the pinned local Biolink files downloaded to
 Biolink version requires changing `download.yaml`, the lock file, fixtures, and
 tests together.
 
+METPO is pinned the same way. `METPO_VERSION` in `transform_utils/constants.py`
+is the single source of truth, and the ontology (`metpo.owl`, `metpo.json`) and
+the two ROBOT templates built from it must all move to the same tag together.
+Never fetch METPO from `refs/heads/main`: an upstream obsoletion then arrives
+with a download and changes nothing observable, which is how a deprecated
+predicate reached 706,765 shipped edges. `tests/test_metpo_version_pin.py`
+enforces this against `download.yaml`. See issues #900 and #909.
+
 ## Operational references
 
 - [Data hosting and ontology build costs](docs/DATA_HOSTING.md)
--- a/download.yaml
+++ b/download.yaml
@@ -351,11 +351,11 @@
 # METPO
 #
 -
-  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/heads/main/metpo.owl
+  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-06-12/metpo.owl
   local_name: metpo.owl
   tag: ontologies
 -
-  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/heads/main/metpo.json
+  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-06-12/metpo.json
   local_name: metpo.json
   tag: ontologies
 
@@ -600,11 +600,11 @@
 
 # Immutable METPO templates consumed by multiple trait transforms
 -
-  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-03-24/src/templates/metpo_sheet.tsv
+  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-06-12/src/templates/metpo_sheet.tsv
   local_name: metpo_sheet.tsv
   tag: mappings
 -
-  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-03-24/src/templates/metpo-properties.tsv
+  url: https://raw.githubusercontent.com/berkeleybop/metpo/refs/tags/2026-06-12/src/templates/metpo-properties.tsv
   local_name: metpo-properties.tsv
   tag: mappings
 
--- a/kg_microbe/transform_utils/constants.py
+++ b/kg_microbe/transform_utils/constants.py
@@ -776,6 +776,19 @@ UNIPROT_DATA_LIST = [
 ]
 
 BACDIVE_MAPPING_FILE = "bacdive_mappings.tsv"
+
+# The METPO release every METPO artifact in this repo is pinned to: the ontology
+# (metpo.owl / metpo.json) and the two ROBOT templates built from it. They must move
+# together — tracking the ontology from `main` while the templates sat on an older tag
+# is how an obsoleted predicate reached 706,765 shipped edges unnoticed (#900, #909).
+# `download.yaml` repeats these URLs and cannot import them; a test asserts they agree.
+METPO_VERSION = "2026-06-12"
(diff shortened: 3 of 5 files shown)


realmarcin:
Two review rounds. Round 1 found two things, one fixed here and one filed as its own issue; round 2 clean.

## Round 1

**#912 — fixed here.** The refactor added a *second* import from `transform_utils.constants`, nineteen lines into the body with an `# noqa: E402`. There is no cycle to work around — `constants.py` imports nothing from `kg_microbe` — so the suppression was silencing a rule that was right, and it reads as though someone hit a circular-import problem. Merged into the existing line-13 import.

**#911 — filed, not fixed here.** `kg download` skips any entry whose `local_name` exists and records nothing about which URL produced it, so changing a pin in `download.yaml` has no effect on disk. That is a downloader-semantics problem affecting every pinned artifact, not something to bolt onto this PR.

It does mean this PR's practical reach is split, which is worth stating plainly:

| artifact | cached in `data/raw`? | new pin takes effect |
|---|---|---|
| `metpo_sheet.tsv`, `metpo-properties.tsv` | no | **immediately** — `_metpo_template_response` falls through to the pinned URL |
| `metpo.owl`, `metpo.json` | yes, 275 KB from 2026-07-27 | only after `kg download -i -t ontologies` |

So the mapping side of the move is live on the next transform run; the ontology side needs an explicit invalidation, which will come with the pending transform rerun anyway.

## Round 2 — clean

- **No stale reference to `2026-03-24` remains** anywhere outside a test docstring that deliberately names it.
- **No other hardcoded METPO URL exists** in code or config; the one in `.claude/skills/add-transform/SKILL.md` is a `<tag>` placeholder.
- **The local-first fallback does not defeat the pin.** `_metpo_template_response` prefers `data/raw/<template>` over the network, which would have made the change inert twice over — but neither template is cached, so both fetch from the pinned URL.
- **Tag granularity** is coarser than it looks: `download.yaml` entries carry a single `tag:` string (no list form, checked across all 63 entries), and METPO shares `ontologies` with 16 others, so invalidating METPO re-downloads NCBITaxon too. Recorded on #911, where it belongs, rather than retagging here — moving METPO to its own tag would silently drop it from `-t ontologies`.

Issue #912, opened 2026-09-02
mapping_file_utils imports from constants twice, the second mid-file with a noqa


realmarcin:
Found in the review of #910.

#910 moved the METPO template URLs into `transform_utils/constants.py` and re-exported them, but added a second import statement rather than extending the existing one:

```python
13: from kg_microbe.transform_utils.constants import PREFIXMAP_JSON_FILEPATH, RAW_DATA_DIR
...
32: from kg_microbe.transform_utils.constants import (  # noqa: E402
33:     METPO_CLASSES_ROBOT_TEMPLATE_URL,
34:     METPO_PROPERTIES_ROBOT_TEMPLATE_URL,
35: )
```

Two imports from the same module, the second nineteen lines into the body with an `E402` suppression. The `noqa` is the tell: it silences a lint rule that was correctly complaining, and there is no reason to — `constants.py` imports nothing from `kg_microbe`, so there is no cycle to work around and the import belongs on line 13 with the others.

Cheap to fix, and worth fixing rather than leaving: a mid-file import with a suppression reads as though someone hit a circular-import problem and worked around it, which invites the next person to preserve the workaround.

realmarcin:
Fixed on master by `fc8a151e` (PR #910), whose commit body named `(#912)` without a closing keyword — hence this outliving its fix.

Verified on `fc8a151e`:

```
$ grep -n "from kg_microbe.transform_utils.constants import" kg_microbe/utils/mapping_file_utils.py
13:from kg_microbe.transform_utils.constants import (
```

One import, at the top, no `noqa`. `METPO_CLASSES_ROBOT_TEMPLATE_URL` and `METPO_PROPERTIES_ROBOT_TEMPLATE_URL` now come in alongside `PREFIXMAP_JSON_FILEPATH` and `RAW_DATA_DIR`, with a comment at the old site saying where they come from and why the module no longer spells them out. `ruff check` passes with no suppression, confirming there was no cycle to work around.

Issue #828, opened 2026-08-19
data/merged/ carries January merged-kg_default_*.tsv next to the current tarball


realmarcin:
Noticed while reviewing #826 — it actively misled the review.

`data/merged/` contains both:

    merged-kg.tar.gz               137 MB   Aug 15   <- current
    merged-kg_default_nodes.tsv    208 MB   Jan 30   <- stale
    merged-kg_default_edges.tsv   1034 MB   Jan 30   <- stale

The external reviewer read the `_default_` TSVs and reported the merged graph as 1.51M nodes / 6.13M edges, contradicting the 2,852,793 / 14,664,554 in the current tarball, and flagged it as an unexplained discrepancy in the PR. Both numbers were right about different files; nothing indicates which is current except the mtime.

They are also 1.2 GB of stale bytes sitting in the working tree.

Worth deciding: are the `_default_` TSVs a deliberate uncompressed convenience copy that should be regenerated on every merge, or a leftover? If the former, the merge should rewrite them (or delete them when it cannot); if the latter, they should be removed and gitignored.

Either way an artifact that is seven months stale, in the directory consumers read, with no marker saying so, is the same shape as #812 — a thing that looks current and is not.

realmarcin:
Addressed in #846 — the merge now reports artifacts in its output directory that it did not write, with size and age. It warns rather than deletes: 1.2 GB of untracked files may be a deliberate copy, and removing them is not a merge step's call.

The sharper version of this is #847: all six configs write to the same `merged-kg.tar.gz`, so a variant run overwrites the standard graph in place with nothing recording which config produced it. That is likely where the `merged-kg_default_*` files came from.

PR #913, opened 2026-09-02
feat(merge): check the strain-parent invariant where sources actually meet (#896, #914, #915, #916)

### LINKED ISSUE #896: The multi-parent guard is per-transform; nothing checks the merged graph for strains with several NCBITaxon parents
Found in the review of #893 (fixes #892).

#893 buffers deposit parent claims inside the BacDive transform and asserts a parent only when every claiming record agrees. That is correct for BacDive but it is a **per-transform** guard: it only sees claims made by the transform that holds the buffer.

`kgmicrobe.strain:*` is a deliberately shared namespace — LPSN's `_extract_strain_curies` mints the same CURIEs so the merge reconciles both sides. Today no other source adds a competing `subclass_of` to NCBITaxon on those nodes, so the guard is sufficient:

```
subclass_of edges with a kgmicrobe.strain subject and an NCBITaxon object
  bacdive           251,916
  metatraits             58   (kgmicrobe.strain:Genus_species_STRAIN shape, 0 overlap with bacdive)
  lpsn / lpsn_api / mediadive   0
  microbedecoder    19,114   (all lpsn: objects, not NCBITaxon)
strains with >1 NCBITaxon parent across all four: 525 — all of them from bacdive alone
```

But that is a property of the current source set, not something enforced. A future transform that emits `kgmicrobe.strain:<code> subclass_of NCBITaxon:X` reintroduces #892 silently, and #893's guard cannot see it because the conflicting claim never enters its buffer.

Proposed: a merge-stage or post-merge QC assertion — every `kgmicrobe.strain:*` node has at most one `biolink:subclass_of` edge to an `NCBITaxon:*` object — reported alongside the other merged-graph stats, so the invariant is checked where sources actually meet rather than in each transform separately.

Related: #892 (the original report), #893 (the BacDive-side fix), #894 (deposit nodes lack a link back to their asserting records).

realmarcin:
Closes #896

## Why the merge and not the transform

#892 stopped BacDive giving a strain node several NCBITaxon parents. That guard buffers claims inside the transform, so it can only see claims BacDive itself makes. `kgmicrobe.strain:*` is a namespace several sources mint into — LPSN's `_extract_strain_curies` mints the same CURIEs precisely so the merge reconciles both sides — so the rule is only really enforced after the merge.

Today nothing else emits a competing parent:

```
subclass_of edges with a kgmicrobe.strain subject and an NCBITaxon object
  bacdive         251,916
  metatraits           58   (Genus_species_STRAIN shape, 0 overlap with bacdive)
  lpsn / lpsn_api / mediadive   0
  microbedecoder   19,114   (lpsn: objects, not NCBITaxon)
```

But that is a property of the current source set, not something enforced. A future transform reintroduces #892 silently, and #892's own guard cannot see it.

## What it does

`kg_microbe/merge_utils/invariants.py` streams the merged edge file and reports any `kgmicrobe.strain:*` node carrying more than one **distinct** `NCBITaxon:*` `subclass_of` parent, naming the knowledge sources involved — so the report says who to go and fix, not only what broke. It runs from `_cleanup_merged_outputs`, alongside the TSV normalization that already happens there.

`merged_strain_parent_violations.tsv` is written beside the merged TSVs, atomically, **and written even when empty** — an absent report cannot be told from a check that never ran, and a clean merge is the thing worth being able to demonstrate.

**Warns rather than aborts.** The merge is a multi-hour operation and this finding is about data quality, not a corrupt artifact; failing at the end would cost the run and fix nothing.

## Canary

Run against the shipped `data/merged/20260815` graph:

```
strains with >1 NCBITaxon parent: 524   (18s over a 2 GB edge file)
  kgmicrobe.strain:AJ-1378     2  NCBITaxon:43767|NCBITaxon:698417    infores:bacdive
  kgmicrobe.strain:AS-1        2  NCBITaxon:1122236|NCBITaxon:398036  infores:bacdive
  kgmicrobe.strain:AS-1.2123   2  NCBITaxon:1114856|NCBITaxon:1227500 infores:bacdive
knowledge sources implicated: ['infores:bacdive']
```

**524 is exactly the number #892 reported for that build**, and the check attributes all of them to the source that caused it. So it demonstrably finds the historical bug on the historical graph, which is the only way to know it would catch a new one.

## Tests

8, covering the cases that would make it useless rather than merely absent:

- one parent is not a violation, or the report is noise;
- **the same parent asserted by two sources is not a violation** — two sources agreeing is the merge working, and counting rows rather than distinct parents would fire on every strain a future source correctly corroborates;
- only `subclass_of` to `NCBITaxon` counts — not `close_match` to a deposit (#894) or microbedecoder's `strain → lpsn:` edges, neither of which is a taxonomic parent;
- the report names sources, not just nodes;
- a clean run still writes the report;
- an unexpected header is skipped, not crashed — a merge that took hours must not die on a header this check did not expect;
- an empty file does not raise.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -122,6 +122,12 @@ the claims report, and the node carries a description saying so. See issues
 - Cache existence must imply completeness. Use `atomic_write` from
   `kg_microbe.utils.atomic_io`, including completion markers where the helper
   requires them; never publish a partially written cache path.
+- The merge checks invariants that no single transform can. A transform only
+  polices the edges it writes, and `kgmicrobe.strain:*` is minted by several
+  sources, so `merge_utils/invariants.py` re-checks after the merge and writes
+  `merged_strain_parent_violations.tsv` beside the merged TSVs — empty on a
+  clean run, because an absent report cannot be told from a check that never
+  ran. See issues #892 and #896.
 - Ontology acquisition or adapter failures abort the run. Do not catch them as
   per-row lookup failures, and construct/resolve adapters in the parent process
   before creating a `Pool`; adapters are large and generally not picklable.
--- a/kg_microbe/merge_utils/invariants.py
+++ b/kg_microbe/merge_utils/invariants.py
@@ -1,0 +1,139 @@
+"""
+Invariants checked against the merged graph, where sources actually meet.
+
+Each transform can only police the edges it writes itself. ``kgmicrobe.strain:*``
+is a deliberately shared namespace -- LPSN mints the same CURIEs BacDive does so
+the merge reconciles them -- so a rule about those nodes is only really enforced
+after the merge. #892 fixed multi-parent strains inside the BacDive transform;
+this module is the check that notices if another source reintroduces them (#896).
+"""
+
+import csv
+import logging
+from collections import defaultdict
+from pathlib import Path
+from typing import Dict, List, Optional, Set, Tuple
+
+from kg_microbe.transform_utils.constants import (
+    NCBITAXON_PREFIX,
+    OBJECT_COLUMN,
+    PREDICATE_COLUMN,
+    PRIMARY_KNOWLEDGE_SOURCE_COLUMN,
+    STRAIN_PREFIX,
+    SUBCLASS_PREDICATE,
+    SUBJECT_COLUMN,
+)
+
+logger = logging.getLogger(__name__)
+
+#: Written beside merged-kg_nodes.tsv / merged-kg_edges.tsv.
+STRAIN_PARENT_REPORT = "merged_strain_parent_violations.tsv"
+
+STRAIN_PARENT_REPORT_HEADER = ["strain_id", "parent_count", "parents", "knowledge_sources"]
+
+
+def find_multi_parent_strains(
+    edges_file: Path,
+) -> Dict[str, Tuple[Set[str], Set[str]]]:
+    """
+    Return strain nodes carrying more than one NCBITaxon parent in the merged graph.
+
+    Streams the edge file rather than loading it: the merged graph is tens of
+    millions of rows, and only the strain rows are retained.
+
+    :param edges_file: Path to ``merged-kg_edges.tsv``.
+    :return: ``{strain CURIE: (parent CURIEs, knowledge sources)}`` for every strain
+        with more than one distinct parent. Empty when the invariant holds.
+    """
+    parents: Dict[str, Set[str]] = defaultdict(set)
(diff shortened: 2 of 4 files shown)


realmarcin:
Three review rounds, one finding each, all filed and fixed.

## Round 1 — #914: the check could suppress artifact delivery

It sat between `_normalize_edges_tsv` and `_rewrite_tarball`, inside a function `load_and_merge` wraps in a blanket `except Exception`. So any failure in the check — an unexpected column type, a permissions problem, a bug in the loop — would skip the tarball rewrite and leave normalized loose TSVs beside a **stale archive**, with the merge still reporting success and one printed line as the only trace. That is exactly the confusion `_warn_about_stale_siblings` exists to prevent (#828).

A data-quality report must not be able to decide whether the artifact ships. It now has its own handler. Moving the call after `_rewrite_tarball` would not work — when the tarball was extracted for normalization the loose TSVs are unlinked immediately after, so the edge file may be gone.

The regression test drives `_cleanup_merged_outputs` with a check that raises and asserts the rewrite still happened. **Verified by mutation**: with the handler removed it fails with `RuntimeError: simulated check failure`.

## Round 2 — #915: last-wins column resolution, failing in the reassuring direction

The column index was a dict comprehension, which is last-wins on a repeated column name. `merge_kg._first_index` exists precisely because merged headers can repeat one — KGX takes the column union across sources — and the module next door already uses it.

On a header repeating `subject`, the check would read the wrong column, find no strain subjects, and **report a clean graph**. Silent, and in the reassuring direction, which is the worst way for a check like this to fail.

`_normalize_edges_tsv` writes a canonical deduplicated header immediately before, so it should not arise on the merge path — but `find_multi_parent_strains` takes an arbitrary path and is the natural thing to point at a historical graph under `data/merged/` that predates that normalization. **Verified by mutation**: the new test fails with the dict comprehension restored.

## Round 3 — #916: my own #912 pattern, one PR later

`check_merged_invariants` imported `atomic_write` inside the function body with no cycle to avoid — `atomic_io` imports only stdlib. Filed rather than folded in silently, because the point of the pass is that findings get recorded including the ones the reviewer caused.

## Canary, re-run after all three

```
strains with >1 NCBITaxon parent: 524   (19s over the 2 GB data/merged/20260815 edge file)
```

Still exactly the 524 that #892 reported for that build, unchanged by the column-resolution and isolation fixes.

Tests 8 → 10. `tox` green apart from the pre-existing `test_ontologies_stubs` failure that also fails on master.

PR #917, opened 2026-09-02
fix(bacdive): give unresolvable NCBITaxon parents a typed stub node (#895, #919)

### LINKED ISSUE #895: 113 NCBITaxon parents cited by BacDive strain edges are absent from the pinned NCBITaxon extract

realmarcin:
Closes #895

## Correcting the count first

I wrote #895, and its headline number is right about the wrong thing. 113 strain parents are absent from the pinned NCBITaxon extract — but BacDive writes 20,410 NCBITaxon node rows of its own, so most of them already arrive typed:

```
strain subclass_of parents absent from the ontologies extract:   113
  already carrying a BacDive-written node row:                    94
  no node row anywhere -> untyped biolink:NamedThing:             19   (18 bacdive, 1 metatraits)
```

19 is the number that matches what #892 described.

## Why they are unresolvable

All 113 checked against the full NCBITaxon adapter:

| | count | meaning |
|---|---|---|
| adapter does not know the id | 110 | retired or merged by NCBI; BacDive carries a stale taxid |
| adapter knows it, extract does not | 3 | `Bosea`, `Moorella`, `Micromonas sp.` — whole lineage missing, parents included |

None appear anywhere in the extract, in any column, **not even as an alternative id** — so there is no in-repo forwarding table to migrate them with.

## Dropping was measured and rejected

The obvious fix is to refuse an edge whose object will have no node:

```
strains touching a dangling parent:        862
strains that would lose their ONLY parent: 857   (300 record nodes, 557 deposits)
```

Orphaning 857 strains from the taxonomy is worse than an unresolvable parent. The edges stay.

## What this does

Emits a typed stub, the pattern #790 established for isolation-source targets — `biolink:OrganismTaxon`, the adapter's label where recoverable, and a description that distinguishes the two causes:

```
NCBITaxon:216400  biolink:OrganismTaxon  Moorella
    Referenced by BacDive but absent from the pinned NCBITaxon release
    (label recovered from the ontology adapter).

NCBITaxon:2923281 biolink:OrganismTaxon  NCBITaxon:2923281
    Referenced by BacDive but absent from the pinned NCBITaxon release,
    including as an alternative id; most likely retired or merged by NCBI.
```

**Implemented as a pass over the transform's own edge output**, not hooked into the ~10 places that emit a taxon edge. A future emission path is covered without anyone having to remember it — and the test suite pins that by checking a `related_to` edge is covered too, not just `subclass_of`.

## Canary

Driven over the real `data/transformed/bacdive/edges.tsv` (4.2M rows) against a copy of the real `nodes.tsv`:

```
extract labels loaded: 925,220
stubs written: 18   (3s)
  with a recovered label:  1   (NCBITaxon:216400 -> 'Moorella')
  unlabelled:             17
```

18 rather than 19 because the remaining one is referenced only by metatraits, which this transform cannot see.

## Tests

7, including the two that would make this actively harmful if wrong:

- **a taxon already in the pinned release gets no stub** — writing our own would duplicate 925,220 nodes and override their real labels;
- **a taxon this run already wrote gets no second row** — `drop_duplicates` keeps one row per id, so a duplicate would replace a real label with a bare CURIE.

## Not fixed here

The 1 metatraits-sourced untyped taxon. That wants the cross-source check, not a second copy of this pass — filing as a follow-up against the `merge_utils/invariants.py` module added in #913.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -957,6 +957,75 @@ class BacDiveTransform(Transform):
         self._ncbitaxon_ancestor_cache[ncbitaxon_id] = ancestors
         return ancestors
 
+    def _emit_stubs_for_unresolvable_taxa(self) -> int:
+        """
+        Give every NCBITaxon this run pointed at a typed node, or KGX invents one.
+
+        A ``subclass_of`` edge whose object has no node row anywhere makes KGX
+        synthesize a bare ``biolink:NamedThing`` with no label. #892 saw 1,652 of
+        those in the 20240826 build and asked whether they were still being
+        emitted; 113 were. They are taxids BacDive carries that the pinned
+        NCBITaxon release does not contain -- retired or merged by NCBI, or newer
+        than the release -- and none of them appear anywhere in the extract, not
+        even as an alternative id.
+
+        Dropping the edges instead was measured and rejected: 857 of the 862
+        affected strains have no other parent, so removing the edge orphans them
+        from the taxonomy entirely, which is worse than an unresolvable parent.
+
+        Reads back this run's own edges rather than hooking the ten places that
+        emit a taxon edge, so a path added later is covered without being
+        remembered.
+
+        :return: Number of stub nodes written.
+        """
+        edge_objects: set = set()
+        with open(self.output_edge_file, encoding="utf-8", errors="replace") as handle:
+            handle.readline()
+            for line in handle:
+                fields = line.split("\t")
+                if len(fields) > 2 and fields[2].startswith(NCBITAXON_PREFIX):
+                    edge_objects.add(fields[2])
+        if not edge_objects:
+            return 0
+        node_ids: set = set()
+        with open(self.output_node_file, encoding="utf-8", errors="replace") as handle:
+            handle.readline()
+            for line in handle:
+                node_id = line.split("\t", 1)[0]
+                if node_id.startswith(NCBITAXON_PREFIX):
+                    node_ids.add(node_id)
+        # Ids in the extract get their node row from the ontologies transform at
+        # merge time, so they are not missing even though we did not write them.
+        missing = sorted(edge_objects - node_ids - set(self.ncbitaxon_labels))
+        if not missing:
+            return 0
+        with open(self.output_node_file, "a", encoding="utf-8", newline="") as handle:
+            writer = csv.writer(handle, delimiter="\t")
+            for curie in missing:
+                label = self._get_ncbitaxon_label(curie)
+                writer.writerow(
+                    self._create_node_row(
+                        curie,
+                        NCBI_CATEGORY,
+                        label or curie,
+                        description=(
+                            "Referenced by BacDive but absent from the pinned NCBITaxon release"
+                            + (
+                                " (label recovered from the ontology adapter)."
+                                if label
+                                else ", including as an alternative id; most likely retired or merged by NCBI."
+                            )
+                        ),
+                    )
+                )
+        logger.info(
+            "[bacdive] %s NCBITaxon parents are not in the pinned release; "
+            "wrote typed stub nodes so they do not become untyped NamedThing (#895)",
+            f"{len(missing):,}",
+        )
+        return len(missing)
+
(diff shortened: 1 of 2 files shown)


realmarcin:
Three review rounds. Rounds 1 and 3 clean; round 2 found something worth more than the original issue.

## Round 1 — two candidates, both dissolved on verification

**Mixed line endings.** The stub pass appends with `csv.writer`, whose default terminator is `\r\n`, while the existing `nodes.tsv` is entirely bare LF. Measured: the canary file did come out with 18 CRLF rows among 275,388 LF ones. But `drop_duplicates` runs immediately after the append and rewrites the file through pandas, normalising everything — so it never reaches disk that way in the real flow. Not filed.

**Dedup ranking preferring the stub.** `drop_duplicates(dedup_on_sort_column=True)` ranks rows by "non-empty name, then non-empty description, then most populated", and a stub has both — so it would win against a real row that lacks a description, replacing a real label with a bare CURIE. It cannot happen, because the pass skips any id that already has a node row, and there is a test pinning exactly that. Not filed.

## Round 2 — #919: the rows this PR relies on were themselves wrong

Checking the quality of the 94 unresolvable taxa that already have a BacDive node row turned up this:

```
NCBITaxon:1037355  name = "Parabacteroides chartae NS31-3 is an anaerobe, mesophilic,
                           Gram-negative bacterium that was isolated from wastewater
                           from paper mill."
```

`bacdive.py` fell back to BacDive's `General.description` when a taxon could not be labelled, and wrote that sentence as the node's **name**. 93 of 20,410 NCBITaxon rows carried one, averaging 130 characters, with the `description` column sitting empty beside them.

It is worse than being unlabelled, because it passes every "does this node have a name" check while being unusable for lookup, display or matching — and it is the *same population* as #895, approached from the other side. #895 counted the 19 unresolvable taxa arriving with no node row; these 93 do get a row, and it was wrong.

Both write sites now fall back to the CURIE and pass the sentence as `description`, matching what this PR's stub pass already does. The two paths should agree, and now do.

## Round 3 — clean

Checked whether the prose-as-name pattern reached any other node type. It does not:

```
node prefixes whose name looks like prose:
  NCBITaxon   93 / 20,410
  (no other prefix)
```

Also confirmed `ncbi_label` has no other consumer — the strain node's own label is built from `ncbitaxon_id`, not from it — so no longer defaulting it to the description sentence changes nothing else.

Tests 7 → 9. `tox` green apart from the pre-existing `test_ontologies_stubs` failure that also fails on master.

PR #920, opened 2026-09-02
fix(metpo): sync the proposal to the release and make the diff reproducible (#901)

### LINKED ISSUE #901: The METPO release-diff report is a release behind, its generator is gone, and its three action items were never filed
Found while auditing the METPO proposal state.

`docs/METPO_PROPOSALS_SUMMARY.md` and `docs/metpo/metpo_2026_06_02_release_diff.md` are the two documents that say where our proposal stands against upstream. Both have gone stale in ways that make them misleading rather than merely old.

## The diff is one release behind, and cannot be rerun

`docs/metpo/metpo_2026_06_02_release_diff.md` compares our proposal TSVs against the METPO `2026-06-02` release. `2026-06-12` has been Latest since June. The document itself says to rerun on each release, and then:

> Source script (kept under `/tmp` while iterating; rerun whenever a new METPO release ships): the diff is generated by `/tmp/metpo-release/diff_proposals.py`

That script is gone. The report cannot be regenerated, so the instruction it gives cannot be followed.

It also compares against 37 classes and 7 properties. The templates now hold 49 and 9 — the 12 PREGO-cohort classes (#677) and two properties (`METPO:2000069`, `METPO:2000802`) landed after the diff was taken, so a quarter of the proposal has never been diffed against any release.

## Its three action items were never filed

The report ends with recommended next steps and none became an issue:

1. **Definition wording** — upstream rewrote every proposed definition into genus-differentia form (`"A phenotypic quality describing …"` → `"A phenotype characterized by …"`), per berkeleybop/metpo#460. Our TSVs still carry the old wording, so every future diff re-reports the same 28 rows.
2. **Parent reassignment** — 12 classes we proposed under `METPO:1000000` were re-parented to `METPO:1000059` (`phenotype`) on ingest: `1007005, 1007062, 1007070, 1007072, 1007073, 1007074, 1007076, 1007077, 1007080, 1007081, 1007082, 1007089`. That is the curator convention and our TSVs do not follow it.
3. **Dropped xrefs** — 7 GO/PATO `oboInOwl:hasDbXref` values did not survive into the release (`METPO:1007005`→GO:0001539, `1007062`/`1007063`→PATO:0000052, `1007076`→GO:0042597, `1007077`→GO:0042710, `1007083`→GO:0004096, `1007085`→GO:0004129, `1007087`→GO:0009039). Never confirmed as intentional, never re-encoded, never re-proposed.

Item 2 is the one with teeth: proposing terms under the wrong parent guarantees they get silently re-parented again on the next adoption round, and our local TSV keeps disagreeing with the released ontology.

## `docs/METPO_PROPOSALS_SUMMARY.md` predates the current approach

Dated 2026-03-28 and headed "REVISED to use predicate-based patterns", describing 9 new data properties and a fermentation/electron-acceptor framing. Everything since — the tolerance pair, the isolation-quality pair, the PREGO cohort, `is nomenclatural type of` — is absent. Anyone reading it for current state gets a picture two approaches old.

Suggested: commit the diff generator under `scripts/`, rerun it against the current release, apply the mechanical A/B/D fixes to the proposal TSVs, and either refresh or retire the summary doc.

Related: #750 (7 of 9 properties still carry `TODO:add_citation`), #900 (METPO pulled unpinned from main), berkeleybop/metpo#424 (the submission, now refreshed with current counts), berkeleybop/metpo#528 (nine proposed terms contested upstream).

realmarcin:
Closes #901

## The report could not be refreshed

`docs/metpo/metpo_2026_06_02_release_diff.md` said where our METPO proposal stood against upstream, told the reader to rerun it on each release, and named a generator that lived in `/tmp`. So the instruction could not be followed, and the report went a release stale with its three action items unfiled — and growing:

| action item | when the report was written | on master before this PR |
|---|---|---|
| classes parented at `METPO:1000000` | 12 | **14** |
| rows carrying GO/PATO xrefs | 7 | **10** |
| definitions in the superseded wording | — | **10** |

`scripts/diff_metpo_proposals.py` now lives in the repo and regenerates the report against `METPO_VERSION` or any tag. Everything below is what it found.

## Definitions — adopt the release wording

Upstream rewrites every definition into genus-differentia form on ingest ([metpo#460](https://github.com/berkeleybop/metpo/pull/460)):

```
ours     A phenotypic quality describing the arrangement pattern of flagella on a bacterial or archaeal cell...
release  A phenotype characterized by the arrangement pattern of flagella on a bacterial or archaeal cell...
```

The generator now carries the released wording verbatim for all 21 landed terms, so the diff stops re-reporting the same rows forever.

## Parents — one constant, fourteen rows

All 14 classes we proposed under `METPO:1000000` are in the release under `METPO:1000059`. Not guessed — read from the release, term by term. `_PHENO_PARENT` moves once and the set follows. **ROBOT template + ELK reasoning passes with no UNSAT classes.**

## Xrefs — answered, and not about us

The old report's third action item was "confirm with METPO curators whether these xrefs were dropped intentionally". Answered empirically:

```
terms in the 2026-06-12 release:            1,619
carrying any oboInOwl:hasDbXref:                0
carrying a GO: or PATO: xref anywhere:          0
```

Release-wide policy, not an oversight on our rows. The `xrefs` column stays as our own record of which GO/PATO term each proposal corresponds to, but it will not land; upstream is moving cross-ontology mappings to SSSOM ([metpo#344](https://github.com/berkeleybop/metpo/issues/344)).

## Result

```
                    before        after
classes  landed unchanged   8   ->   31
         landed differing  23   ->    0
         not in release    18        18
properties landed unchanged 3   ->    4
         landed differing   1   ->    0
         not in release     5         5
```

Everything upstream has adopted now matches the release exactly. What remains outstanding is what upstream has **not** adopted — the six HIGH environmental-tolerance min/max classes, the twelve PREGO-cohort classes, and five properties including both CRITICAL optimum-value ones. That is [metpo#528](https://github.com/berkeleybop/metpo/issues/528)'s subject, a substantive objection rather than a process gap on our side.

## Guards

The generator is edited, never the TSVs — `test_extract_metpo_proposals.py` is a regenerate-and-diff gate. A new `tests/test_metpo_proposal_release_sync.py` guards the two conventions that caused the drift: a class proposed at the ontology root, or a definition in the superseded wording, now fails the suite. It also asserts the diff generator is in the repo, since an instruction to rerun a script nobody has is what started this.

`docs/METPO_PROPOSALS_SUMMARY.md` was dated 2026-03-28 and described a fermentation/electron-acceptor framing two approaches out of date. Rewritten to explain what the report means rather than be a stale second copy of it.

## One thing worth noting

The diff caught a bug in my own sync before it landed: `textwrap` split `non-NaCl` at the hyphen and rejoining with a space produced `non- NaCl`, which showed up as the single remaining difference. Fixed with `break_on_hyphens=False`. The report earning its keep on its first run is the argument for having it in the repo.


realmarcin:
Three review rounds. Two findings, both filed; round 3 clean.

## Round 1 — #921: the new report reproduced the failure it was meant to fix

#901 existed because a report told the reader to rerun a generator that no longer existed. This PR put the generator in the repo — and then emitted a report naming neither the script nor the command. A reader opening `docs/metpo/metpo_proposal_release_diff.md` in six months would be right back in #901's position, holding a stale file with no stated way to refresh it and no signal that hand-editing is pointless. The command existed only in `METPO_PROPOSALS_SUMMARY.md`, a different file the report does not link to.

The report now carries its own provenance header with the exact invocation, and says the proposal TSVs are generated too. A test asserts it.

## Round 2 — berkeleybop/metpo#616: adopting the release wording erased a claim that was never realised

Checking whether any of the 20 adopted definitions dropped content: one did.

```
ours     ... Parent class for METPO:1005031-1005037 (peritrichous, polar, ...).
release  ... (e.g. peritrichous, polar, ...)
```

The DEF-FORM rewrite turned an assertion about intended children into an illustrative list. That would be fine if upstream had made it structural. It did not:

```
children of METPO:1007005 'flagellar arrangement' in 2026-06-12:  1  (METPO:1007006, also ours)
METPO:1005031-1005037 peritrichous/polar/amphitrichous/...:        all under METPO:1000704 'flagellated'
```

`flagellated` is a **boolean** phenotype; the seven are mutually exclusive values on an **arrangement** axis. So the release has a near-empty grouping class sitting beside the seven siblings it was created to group, with the axis split across two parents — exactly what the class was proposed to fix. Filed upstream as berkeleybop/metpo#616.

The claim now lives in the `observations` column, which is our own note field and does not ship upstream, so it survives the next definition rewrite. Also gave the fetch a 60s timeout — this is a script people are told to run, and an unbounded hang looks identical to a slow release.

## Round 3 — clean

Widened the same check from "lost a METPO CURIE" to "lost a structural phrase" (`parent class for`, `subaspect of`, `subclass of`, `equivalent to`, …). Two hits, and the second is a false positive worth stating: `METPO:1007063 colony shape` lost `"Subaspect of colony morphology"`, but the genus-differentia rewrite *encodes* that — `"A colony morphology characterized by ..."` — and the release does give it `METPO:1007062` as its parent. Verified structurally rather than assumed.

ROBOT template + ELK reasoning still passes with no UNSAT classes after the `observations` change, and the report still regenerates byte-identically.

Issue #922, opened 2026-09-02
The behaviour fingerprint treats a docstring change as a behaviour change


realmarcin:
Found while working #909: editing a **docstring** in `gold.py` — no code touched — failed `test_gold_merge_wiring.py::StaleOutputGuardTest::test_the_gold_output_was_built_by_the_current_code`, marking gold's output stale and demanding a rerun of an expensive transform for a documentation fix.

`_behaviour_digest` in `kg_microbe/utils/transform_fingerprint.py`:

```python
return hashlib.sha256(ast.dump(ast.parse(source.decode("utf-8"))).encode("utf-8")).digest()
```

`ast.dump` includes docstrings, because a docstring is an `ast.Expr` holding an `ast.Constant` like any other string literal. So:

```
whitespace-only change -> same digest   (this is what #880 fixed, and it works)
docstring-only change  -> DIFFERENT digest
```

#880 made the fingerprint content-based specifically so that a formatting sweep would not invalidate every transform's output. A docstring is behaviour-neutral for exactly the same reason whitespace is — it cannot change what the transform emits — so it belongs on the same side of the line, and right now it does not.

## Why it matters beyond the annoyance

The cost is asymmetric and pushes the wrong way. Reruns here are hours, so a contributor who notices that improving a comment will invalidate a transform's output has an incentive to leave the comment wrong. In this instance the docstring in question cites `METPO:1001000|biolink:Procedure` as an established precedent, a pairing #909 removes — so it is now actively misleading, and I reverted the correction rather than force a gold rebuild for prose.

## Fix

Strip docstrings before hashing — walk the parsed tree and drop the leading `Expr(Constant(str))` from `Module`, `ClassDef`, `FunctionDef` and `AsyncFunctionDef` bodies, then `ast.dump` the result. `tests/test_no_deprecated_metpo_terms.py::_curies_in_code` already does exactly that walk and can be the model.

Note this changes every transform's digest once, so it needs `FINGERPRINT_VERSION` bumped, and the first run after it will report everything stale — which is why it wants its own change rather than riding along with #909.

Related: #879 and #880 (the content-based fingerprint this extends), #909 (where it surfaced).

Issue #911, opened 2026-09-02
Changing a pinned URL in download.yaml has no effect until someone knows to invalidate the cache


realmarcin:
Found in the review of #910.

`kg download` skips any entry whose `local_name` already exists, and nothing records which URL produced the cached file. So editing a pinned version in `download.yaml` changes nothing on disk, and the pipeline keeps consuming the old artifact with no indication that the declared pin and the cached bytes disagree.

Concretely, right now on `master` with #910 applied:

```
download.yaml declares: .../refs/tags/2026-06-12/metpo.json
data/raw/metpo.json    :   275,086 bytes   (fetched 2026-07-27 from main, 403 terms)
the pinned artifact    : 1,044,711 bytes   (1,619 terms)
```

A `kg download` run reports success and leaves the 275 KB file in place. The only way to pick up the new pin is `kg download -i -t ontologies`, which a reader has to know to run — and `-i` invalidates *every* entry carrying that tag, so it is a blunt instrument. For MediaDive, CLAUDE.md notes invalidation can trigger an hour-long crawl, which is a real disincentive to reaching for it.

This is not METPO-specific. Every pinned version in `download.yaml` has the same property: the pin is a declaration, not a guarantee, and the gap is silent in both directions.

## Why it matters more now

#910 adds `tests/test_metpo_version_pin.py`, which asserts `download.yaml` and the code agree on `METPO_VERSION`. That guard is real but it checks the *declaration*. Nothing checks that `data/raw` actually holds what was declared, so a green test suite is compatible with the pipeline reading a three-month-old file from a URL that no longer appears anywhere in the repo.

## Suggested

Record the source URL alongside each downloaded artifact — a sidecar, or an entry in a manifest under `data/raw` — and have `kg download` re-fetch when the recorded URL differs from the declared one, rather than treating existence as sufficient. That makes a pin change take effect on the next ordinary run, and makes "what produced this file" answerable, which it currently is not.

A cheaper interim step: a check that reports when a tracked artifact's recorded URL is absent or stale, so the mismatch is at least visible.

Related: #900 (the pinning this exposed), #909 (what drifting silently cost), and the `atomic_io` rule in CLAUDE.md — same principle, that an artifact's existence should imply something about its content.

realmarcin:
Two details from the #910 review that narrow this and make the cost concrete.

[...]

PR #923, opened 2026-09-02
fix(metpo): stop asserting the obsolete observation class, and guard the rest (#909, #924, #925, #926, #927)

realmarcin:
Refs #909 — closes the half that has an answer, and makes the other half visible instead of silent.

## Both halves of the design were obsolete, not one

#909 reported `METPO:2000511` on 706,765 edges. Investigating it turned up the other half:

| term | role | where | upstream |
|---|---|---|---|
| `METPO:2000511` `has observation` | organism → assay predicate | 706,765 edges | **deprecated** |
| `METPO:1001000` `observation` | second category on assay nodes | 503 nodes | **deprecated** |

They were one deliberate design. The comment said so:

> Use METPO:2000511 (has observation); the assay nodes carry METPO:1001000 (observation) in their multi-category **so the predicate's existing METPO:1001000 range is satisfied without an upstream change**.

Upstream changed anyway, and retired both ends.

A scan of every node category across all transforms found exactly one METPO term used as a category anywhere — this one, and it is deprecated. My original scan in #909 only checked edge predicates and endpoints, which is why it was missed.

## The category half is fixed here

`ASSAY_CATEGORY` drops to plain `biolink:Procedure`. A range that no longer exists cannot be satisfied, so the METPO half was asserting a retired class on 503 nodes and buying nothing.

## The predicate half has no answer available

I read all **112 live METPO properties** in the pinned release: none covers organism-to-assay, and nothing mentions assay, test, observation or measurement. Biolink's nearest match, `has procedure`, is `is_a: node property` — a node attribute slot, not an association predicate, so it cannot be an edge predicate. `biolink:related_to` is live but says nothing, which is what got `has observation` retired in the first place.

Upstream **metpo#461** ("Model biochemical tests and their outcomes") is the open modelling issue. So `METPO:2000511` stays — recorded as debt rather than left invisible.

Worth noting for the eventual decision: `METPO:2000511` is **not** redundant with the outcome predicates. The two sets are completely disjoint — 706,765 pairs carry `has observation` and no outcome, 449,568 carry an outcome and no `has observation`, overlap **zero**. Substrate panels (API 50CH, biotype100, 20NE) go through the former; enzyme panels (API zym) through `shows activity of`. Dropping the edges would lose the substrate panels entirely.

## The guard is the durable part

`tests/test_no_deprecated_metpo_terms.py` checks live code and bacdive's edge output against the pinned `metpo.json`. `KNOWN_DEPRECATED` is an explicit allowlist and every entry must name a tracking issue — an exemption without one is just a silenced check.

Three terms are allowlisted: `METPO:2000511`, plus `METPO:2000054` and `METPO:2000508`, which metatraits assigns as predicates on growth-observation paths that do not currently fire but are live code.

**The scan reads code, not prose.** A raw text search flagged three things that cannot reach the graph — the comment explaining the removal, `gold.py`'s docstring citing the pattern as precedent, and the allowlist itself — so the check would have fired on its own fix. It now walks the AST and considers only non-docstring string literals.

**It skips loudly rather than passing quietly.** `data/raw/metpo.json` on disk declares zero deprecated terms, because #911 means the #900 pin change never refreshed the cache. Rather than pass against the wrong file, the test skips with the command to fix it.

## Canary

Against the real 2026-06-12 release:

```
deprecated terms in the release:                1,216
unallowlisted deprecated CURIEs in live code:   none
reintroducing METPO:1001000 into ASSAY_CATEGORY is caught:  True
```

## One thing backed out

Correcting `gold.py`'s docstring — which cites the `METPO:1001000|biolink:Procedure` pairing this PR removes — marked gold's output stale and failed `StaleOutputGuardTest`. A **docstring-only** edit changes the behaviour fingerprint, because `ast.dump` includes docstrings. #880 made formatting-only edits digest-identical; docstrings are behaviour-neutral for the same reason and were missed. Filed as **#922** and reverted here rather than force an expensive rebuild for prose. The digest fix bumps `FINGERPRINT_VERSION` and reports every transform stale once, so it wants its own change.


realmarcin:
Three review rounds, four findings, all filed and fixed. Each one was the guard being narrower than the defect that motivated it.

## Round 1

**#924 — the guard ran nowhere.** It read `data/raw/metpo.json`, and `/data/raw/**` is gitignored, so all three substantive checks skipped in CI — the one place a guard against reintroducing a retired term needs to run. They skipped locally too, because #911 means the #900 pin change never refreshed the cache. The scans do not need the ontology, only the set of deprecated CURIEs, so that is now vendored (1,216 entries, 17 KB) with `scripts/refresh_deprecated_metpo_curies.py` to regenerate it. A separate reconciliation against the ontology runs when it is present, and is now the only test that skips.

**#925 — the artifact scan missed the case it was written for.** It read `bacdive/edges.tsv` and inspected the predicate column. `METPO:1001000` was deprecated while sitting in the **category** of 503 assay nodes: different column, different file, never opened. It now sweeps predicates, node categories (splitting on the pipe, which is how `METPO:1001000` was carried) and edge endpoints across every transform.

Widening it makes it fail on the current artifacts, **correctly** — `bacdive/nodes.tsv` still carries `METPO:1001000` because the fix changes what future runs emit and the transform has not been rerun. So this branch has two known-failing local tests, this one and the pre-existing `test_ontologies_stubs`; both pass in CI, where `data/` is absent. The assertion message says which side is behind, so a reader does not have to work it out.

## Round 2

**#926 — the package reached into the test tree.** `metpo_liveness.py` located its vendored data with `parents[2] / "tests" / "resources"`. `kg_microbe` is distributable and `tests/` is not shipped with it, so installed anywhere but a source checkout it would have returned an empty set — and empty means "nothing is deprecated", the reassuring direction. The vendored list is test scaffolding, so it and its loader moved into the test; the package keeps `deprecated_metpo_terms()`, which reads the real ontology and is the right entry point at runtime.

## Round 3

**#927 — the vendored set could outlive its release.** Only the deep reconciliation checked the `# release:` header, and that needs the ontology, so it skipped everywhere. When `METPO_VERSION` next moved, the list would keep describing the old release and every scan would pass against it — clean, and blind to exactly the newly deprecated terms. The header is now checked against `METPO_VERSION` directly, which needs no ontology. Same shape as the `download.yaml` guard in `test_metpo_version_pin.py`: a file that cannot import the constant is checked against it instead.

**Verified by mutation:** bumping `METPO_VERSION` without refreshing fails with the refresh command in the message; reintroducing `METPO:1001000` into `ASSAY_CATEGORY` is caught by the code scan.

## Where #909 stands

The category half is fixed and guarded. The predicate half — `METPO:2000511` on 706,765 edges — has no live successor in either vocabulary and stays as allowlisted debt pending metpo#461, so **#909 remains open deliberately**; this PR references rather than closes it.

Issue #914, opened 2026-09-02
The merged-graph invariant check can suppress the tarball rewrite if it raises


realmarcin:
Found in the review of #913.

`check_merged_invariants` is called inside `_cleanup_merged_outputs`, between `_normalize_edges_tsv` and `_rewrite_tarball`:

```python
if edges_file.is_file():
    _normalize_edges_tsv(edges_file)
    check_merged_invariants(edges_file, output_dir)      # <- reporting

if dest.get("compression") == "tar.gz":
    if nodes_file.is_file() and edges_file.is_file():
        _rewrite_tarball(archive, [nodes_file, edges_file])   # <- delivery
```

and `_cleanup_merged_outputs` is itself wrapped by `load_and_merge` in a blanket handler that deliberately swallows everything:

```python
try:
    _cleanup_merged_outputs(yaml_file)
except Exception as exc:  # noqa: BLE001
    print(f"[merge] post-merge cleanup skipped: {exc}")
```

So an exception anywhere in the new check — an unexpected column type, a permissions problem writing the report, a bug in the streaming loop — skips `_rewrite_tarball`. The merge then reports success while leaving the normalized loose TSVs beside a **stale tarball**, and the only trace is one printed line. Consumers reading `merged-kg.tar.gz` get the previous graph, which is precisely the confusion `_warn_about_stale_siblings` exists to prevent (#828).

That is a bad trade. The check reports on data quality; it must not be able to affect whether the artifact is delivered.

Fix: give it its own `try`/`except` so a failure is logged and the merge proceeds. Moving the call after `_rewrite_tarball` is not equivalent — when the tarball was extracted for normalization the loose TSVs are unlinked immediately afterwards, so the edge file may be gone by then.

Worth noting the general shape, since the blanket handler makes it easy to reintroduce: anything added to `_cleanup_merged_outputs` after a delivery step inherits the power to cancel it.

Issue #884, opened 2026-08-26
kgm-freshness-check silently reports optimistic verdicts when kg_microbe is not importable


realmarcin:
## Summary

`kgm-freshness-check` gives **different answers depending on how it is launched**,
with nothing in the output saying so. Under bare `python` it silently loses both
the content fingerprint and every `DATA_INPUTS` declaration, and the resulting
verdicts are *more optimistic* than the truth — which is the exact failure mode
#812 was filed to eliminate.

Found while verifying #879's fix.

## Reproduction

Same script, same working tree, same commit (b2f1ad1c), back to back:

```
$ python      .claude/skills/kgm-freshness-check/kgm_freshness_check.py
$ poetry run python .claude/skills/kgm-freshness-check/kgm_freshness_check.py
```

| source | bare `python` | `poetry run` |
|---|---|---|
| gold | `STALE_VS_CODE` | **`FRESH`** |
| madin_etal | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |
| mediadive | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |
| metatraits | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |
| metatraits_gtdb | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |
| microbedecoder | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |
| ontologies_stubs | `STALE_VS_CODE` | **`STALE_VS_CODE_AND_DATA`** |

Summary line, bare: `STALE_VS_DATA 0`, `STALE_VS_CODE_AND_DATA 0`.
Summary line, poetry: `STALE_VS_CODE_AND_DATA 6`, `FRESH 1`.

**Six sources are stale against their curation data and the bare run does not say
so.** A `mappings/` correction lands, the operator runs the pre-flight, sees only
`STALE_VS_CODE`, and has no signal that the data moved — #812, reintroduced
through the tool built to detect it.

## Cause

Three helpers all swallow the import and degrade to "nothing declared":

- `_fingerprint_verdict` — `except ImportError: return None` → falls back to timestamps
- `_declared_data_inputs` — `except Exception: return ()` → no data staleness possible
- `_declared_transform_inputs` — `except Exception: return ()` → no `STALE_VS_UPSTREAM` possible

[...]

PR #928, opened 2026-09-02
fix(download): make a changed pin take effect on the next ordinary run (#911, #929, #930, #931)

### LINKED ISSUE #911: Changing a pinned URL in download.yaml has no effect until someone knows to invalidate the cache

realmarcin:
Closes #911

## Existence is the wrong cache key

`kghub_downloader.download_from_yaml` decides entirely on `outfile_path.exists()`:

```python
if outfile_path.exists():
    if ignore_cache:
        outfile_path.unlink()
    else:
        logging.info("Using cached version of {outfile_path")
        continue
```

It stores nothing about which URL produced the file. So editing a pinned version in `download.yaml` changes nothing on disk — the run reports success and the pipeline keeps consuming the old artifact, with the declared pin and the cached bytes silently disagreeing.

That is not hypothetical. #900 moved METPO from `refs/heads/main` to `refs/tags/2026-06-12`; `data/raw/metpo.json` is still the file fetched on 2026-07-27, three releases behind. #909's guard reads that file, so it skips — a check written to catch retired terms is disabled by a pin change that appeared to succeed.

The library is third-party, so the fix lives in our `download()` wrapper, which already pre-processes the config.

## What it does

Each run records `local_name -> url` in `data/raw/.download_manifest.json`. Before handing over, artifacts whose declared URL has moved are removed, so `kghub_downloader` fetches them again.

**An artifact with no record is left alone.** This is the design decision that matters. Every file predates the manifest, so treating "unknown" as "stale" would re-fetch the entire corpus on first run — including an hour of MediaDive crawling — to learn what is already on disk. Unknown provenance is not wrong provenance; records accumulate as things download.

## Canary

Against the real `download.yaml` and `data/raw`:

```
manifest exists? False   recorded: 0
files stale-by-URL right now: 0        <- landing this re-fetches nothing
entries whose artifact is on disk: 59 / 63
```

Simulating the #900 change end to end:

```
after the pin change, stale: ['metpo.json']
[download] re-fetching metpo.json: its recorded URL differs from the one declared in the config
file gone? True
```

So it is inert on arrival and correct on the case that motivated it.

## Deliberately inert where it would be wrong

- **snippet runs** (`-x`) write truncated files; recording a URL against one would claim it is the real artifact;
- **`-i` runs** invalidate everything already;
- **a corrupt manifest** is treated as no manifest. Provenance is an optimisation and must never be able to block a download, so an unparseable file degrades to today's behaviour rather than to an error.

## Tests

10, weighted toward the ways this could do damage rather than the happy path:

- an unrecorded artifact is never stale — the property that keeps first-run cost at zero;
- a tag-filtered run only considers its own entries, so `kg download -t ontologies` cannot delete a MediaDive artifact;
- dropping stale removes only the moved artifact, not its neighbours;
- only artifacts that exist are recorded, so a skipped or failed entry gains no provenance claim;
- a corrupt manifest never blocks a download;
- the manifest is written atomically and sorted, so a half-written one cannot claim provenance arbitrarily.

`tox` green apart from two known failures that also fail on master: `test_ontologies_stubs` (pre-existing) and `test_no_deprecated_metpo_terms::test_transform_outputs_carry_no_unexpected_deprecated_term`, which is the true positive from #923 — `bacdive/nodes.tsv` still carries `METPO:1001000` until the transform is rerun.

## Follow-on

This unblocks #909's guard: once `kg download -t ontologies` runs, `metpo.json` refreshes to the pinned release on its own and the deprecation checks stop skipping.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -36,8 +36,7 @@ poetry run kg merge -y merge.yaml
 make run-summary
 ```
 
-`kg download` skips existing files; `-i` invalidates every selected entry, so
-always combine it with one or more `-t` tags. MediaDive invalidation also clears
+`kg download` skips existing files, but a pin change now takes effect on its own: the URL behind each artifact is recorded in `data/raw/.download_manifest.json`, and a file whose declared URL has moved is re-fetched. A file with no record is left alone — unknown provenance is not wrong provenance (#911). `-i` still invalidates every selected entry, so always combine it with one or more `-t` tags. MediaDive invalidation also clears
 its response cache and can trigger an approximately one-hour crawl.
 
 The normal merge creates `data/merged/merged-kg.tar.gz` and removes the loose
--- a/kg_microbe/download.py
+++ b/kg_microbe/download.py
@@ -84,17 +84,23 @@ def download(
     finally:
         if effective_yaml != yaml_file:
             Path(effective_yaml).unlink(missing_ok=True)
-
-    # Snippet runs write deliberately truncated files; recording a URL against
-    # one would claim it is the real artifact.
-    #
-    # Reads the original config, not the filtered one: the filtered copy is a temp
-    # file already unlinked above. Pending-hosting entries are excluded explicitly
-    # instead — `_report_pending` tells the user to satisfy those by hand, so the
-    # artifact often exists without ever having been downloaded, and recording it
-    # against a REPLACE_ME_ placeholder claims a provenance it never had (#929).
-    if not snippet_only:
-        record(yaml_file, output_dir, tags, skip_names=[_local_name_of(e) for e in pending])
+        # Recorded even when the download raised. kghub-downloader aborts the
+        # whole run on the first bad URL, so waiting for success means one dead
+        # source among 63 leaves the manifest unwritten forever — and #911's fix
+        # silently does nothing (#930). Only files that exist are recorded, so a
+        # partial run records exactly what landed.
+        #
+        # Snippet runs write deliberately truncated files; recording a URL against
+        # one would claim it is the real artifact.
+        #
+        # Reads the original config, not the filtered one: the filtered copy is
+        # unlinked just above. Pending-hosting entries are excluded explicitly
+        # instead — `_report_pending` tells the user to satisfy those by hand, so
+        # the artifact often exists without ever having been downloaded, and
+        # recording it against a REPLACE_ME_ placeholder claims a provenance it
+        # never had (#929).
+        if not snippet_only:
+            record(yaml_file, output_dir, tags, skip_names=[_local_name_of(e) for e in pending])
 
     # Post-download: Trigger MediaDive bulk download if mediadive.json was downloaded
     if snippet_only:  # Skip bulk download in snippet mode
--- a/kg_microbe/download.py
+++ b/kg_microbe/download.py
@@ -7,6 +7,7 @@ from typing import List, Optional, Sequence, Tuple
 import yaml
 from kghub_downloader.download_utils import download_from_yaml
 
+from kg_microbe.utils.download_manifest import drop_stale, record
 from kg_microbe.utils.mediadive_bulk_download import download_mediadive_bulk
 
 # Tag (in download.yaml) of the entry that pulls the MediaDive media list. A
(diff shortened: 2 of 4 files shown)


realmarcin:
Three review rounds, three findings, all filed and fixed. Two of them were about the same thing from different angles: bookkeeping getting in the way of the work it describes.

## Round 1 — #929: a hand-placed artifact gained provenance it never earned

`record()` read the unfiltered config while `drop_stale()` read the filtered one, and the expression choosing between them —

```python
effective_yaml if effective_yaml == yaml_file else yaml_file
```

— returns `yaml_file` in both branches. It was written that way because the filtered temp file is unlinked before `record()` runs, but it hid that rather than saying it.

The consequence: `_report_pending` tells the user to satisfy pending-hosting sources **by hand** ("Copy these from a reference `data/raw_*` directory"), so the artifact often exists having never been downloaded. `record()` would attribute it to the `REPLACE_ME_` placeholder — not a URL, and not where the file came from. The manifest exists to answer "what produced this file"; a placeholder is worse than nothing, and `stale_by_url` already treats "no record" as "leave alone".

## Round 2 — #930: one dead URL meant the manifest was never written

`record()` sat after the `try`/`finally`, so an exception propagated past it. `kghub_downloader` aborts the whole run on the first bad URL — that is why `_without_pending_hosting` exists — so one dead source among 63 left the manifest unwritten even though the entries before it downloaded fine.

And it compounds: those artifacts then exist unrecorded, `stale_by_url` correctly leaves unrecorded files alone, and the next run hits the same dead URL. A repo with one persistently broken source would never build a manifest, and this PR would silently do nothing.

`record()` only writes entries whose file exists, so recording after a partial failure is *more* accurate, not less — it captures exactly what landed.

## Round 3 — #931: the fix for #930 could mask the error it recorded alongside

An exception raised in a `finally` **replaces** the one propagating through it:

```
caller sees: OSError: manifest write failed — masks it
original preserved only as __context__: RuntimeError
```

So a manifest write failing on a read-only directory would be reported instead of the upstream 404 that actually stopped the download. Same argument as #914: a reporting step must not describe whether the real work succeeded. `record()` now has its own handler.

**Both #930 and #931 are verified by mutation** — moving `record()` back out of the `finally` fails the partial-record test with "a partial run recorded nothing"; removing the handler fails the masking test because the caller sees `OSError` instead of the `RuntimeError` it matches.

## Also checked, clear

- **No duplicate `local_name`** across the 63 entries, so the manifest key is unambiguous.
- **No templated URLs and no API-based entries**, so the declared URL is the fetched URL and comparing it is sound.
- The tests fetch the module via `importlib` because `from .download import download` in `kg_microbe/__init__.py` shadows the submodule with the function — the same re-export behind #884.

Tests 10 → 15. `tox` green apart from the two failures that also fail on master.

Issue #932, opened 2026-09-02
LPSN references 4,233 deposit CURIEs it never writes a node for


realmarcin:
Found while sizing #918 against the merged `20260815` graph.

LPSN emits `lpsn:<record> --biolink:close_match--> kgmicrobe.strain:<deposit>` for every culture-collection deposit it extracts from `nomenclatural_type`, but writes no node row for the deposit. BacDive writes deposit nodes only for deposits **it** cited, so any deposit LPSN knows about and BacDive does not arrives in the merged graph as a KGX-synthesized stub:

```
kgmicrobe.strain stubs in merged/20260815:  4,233
edges touching them, by knowledge source:
    infores:lpsn      5,730
    (no other source)
e.g. kgmicrobe.strain:ATCC-10145-U, ATCC-12209, ATCC-12368, ATCC-14631-C
```

Every one is LPSN's. They are typed `biolink:NamedThing` with no label — the shape #892 objected to, in our own namespace.

This is the mirror of #894. That issue was "BacDive links to deposits but nothing links back"; this is "LPSN links to deposits that nobody declares". #894 fixed BacDive's half by emitting the link; the node itself is still only ever written by BacDive.

`_extract_strain_curies` already has everything needed — it has the CURIE and the cleaned deposit code that would serve as the label. Writing a node row beside the `close_match` edge would type them `biolink:OrganismTaxon` and give them a name, the same way #895 handles unresolvable taxa: if you reference it, declare it.

Worth checking at the same time whether the 6,586 `lpsn:` stubs in that graph have the same cause — LPSN record ids referenced by microbedecoder's `strain -> lpsn:` edges without LPSN writing a node for every record it is pointed at.

Related: #894 (BacDive's half), #895 (the same "declare what you reference" fix for taxa), #918 (the merged-graph check that would report this).

Issue #915, opened 2026-09-02
The invariant check resolves columns last-wins, contradicting the convention merge_kg already established


realmarcin:
Found in the review of #913.

`find_multi_parent_strains` builds its column index as a dict comprehension:

```python
index = {name: i for i, name in enumerate(header)}
```

which is **last-wins** on a repeated column name. `merge_kg.py` next door goes out of its way to do the opposite:

```python
def _first_index(header: List[str], name: str) -> Optional[int]:
    """Return first index of `name` in header, or None if absent."""
```

and uses it for `knowledge_source` / `primary_knowledge_source` in `_normalize_edges_tsv`. That helper exists because merged headers really can carry duplicate column names — `kgx.cli.cli_utils.merge` takes the column union across sources, which is the whole reason `_resolve_column_plan` and `EDGE_COLUMNS_TO_DROP` are there.

Concretely, on a header like `subject predicate object … subject`:

```
dict comprehension  ->  subject: 3
_first_index        ->  subject: 0
```

The check would then read column 3 as the subject, find nothing starting with `kgmicrobe.strain:`, and report a clean graph. **The failure is silent and in the reassuring direction**, which is the worst shape for a check whose whole job is to notice something.

In practice `_normalize_edges_tsv` runs immediately before and rewrites the file with a deduplicated canonical header, so a duplicate should not reach the check on the merge path. But `find_multi_parent_strains` is a public function that takes an arbitrary path — it is called directly in tests and is the natural thing to point at an arbitrary merged graph, including the historical ones under `data/merged/*/` that predate the normalization. Relying on a caller three steps away to have cleaned the header is not a property worth depending on when the module next door already has the right helper.

Fix: resolve with first-occurrence semantics, matching `_first_index`.

Issue #903, opened 2026-09-02
The deposit report is written non-atomically, so a missing or truncated file is indistinguishable from 'no conflicts'


realmarcin:
Found in the second review pass on #893.

`bacdive_strain_deposit_conflicts.tsv` is written with a plain `open(..., "w")`, outside the `with` block that owns `nodes.tsv` and `edges.tsv`:

```python
conflicts_file = os.path.join(self.output_dir, BACDIVE_DEPOSIT_CONFLICTS_FILE)
with open(conflicts_file, "w", newline="") as f:
    conflict_writer = csv.writer(f, delimiter="\t")
    conflict_writer.writerow(DEPOSIT_CONFLICT_HEADER)
    conflict_writer.writerows(deposit_conflict_rows(contested_deposits))
```

CLAUDE.md is explicit about this shape:

> Cache existence must imply completeness. Use `atomic_write` from `kg_microbe.utils.atomic_io`, including completion markers where the helper requires them; never publish a partially written cache path.

Here it does not. Three states are conflated:

| on disk | means |
|---|---|
| file absent | the run died before reaching it — **or** it was never run |
| header only | no deposit had contested claims — **or** the write was cut short |
| N rows | N contested deposits — **or** a truncated prefix of more |

Because `nodes.tsv` and `edges.tsv` are closed and complete by the time this runs, a crash in the gap leaves outputs that look finished next to a report that silently understates what was suppressed. That is the same silent-failure shape #892 was about: nothing errors, and a consumer reading an empty report concludes no deposit lost its parent when 437 did.

The window is small — only the `bacdive_media_links.txt` write sits between — but the fix is one helper call, and this report is precisely the artifact whose absence must not be read as good news.

Fix: write it through `atomic_write`. Note `bacdive_media_links.txt` immediately above has the same shape and predates this PR; worth converting in the same pass rather than leaving one of the two done. Related to #820, which is the same class for `lpsn/nodes.tsv`.

PR #933, opened 2026-09-02
feat(merge): report the nodes KGX invented for undeclared endpoints (#918, #934, #935, #936)

### LINKED ISSUE #918: Extend the merged-graph invariants to report edge targets with no node row
Follow-up from #895 and the invariants module added in #913.

#895's fix gives BacDive's unresolvable NCBITaxon parents a typed stub, so they stop arriving as untyped `biolink:NamedThing`. It is a per-transform fix and covers 18 of the 19 cases. **The remaining one is referenced only by metatraits**, which the BacDive pass cannot see.

Copying the pass into metatraits would work and would be wrong for the same reason #896 was filed: the property being checked — "every edge target has a node row" — is a property of the *merged* graph, and each transform can only see its own half. metatraits legitimately emits taxon edges whose node rows come from the ontologies transform; only after the merge is it knowable which targets nobody supplied.

`kg_microbe/merge_utils/invariants.py` is now the place for this. It already streams the merged edge file; adding a pass over the merged node file and reporting edge endpoints with no node row is a natural second invariant, and it would catch:

- the metatraits case above;
- any future source pointing at a taxon nobody writes;
- the same class for non-taxon endpoints, which nothing checks today.

Worth deciding whether it reports all unresolved endpoints or only ones in prefixes we expect to resolve — a merged graph legitimately contains CURIEs from ontologies we do not ingest whole, so an unfiltered version may be mostly noise. Sizing that is the first step.

Related: #895 (the per-transform half), #896 (the same argument for strain parents), #790 (the stub-node pattern being reused).

realmarcin:
Closes #918

## Sizing changed the question

#918 asked for a check on "edge targets with no node row", and asked for it to be sized first, since a merged graph legitimately references CURIEs from ontologies we do not ingest whole. Sized against `merged/20260815`:

```
merged nodes: 2,852,793
merged edges: 14,664,554
endpoint references with no node row: 0
```

**Zero.** KGX leaves no endpoint undeclared — it invents a node instead. So the question worth asking after a merge is not what is missing, but what arrived as an invention:

```
nodes typed biolink:NamedThing with no name: 60,990  across 15 prefixes
```

That is the shape #892 objected to — an id, a category worth nothing, no label.

## The split is what makes it usable

```
prefix                count   expected
IMG                  18,374   yes
GTDB                 15,609   yes
GOLD                 15,585   yes
lpsn                  6,586   no
kgmicrobe.strain      4,233   no
NCBITaxon               486   no
GO                       66   no
UBERON                   33   no
FOODON                    9   no
...
expected:    49,568
unexpected:  11,422  across 12 prefixes
```

`IMG`, `GOLD` and `GTDB` are cross-reference identifiers naming records in someone else's system. We assert edges to them deliberately and never ingest those systems as nodes, so a stub is the correct outcome. They are 81% of the total — without the split, the actionable 19% is buried under them, which is exactly the "mostly noise" risk #918 raised.

What is left is sources referencing things they should declare:

- **lpsn 6,586** — multi-source; `infores:faprotax` accounts for 21,164 of the touching edges;
- **kgmicrobe.strain 4,233** — our own namespace, **all** from LPSN alone. Filed as **#932**: LPSN emits `close_match` to deposit CURIEs it never writes a node for, the mirror of #894;
- **NCBITaxon 486** — the #895 family, at merge scale.

## Notes

Written atomically beside the strain-parent report. The nodes file is derived from the edge file's name so both reports describe one graph, and an absent nodes file is tolerated rather than fatal — the strain-parent check needs none, and letting the stub check break the check beside it is the coupling #914 was filed about.

The `kgmicrobe.*` prefix can never be allowlisted as expected, since a stub in our own namespace can never be someone else's system. There is a test asserting that.

## Tests

18 in the file, 8 new. The ones that matter are the discriminations:

- a `biolink:NamedThing` that **carries a name** is not a stub — flagging on category alone would report real nodes as inventions;
- cross-reference prefixes are marked expected while `kgmicrobe.strain` is not;
- rows lead with the largest prefix, because a report nobody can skim is a report nobody reads;
- an unexpected header is skipped, not crashed.

`tox` green apart from the two failures that also fail on master.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -126,7 +126,11 @@ the claims report, and the node carries a description saying so. See issues
   sources, so `merge_utils/invariants.py` re-checks after the merge and writes
   `merged_strain_parent_violations.tsv` beside the merged TSVs — empty on a
   clean run, because an absent report cannot be told from a check that never
-  ran. See issues #892 and #896.
+  ran. It also writes `merged_stub_nodes.tsv`: KGX invents a node for any
+  endpoint no source declared, so after a merge the question is not what is
+  missing but what arrived as an invention. Cross-reference prefixes we never
+  supply (IMG, GOLD, GTDB) are marked expected; everything else is a source
+  referencing something it should declare. See issues #892, #896 and #918.
 - Never assert a METPO term the pinned release has deprecated. A hardcoded CURIE
   keeps being emitted long after the ontology retires it — nothing errors, and
   `METPO:2000511` reached 706,765 edges that way. `tests/test_no_deprecated_metpo_terms.py`
--- a/kg_microbe/merge_utils/invariants.py
+++ b/kg_microbe/merge_utils/invariants.py
@@ -15,6 +15,8 @@ from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple
 
 from kg_microbe.transform_utils.constants import (
+    CATEGORY_COLUMN,
+    NAME_COLUMN,
     NCBITAXON_PREFIX,
     OBJECT_COLUMN,
     PREDICATE_COLUMN,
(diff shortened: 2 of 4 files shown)


realmarcin:
Three review rounds, three findings, all filed and fixed. Two of them are the same mistake I had already argued against elsewhere, which is worth noting rather than glossing.

## Round 1 — #934: I contradicted my own principle one PR later

The stub report was written only when there were stubs, while the strain-parent report beside it is unconditional — and #903 spells out why:

> an absent file would be indistinguishable from a clean run, and a clean run is the thing worth being able to demonstrate.

That argument does not stop applying to the report next to it. Two files behaving differently for no stated reason is worse than either choice consistently: a reader who learns one is always present will assume the same of the other.

The `nodes_file.is_file()` guard is separate and stays — **not being able to check is not a clean result**, so an absent nodes file writes no report rather than claiming zero stubs on a graph whose nodes were never read. Both behaviours are now pinned by tests.

## Round 2 — #935: a verdict with no reasoning

`EXPECTED_STUB_PREFIXES` carries a reason per prefix and the report only tested membership, printing `yes`. "Expected" is the claim in that file that most needs justifying — it separates the 49,568 rows that are fine from the 11,422 that are not — and a reader had to take it on faith or go read the source. The reason was already written and cost a column:

```
prefix             count   expected  note                                                examples
IMG                18374   yes       IMG genome and taxon identifiers, referenced from…  IMG:2140918011|…
GTDB               15609   yes       GTDB taxon strings, referenced as cross-references   GTDB:GB_GCA_000007325.1|…
GOLD               15585   yes       GOLD study, project and biosample identifiers        GOLD:Go0000002|…
lpsn                6586   no                                                             lpsn:1002|…
kgmicrobe.strain    4233   no                                                             kgmicrobe.strain:ATCC-10145-U|…
```

It also makes the dict's values load-bearing, so a prefix allowlisted without a justification is visibly odd rather than silently exempted — there is a test for that.

## Round 3 — #936: a correct path discarded, then guessed at

`check_merged_invariants` rebuilt the nodes path from the edge file's name, while the only caller already held it, built from the same base a few lines earlier.

The guess **fails silently**: `str.replace` leaves the name untouched when the pattern is absent, the derived path does not exist, and the `is_file()` guard turns that into "no stubs". Clean, and in the reassuring direction — the same failure mode #915 was filed about in this very module. Since `base` comes from the merge config's `filename`, a destination not following the `_edges`/`_nodes` convention would break the derivation while the merge itself was fine.

`nodes_file` is now an optional argument, the same shape as `output_dir`. The test drives both paths against a deliberately oddly-named nodes file, so the guess-and-the-hand-off are covered separately.

Tests 18 → 24. `tox` green apart from the two failures that also fail on master.

Issue #937, opened 2026-09-02
The first download after #911 cements pre-existing files against URLs they never came from


realmarcin:
Found while assembling the transform-rerun command.

#911 records `local_name -> url` after each run and re-fetches when the declared URL changes. Its stated safety property is that **unknown provenance is not wrong provenance** — `stale_by_url` deliberately leaves unrecorded files alone rather than re-fetching the whole corpus.

But `record()` does not hold to that. It writes the *currently declared* URL for every entry whose file exists, whether or not the file came from there. So on the first run, every pre-existing artifact is assigned a provenance nobody verified.

Demonstrated with today's actual state:

```
before: recorded={}  stale=[]
after first record(): {'metpo.json': '.../refs/tags/2026-06-12/metpo.json'}
now stale? []
```

`data/raw/metpo.json` was fetched on 2026-07-27 from `refs/heads/main`. #900 then moved the pin to `refs/tags/2026-06-12`. A plain `kg download` now records the **new** URL against the **old** file — and from then on the URLs match, so it is never re-fetched. The stale artifact is cemented permanently, and #909's deprecation guard, which reads that file, keeps skipping.

The two halves are inconsistent: `stale_by_url` refuses to judge an unrecorded file, then `record` confidently labels it.

## Fix

`record()` should only claim provenance it can support — the entries this run actually downloaded. `download_from_yaml` does not report which those were, but the wrapper can tell: the set of entries whose file did **not** exist before the call and does after. Anything already on disk beforehand stays unrecorded, which is exactly the "leave alone" the design already commits to for reads.

A cheaper variant, if that bookkeeping is unwanted: record only when `ignore_cache` is set, since those entries were definitely fetched.

## Until then

Any pin changed before the manifest exists must be cleared by hand. For the pending rerun that means:

```bash
rm -f data/raw/metpo.json data/raw/metpo.owl
poetry run kg download -t ontologies
```

which fetches only the two missing files rather than the 17 that `-i -t ontologies` would.

Related: #911 (the manifest), #900 (the pin this misses), #909 (the guard it keeps disabled).

realmarcin:
Concrete instance, and the blast radius, from the work in #941.

[...]

PR #941, opened 2026-09-04
fix(schema): fetch the file the pinned Biolink model imports (#939, #940)

### LINKED ISSUE #939: download.yaml never fetches attributes.yaml, so the pinned Biolink model cannot be resolved
Hit while running `kg transform -s ontologies`:

```
FileNotFoundError: [Errno 2] No such file or directory:
  .../data/raw/attributes.yaml
```

raised from `linkml_runtime`'s import resolution, under `bmt.Toolkit` → `prepare_kgx()`.

`biolink-model.yaml` declares:

```yaml
imports: ['linkml:types', 'attributes']
```

`linkml:types` resolves inside the linkml package, but `attributes` is a **sibling file** resolved relative to the model — so it must sit in `data/raw/` next to `biolink-model.yaml`. `download.yaml`'s `schema` tag fetches `biolink-model.yaml`, `predicate_mapping.yaml` and `kgx-format.md`, and never fetches it.

It has never been there: absent from `data/raw`, from every `data/raw_*` snapshot, and from the tarballed snapshot. So the pinned-local-model path has never actually resolved, and `prepare_kgx()`'s own error message — "Run `kg download -t schema` first" — cannot fix it, because that command does not fetch the missing file.

Verified the fix is exactly one file. With `attributes.yaml` placed beside the model:

```
Toolkit constructed OK against the local model + its attributes.yaml
subclass of -> domain: ontology class | range: ontology class
close match -> symmetric True
```

## It is version-sensitive

`attributes.yaml` differs between releases, so it has to be fetched from the same tag as the model:

```
v4.3.6:  12,175 bytes  md5 3ed0e9a0060fc62e4baa20e3c54cecaa
v4.4.2:  15,000 bytes  md5 4a3573df7667dbf4f566c97a1b28c8b2
```

That interacts with #940 — the model currently on disk is 4.3.6 while `download.yaml` declares the v4.4.2 tag — so which one to fetch depends on resolving that first. Fetching v4.4.2's `attributes.yaml` against a 4.3.6 model would pair mismatched halves of a schema.

## Fix

Add `attributes.yaml` to the `schema` tag, from the same tag as `biolink-model.yaml`, and extend `prepare_kgx()`'s existence check to include it so the failure names the real missing input rather than surfacing as a `linkml_runtime` traceback four frames deep.

Worth checking at the same time whether `linkml:types` ever needs vendoring, or whether the packaged linkml always supplies it.

### LINKED ISSUE #940: data/raw/biolink-model.yaml is Biolink 4.3.6, while download.yaml, CLAUDE.md and the docs all say 4.4.2
Found while diagnosing #939.

`download.yaml` fetches the model from the **v4.4.2** tag, CLAUDE.md says the repo is pinned to 4.4.2, and `docs/BIOLINK_4_4_2_REVALIDATION.md` records a revalidation against 4.4.2. The file on disk is **4.3.6**:

```
data/raw/biolink-model.yaml   455,362 bytes  md5 3f1c5f75c692d75cda899e89872fa99e   version: 4.3.6
upstream v4.4.2 tag           512,070 bytes  md5 b541038cd80a70fdf4bd539feabfc457   version: 4.4.2
```

Not a stale `version:` field upstream — the v4.4.2 tag's own file declares 4.4.2. These are different files. The difference is material: **535 slots and 332 classes in 4.4.2 against 523 and 328 in 4.3.6** — 12 slots and 4 classes that anything validating locally has never seen.

So every local BMT/KGX validation has run against 4.3.6 while the repo claims 4.4.2.

## The good news, checked rather than assumed

The four slots this repo's recent modelling decisions rest on are **byte-identical** between the two:

| slot | |
|---|---|
| `close match` | `symmetric: true`, `exact_mappings: [skos:closeMatch, SEMMEDDB:same_as]`, no domain/range |
| `subclass of` | `domain: ontology class`, `range: ontology class` |
| `has procedure` | `is_a: node property` |
| `related to` | `domain/range: named thing`, symmetric |

So #899's argument about `close_match` asserting identity, #909's finding that `has procedure` cannot be an edge predicate, and CLAUDE.md's `subclass_of` house-convention note (#834) all stand unchanged. Nothing shipped needs correcting. But that is luck, not process — the next decision read off this file might land in one of the 12 slots that differ.

## #937 has now cemented it

The manifest from #911 recorded, on its first run:

```
biolink-model.yaml -> https://raw.githubusercontent.com/biolink/biolink-model/v4.4.2/biolink-model.yaml
```

The file predates the manifest and never came from that URL, so #937's bootstrap flaw has now recorded a false provenance — and because the recorded URL matches the declared one, `stale_by_url` will never re-fetch it. The mismatch is now self-perpetuating. This is the concrete instance #937 predicted.

## Fixing it is not a one-line swap

CLAUDE.md is explicit:

> Updating the Biolink version requires changing `download.yaml`, the lock file, fixtures, and tests together.

Going to real 4.4.2 means re-checking `tests/resources/biolink-model-minimal.yaml`, the `bmt` version resolved by the lock, and whatever `docs/BIOLINK_4_4_2_REVALIDATION.md` asserted — which was presumably produced against 4.3.6 and may need redoing under the version it names.

The alternative — declaring 4.3.6 and correcting the docs — is cheaper and equally honest, but loses whatever motivated the 4.4.2 pin in the first place. Worth deciding deliberately rather than by whichever file gets edited next.

Blocks: #939 (which `attributes.yaml` to fetch depends on which model version is correct).

realmarcin:
## The bug

`data/raw/biolink-model.yaml` declares `imports: [linkml:types, attributes]`. linkml resolves the unprefixed `attributes` as a **sibling file**, so `data/raw/attributes.yaml` is a required half of the schema, not an extra. `download.yaml` never fetched it. An ignore-independent search finds it has never existed in `data/raw` or in any `data/raw_*` snapshot.

So the pinned-local-model path has never worked. Worse, `prepare_kgx`'s preflight checked only the model and the predicate map, passed, and let linkml fail later with a bare `FileNotFoundError: data/raw/attributes.yaml` — while the guidance it *did* emit told the caller to run `kg download -t schema`, the one command that could not supply the file.

## The second bug, found by fixing the first

`data/raw/biolink-model.yaml` was **Biolink 4.3.6** — byte-identical to upstream `v4.3.6` (`3f1c5f75…`, 455,362 bytes) and dated 2026-07-24, before 6be5f9301 (2026-08-21) bumped the pin to 4.4.2.

`kg download` skips files that already exist, so the bump never replaced it. **#911 could not rescue it either**: the first manifest run recorded the *declared* `v4.4.2` URL against the 4.3.6 file already on disk, so declared and recorded agreed and the file was never stale. That is [#937](https://github.com/Knowledge-Graph-Hub/kg-microbe/issues/937) happening for real, one day after it was filed — the skew was self-perpetuating.

`kg download -i -t schema` now puts both halves at 4.4.2 and makes the manifest record true.

## This is not a version change

4.4.2 is what `download.yaml`, `pyproject.toml`, `tests/resources/biolink-model-minimal.yaml`, the model-review skill and `docs/BIOLINK_4_4_2_REVALIDATION.md` have all named since August. Only the artifact moved — onto the version already declared. Nothing here touches the lock file, fixtures or the pin.

Checked, not assumed:

| | 4.3.6 | 4.4.2 |
|---|---|---|
| slots | 523 | 535 (**superset**, none removed) |
| classes | 328 | 332 (3 removed) |

The three removed classes are `anatomical entity to anatomical entity part of association`, `chemical or drug or treatment side effect disease or phenotypic feature association` and `gene to disease or phenotypic feature association`. An ignore-independent `grep -r` finds **no reference to any of them** anywhere in the tree. `subclass of`, `close match`, `has procedure` and `related to` — the slots #834, #899 and #909 rest on — are identical across both versions, so nothing already shipped needs revisiting.

`predicate_mapping.yaml` is byte-identical between the two versions, so it was never skewed.

## Changes

- `download.yaml` — declares `attributes.yaml` at the same pinned tag, with a comment saying why the two halves must move together.
- `kg_microbe/utils/biolink_model.py` — `sibling_imports()` reads the model's own `imports:` block, and `prepare_kgx` preflights it. A future Biolink release that adds a sibling import now fails **by name, with the fix**, instead of surfacing a bare path from inside linkml. Regex rather than `yaml.safe_load` because this runs at import time and the model is 512 KB.
- `tests/test_biolink_schema_pin.py` — 6 tests: every imported file is declared, all Biolink artifacts share one tag, all are tag-pinned, prefixed imports are ignored, the error names the missing file, and the guard reads the **real shipped model** rather than a fixture that would keep passing after upstream adds an import. Verified the first test fails against the pre-fix `download.yaml`.
- `CLAUDE.md` — names the third file, and records the trap: a pin bump alone does not refresh `data/raw`; confirm the file on disk actually moved.

## Verification

```
prepare_kgx() -> Toolkit loads 4.4.2
subclass of -> domain: ontology class | range: ontology class
close match / has procedure / related to  resolve
md5 biolink-model.yaml b541038c… attributes.yaml 4a3573df…  (both match upstream v4.4.2)
manifest now records all three schema files truthfully
```

Full suite: **988 passed, 5 skipped, 1 failed**. The failure is `test_transform_outputs_carry_no_unexpected_deprecated_term` — `METPO:1001000` in the stale `data/transformed/bacdive/nodes.tsv`, a pre-existing artifact from before the `ASSAY_CATEGORY` fix. The test's own message diagnoses it: *"the artifact predates the fix and the transform needs rerunning."* Unrelated to this PR, and cleared by the pending rerun that this PR unblocks.

`poetry check --lock` and `scripts/generate_merge_configs.py --check` both clean.

Closes #939, closes #940.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -176,9 +176,22 @@ environment variables, defaults, and risk warnings. Do not duplicate that
 inventory here.
 
 KGX/BMT must use the pinned local Biolink files downloaded to
-`data/raw/biolink-model.yaml` and `data/raw/predicate_mapping.yaml`. Updating the
+`data/raw/biolink-model.yaml`, `data/raw/predicate_mapping.yaml`, and
+`data/raw/attributes.yaml`. The last of these is not optional: the model
+declares `imports: [linkml:types, attributes]` and linkml resolves `attributes`
+as a sibling file, so without it the pinned model cannot be loaded at all. All
+three move together — half a schema at one version is worse than either version
+alone — and `tests/test_biolink_schema_pin.py` checks that `download.yaml`
+fetches every file the shipped model imports, at one pinned tag. Updating the
 Biolink version requires changing `download.yaml`, the lock file, fixtures, and
 tests together.
+
+A pin bump alone does not refresh `data/raw`. The 4.4.2 bump landed in August
+2026 and the 4.3.6 file already on disk was never replaced, so every run since
+validated against a version nothing declared; the download manifest then
+recorded the 4.4.2 URL for it, which made the skew self-perpetuating. After
+changing a pin, confirm the file on disk actually moved. See issues #937, #939
+and #940.
 
 METPO is pinned the same way. `METPO_VERSION` in `transform_utils/constants.py`
 is the single source of truth, and the ontology (`metpo.owl`, `metpo.json`) and
--- a/download.yaml
+++ b/download.yaml
@@ -597,6 +597,15 @@
   url: https://raw.githubusercontent.com/biolink/biolink-model/v4.4.2/predicate_mapping.yaml
   local_name: predicate_mapping.yaml
   tag: schema
+# biolink-model.yaml declares `imports: [linkml:types, attributes]`, and linkml
+# resolves `attributes` as a sibling file. Without it the pinned local model
+# cannot be loaded at all, so this is a required half of the schema, not an
+# extra (#939). It must move with biolink-model.yaml: pairing the two halves at
+# different versions is worse than either version alone.
+-
+  url: https://raw.githubusercontent.com/biolink/biolink-model/v4.4.2/attributes.yaml
+  local_name: attributes.yaml
+  tag: schema
 
 # Immutable METPO templates consumed by multiple trait transforms
 -
--- a/kg_microbe/utils/biolink_model.py
+++ b/kg_microbe/utils/biolink_model.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import os
-import re
 from collections.abc import Mapping
 from pathlib import Path
 from typing import Any
@@ -11,11 +10,6 @@ from typing import Any
 REPO_ROOT = Path(__file__).resolve().parents[2]
 DEFAULT_SCHEMA_PATH = REPO_ROOT / "data" / "raw" / "biolink-model.yaml"
 DEFAULT_PREDICATE_MAP_PATH = REPO_ROOT / "data" / "raw" / "predicate_mapping.yaml"
-
-#: Matches the top-level ``imports:`` block and its ``- item`` entries. Parsing
-#: the whole 512 KB model just to read one list would cost half a second on
-#: every call, and this runs at import time.
-_IMPORTS_BLOCK = re.compile(r"^imports:\s*$\n((?:\s*-\s*\S+\s*$\n?)+)", re.MULTILINE)
 
 
 def sibling_imports(schema_path: Path) -> list[Path]:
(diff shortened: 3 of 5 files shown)


realmarcin:
## Adversarial review round

Four findings, all filed as issues. Two were defects in this PR's own code and are fixed here; two are pre-existing and stay filed.

**Fixed in 53a23d67f:**

- **#942 — the new guard degraded silently.** `sibling_imports` matched the `imports:` block with a regex that only reads block style. `imports: [linkml:types, attributes]` — equally valid YAML, and the exact form both `CLAUDE.md` and the function's own docstring use to quote it — produced no match, so the function returned `[]`, the preflight went back to checking two files, and #939's failure returned as a bare path from inside linkml with nothing indicating the guard had stopped working. A silent degradation in the guard against silent degradation.

  The performance justification did not survive measurement: `prepare_kgx` has two production call sites, each hit once per run, and a real parse costs **~50 ms** with `CSafeLoader` against the **264 ms** BMT already spends building a `SchemaView` from the same file. Now parsed properly, with regression tests for flow style and for a model that imports nothing.

- **#945 —** a test comment still named v4.3.6 as the pinned version.

**Filed, not fixed here:**

- **#943 — a Biolink version change marks no transform stale.** `transform_fingerprint.py` does not reference the model, and no transform declares it in `DATA_INPUTS`, so this PR's 4.3.6→4.4.2 move invalidates nothing. A partial rerun could mix schema versions across sources in one merged graph. Moot for the rerun in progress, which rebuilds everything; a design change, not a fix to fold in here.
- **#944 — `BIOLINK_4_4_2_REVALIDATION.md` says it read the pinned local model**, which was 4.3.6 for the doc's entire life. The conclusions are verified unaffected (the four slots involved are identical across both versions), but the provenance sentence cannot be used to reproduce the review.

Also audited the general case of #937 across every recorded artifact — **0 of 30** predate the URL they are recorded against, so `biolink-model.yaml` was the only casualty. [Posted there](https://github.com/Knowledge-Graph-Hub/kg-microbe/issues/937#issuecomment-5533812340) with the method.

Re-ran `tests/test_biolink_schema_pin.py`, `test_biolink_hierarchy.py` and `test_cli_offline_startup.py` after the review commit: 38 passed. The full 988-test run was against the first commit; the second touches only `biolink_model.py` and its tests.

PR #946, opened 2026-09-04
Refresh the unified chemical mappings; vendored MIM SSSOM was 3 weeks stale

realmarcin:
Both mapping artifacts were last regenerated on **15 Aug**. In the meantime
MediaIngredientMech's curation continued, so the vendored MIM SSSOM had drifted
**50 rows** behind its declared source of truth — and every transform reading
`kgmicrobe_unified_entity_mappings.sssom.tsv.gz` was working from a stale view of
MIM's identities.

| artifact | before | after | Δ |
|---|---:|---:|---:|
| vendored MIM SSSOM | 2,945 | **2,995** | +50 |
| unified mappings | 608,161 | **609,973** | +1,812 |

Regenerated with `scripts/consolidate_chemical_mappings.py`. `sync_mim_sssom()`
pulled the sibling repo's `mappings/ingredient_mappings.sssom.tsv` at MIM `main`;
the vendored copy is now **byte-identical** to it (`diff -q`).

## What the run did beyond the sync

- Enriched **119,463** chemicals with **387,380** ChEBI synonyms via OAK.
- Swept stale MIM xrefs: 98 stale, 79 diverged; 9 more diverged after
  `complex_ingredients`.
- Harvested 128 synonyms onto 94 records from CHEBI xrefs; propagated 354
  cross-CURIE synonyms onto 171 records.
- Applied `kgm_retracted_groundings.tsv`: dropped 2, 13 already absent, 5
  name-level groundings removed across 3 objects, and **4 retracted names blocked
  from returning** through xref propagation.
- Emitted: xrefs=229,232, names=120,608, synonyms=259,977, parent_relations=142,
  hydrate_pairs=14. Nothing skipped as malformed, unknown-prefix, self or empty;
  one known-bad entity.

## Verification

- Both files round-trip-validate with the `sssom` package: **2,995 rows / 23
  prefixes** (vendored) and **609,973 rows / 60 prefixes** (unified).
- Vendored SSSOM byte-identical to MIM's source of truth.
- `gunzip -t` passes on the unified archive.
- Exactly the two intended files changed; nothing else in the tree.

## The net figure decomposed (review pass)

`+1,812` is **2,091 added − 279 removed**, so the removals were checked
individually rather than assumed benign. Of the 177 removals carrying a `MIM:`
subject:

| | count | verdict |
|---|---:|---|
| subject absent from MIM's current SSSOM | 98 | correct sweep (matches the reported "98 stale") |
| subject present, object changed | 78 | correct re-map (matches "79 diverged") |
| subject+object still in MIM | **1** | **a fix, see below** |

The remaining 102 are non-MIM (`chebi_xrefs`, KEGG primary mappings) and follow
the retraction/sweep counts the run reported.

The single flagged row was `MIM:Dry_Cow-manure skos:exactMatch ENVO:00003031` —
an **exact** match to *animal manure*, the generic parent. MIM's current SSSOM
says `skos:narrowMatch` to `ENVO:00003031` and `skos:exactMatch` to
`kgmicrobe.ingredient:dry_cow-manure`. The refresh replaced a
semantics-violating exactMatch with the correct identity; it only surfaced as
suspicious because the check keyed on subject+object while ignoring the
predicate. `ENVO:00003031` remains present on 3 rows.

**No removal is data loss.**

## Also in this PR: #947 fixed

The review of the data change surfaced a trap I had near-missed, so it is fixed
here rather than left filed:

- `MEDIAINGREDIENTMECH_ROOT` now locates the MIM checkout (sibling still the
  default). Filesystem adjacency was never a real contract — worktrees, CI
  checkouts and containers all silently took the stale path.
- An unreachable source of truth now **raises** instead of warning and exiting 0.
- The resolved path and its SHA-256 are logged, so a run records which MIM state
  built the artifact.
- `tests/test_mim_source_of_truth_resolution.py`: 7 tests pinning the contract.
  54 existing tests across the 4 affected suites still pass.

## Why the artifact moves +2 rows in the second commit

Re-running to verify the refactor reproduced its output revealed that it does not
— legitimately. `main()` seeds from the script's own prior output, so names a run
*creates* become seeds on the next run:

```
run 1: 609,973 rows   dd3affe74196bafe
run 2: 609,975 rows   5dd185415b62dd62   +2, -0
run 3: 609,975 rows   5dd185415b62dd62   fixed point
```

The 2 rows are `kgm.name:cas7365-82-4` → CHEBI:65994 and
`kgm.name:cas308066-66-2` → CHEBI:28757. It **converges** rather than growing
without bound, and this PR commits the fixed point, so the artifact matches the
code shipping with it. Property filed as #948 and documented in
`mappings/README.md`.

## #947 now fully closed, and a row-count correction

`--dry-run` added: it exports to a scratch path rather than skipping the write,
so the `sssom` round-trip validation still runs and the preview reports a real
delta — added and removed separately, with samples. Verified against the real
corpus: a preview run left both tracked artifacts byte-identical and reported
`+0` against the converged fixed point.

Also fixed a gap in the first #947 commit: `complex_ingredients.tsv.gz` still
resolved through `base_dir.parent / "MediaIngredientMech"`, so
`MEDIAINGREDIENTMECH_ROOT` never applied to it and it silently reported "not
present" wherever MIM is not literally adjacent. All three synced artifacts now
share one contract.

**Row counts in this description were inflated by one.** `grep -vc '^#'` counts
the header line, which carries no `#`. Corrected throughout: 608,161 → 609,975
(not 608,162 → 609,976). The triple-level figures were always right, since that
comparison excluded the header. The file has zero duplicate rows, so rows and
distinct triples coincide.

## Two things reviewers should know

**Missing legacy inputs are expected.** `mappings/chemical_mappings.tsv` and
`chebi_xrefs.tsv` are absent from this checkout. That is the documented
missing-legacy path in `mappings/README.md`: their rows are already carried by the
existing unified file and re-ingested by `load_existing_unified()`, so the loaders
skip rather than drop data. The +1,812 is a net figure, decomposed below.

**#723 re-reported, deliberately unaddressed.** The run again flagged 39 rows
where a raw upstream column outranks a corrected one on a candidate-order tie
(`Beef heart` → `UBERON:0000948` while the corrected column says
`FOODON:00004410`; `4-Aminobenzoic acid` → `CHEBI:194474` vs `CHEBI:30753`).
That issue records them as open curation questions. A mapping refresh is the
wrong place to decide them, so nothing was changed.

Built in a worktree off `master` — a live session is working in the main
checkout, and its HEAD was not moved.

https://claude.ai/code/session_01FyveZVuXAT9bytKrGKUcY6

Filed from this review: #947 (the consolidator exits 0 on stale data when the MIM sibling is unreachable — a trap I near-missed running from a worktree).


realmarcin:
## Adversarial review

Verified every quantitative claim against the actual bytes rather than re-reading the description. Six findings, all filed.

**The substance holds.** The removal audit reproduces **exactly** — independently recomputed from the shipped `.gz` against MIM's current SSSOM:

```
removed total        279
  MIM: subject       177
    subject gone      98   (PR says 98)
    object changed    78   (PR says 78)
    subj+obj present   1   (PR says 1)
  non-MIM subject    102   (PR says 102)
    flagged: ('MIM:Dry_Cow-manure', 'skos:exactMatch', 'ENVO:00003031')
```

MIM's SSSOM does say `narrowMatch` → `ENVO:00003031` and `exactMatch` → `kgmicrobe.ingredient:dry_cow-manure`; the unified file now carries the corrected exactMatch, and `ENVO:00003031` remains on exactly 3 rows. The `+50` MIM sync is right. All 16 new tests pass.

### Fixed here (2f31e6289)

- **#955 — the figures describe a run that was not shipped.** Actual: **+1,814** net from **2,093** added against 279 removed, 608,161 → **609,975**. The README said +1,812 / 2,091 / 609,973. Rows and distinct triples coincide exactly (zero duplicates either side), so this is not a header-counting artifact — the cause is the non-idempotence documented two paragraphs below it: the numbers are run 1's, the file is run 2, and "run 2 added 2 rows over run 1". Off by exactly that. The description's own correction note already had 609,975; the table and decomposition were never updated to match.
- **#956 —** `--dry-run` printed `Syncing ...` in the vendored-copy-missing branch while copying nothing, and the `_CBAI_SIBLING_RELPATH` comment claimed an extra `synonyms` column that does not exist (headers are byte-identical, and a divergent column could not survive a wholesale sync anyway).

### Filed, out of scope here

- **#952 — MIM's 142 `narrowMatch` rows never reach the unified artifact.** 0 of 142 subject+object pairs present under any predicate; `closeMatch` survives 92/92. Pre-existing (0 of 141 on `master`), so not a regression — but it makes "No removal is data loss" incomplete, and the `Dry_Cow-manure` case reads as fully fixed when the broader relation MIM curates is still dropped. Watch out for a coincidence: the unified file *does* hold 142 `narrowMatch` rows, and they are unrelated `parent_relations`.
- **#953 — the artifact is not reproducible.** `mapping_date` is stamped with the run date on **609,864 of 609,975** rows, and the gzip header carries a wall-clock mtime. Every refresh rewrites the full 13 MB blob whether or not a mapping changed. This is also why "run until the output hash stops changing" in the README cannot terminate as written — it works only on decompressed content, which is exactly where this PR's convergence and `+0` checks looked, and why they passed.
- **#954 —** `DryRunTests` is docstringed "must preview without touching a single tracked byte", but every test exercises `sync_mim_sssom()` alone. Nothing covers `main(["--dry-run"])` end to end or the `complex_ingredients` dry-run branch added in 477668568.

### One note for whoever runs this next

MIM has moved since this PR: `da3f9796` (5 Sep, "Republish the SSSOM") and `84881aa1` (4 Sep, "Curate trace element solution SL-10"). The vendored copy here is already one identity behind (2,995 vs 2,996). Not a defect — that is what vendoring costs — but it does show the drift this PR fixes reopening within a day, which is the argument for #953's reproducibility work before the next refresh.

realmarcin:
Correction to my review above: the #952 finding was wrong and I have [corrected it](https://github.com/Knowledge-Graph-Hub/kg-microbe/issues/952#issuecomment-5557762106). MIM's `narrowMatch` rows are **not** dropped — `MIM:<X>` is translated to the child's primary CURIE by design, and my pair check keyed on the untranslated subject, so it could only ever return zero. 129 of 142 are carried; 13 are genuinely absent, cause still unidentified.

That also means **this PR's `Dry_Cow-manure` analysis was right and complete**, and my qualification of it was not: `kgmicrobe.ingredient:dry_cow-manure skos:narrowMatch ENVO:00003031` is present in the artifact. #952 is retitled to the 13 rows that do remain unexplained. Nothing else in the review changes.

PR #957, opened 2026-09-06
Make the unified mappings artifact a pure function of its inputs (#953)

### LINKED ISSUE #953: The unified mappings artifact is not reproducible: every refresh rewrites 13 MB regardless of content
From the adversarial review of #946.

Two independent sources of churn mean re-running `scripts/consolidate_chemical_mappings.py` always produces a different 13 MB blob, even when not one mapping has changed.

**1. `mapping_date` is stamped with the run date, not the assertion date.**

| artifact | rows carrying the run date | of total |
|---|---:|---:|
| `master` (15 Aug run) | 608,033 `2026-08-15` | 608,161 |
| #946 (4 Sep run) | 609,864 `2026-09-04` | 609,975 |

99.98% of rows. The column therefore records when the consolidator last ran, not when the mapping was asserted, and it is the largest single source of diff noise in the file.

**2. The gzip header carries a wall-clock mtime.**

```
master: mtime field = 1786840810  (2026-08-16T00:40:10Z)
#946:   mtime field = 1788509130  (2026-09-04T08:05:30Z)
```

So two runs with byte-identical decompressed content still produce different archives, and git stores a fresh 13 MB blob each time.

## The guidance this breaks

`mappings/README.md` (as updated in #946) tells the operator:

> If you need the fixed point, run until the output hash stops changing.

Against the archive hash that never terminates — the mtime alone guarantees a new hash every run. It works only against the *decompressed* content, which the instruction does not say. The same gap explains why #946's convergence check ("run 3 was byte-identical to run 2") and its `+0` fixed-point check could both pass while the committed artifact churns: both compare at the triple or decompressed level, which is precisely where the churn is invisible.

## Fix

- Write the archive with `gzip.GzipFile(..., mtime=0)` — or `mtime` pinned to the mapping set version — so identical content yields an identical blob.
- Preserve each row's original `mapping_date` through the seed-and-reload cycle and stamp only genuinely new or changed rows. If a run timestamp is wanted, `mapping_set_version` in the header comment is the place for it, where it costs one line of diff instead of 609,864.
- Then say "run until the output hash stops changing" and have it mean something.

Related: #946, and the non-idempotence already documented in `mappings/README.md` (step 1 seeds from the script's own previous output).

realmarcin:
Three things leaked the environment into a 13 MB committed binary, so every refresh rewrote all of it whether or not a mapping changed.

## 1. `mapping_date` was the run date, not the assertion date

| artifact | rows carrying the run date | of total |
|---|---:|---:|
| before #946 | 608,033 `2026-08-15` | 608,161 |
| after #946 | 609,864 `2026-09-04` | 609,975 |

99.98% of rows restamped, on a refresh that changed 1,814 of them. Each row now keeps the date it was first published with, keyed on its `(subject, predicate, object)` triple; new triples get today.

A previous artifact that is absent or unreadable degrades to today for every row — the pre-#953 behaviour — because a refresh must not fail on account of the file it is replacing.

## 2. The header dates were also the clock

`mapping_set_version` and the header `mapping_date` are now the **newest row date in the set**, which is what a set version should mean and moves only when the data does. Checked that nothing depends on them bumping per build: `read_predicate_semantics` in `kg_microbe/utils/chemical_mapping_utils.py` deliberately avoids date thresholds and documents why.

## 3. The gzip header carried the clock — and the filename

`mtime` was wall-clock seconds. The **filename** was the one I would have missed: `GzipFile(path, ...)` fills in `FNAME` automatically, so the same content written to a scratch path and to the published path produced different bytes — which would have made `--dry-run` disagree with the apply for reasons unrelated to any mapping. The determinism test caught it, not a careful reading.

`open_deterministic_gzip` pins `mtime=0` and passes an explicit `fileobj` with `filename=""`. `FLG` goes `0x08` → `0x00`.

## Verification — on the real artifact, isolated

Ran with `MEDIAINGREDIENTMECH_ROOT` pointed away and `--allow-stale-vendored`, so the MIM refresh could not confound the result. (MIM has moved twice since #946; letting it sync would have mixed a data change into a reproducibility change.)

**Canary first.** `--dry-run` reported the content was already at its fixed point, so anything that moved afterwards was attributable to this change alone:

```
[dry-run] triples 609,975 -> 609,975  (+0)   added 0   removed 0
```

**Then two consecutive full runs:**

```
run 1  md5 2329aecd0d5c038cdc339c46abfd2d72
run 2  md5 2329aecd0d5c038cdc339c46abfd2d72     BYTE-IDENTICAL
```

And the artifact is unchanged where it should be:

```
gzip mtime   1788509130 -> 0        FLG 0x08 -> 0x00
mapping_date histogram               unchanged (609,864 rows still 2026-09-04, not today)
header                               mapping_set_version/mapping_date "2026-09-04", read from the data
all non-date columns                 byte-identical to the previous artifact (diff empty)
rows                                 609,975
sssom round-trip validation          passed, 60 prefixes
```

So `md5` on the archive is now a valid convergence check — which is what `mappings/README.md` already told operators to do, and could not work before. `mapping_tool` still carries the commit, so it moves with the code rather than with the clock; the README now says so explicitly.

## Tests

`tests/test_mappings_artifact_reproducibility.py`, 10 tests: date read-back keyed on the triple, gzipped sets, absent file, blank dates, a corrupt previous artifact degrading rather than failing; identical content → identical bytes, path independence, round-trip, and a guard on the stdlib default that documents the original defect. Plus a check that **the committed artifact itself** has a pinned mtime — it failed before the regeneration and passes now, so it is a real canary rather than a restatement of the writer.

Full suite: **1,330 passed**. Two failures — `test_transform_outputs_carry_no_unexpected_deprecated_term` (stale `data/transformed/bacdive/nodes.tsv`) and `test_every_referenced_curie_has_stub_node` (stale `ontologies_stubs` output) — **reproduce on master with these changes stashed**, and both are cleared by the pending transform rerun.

Closes #953.


realmarcin:
## Adversarial review

Four defects in this PR's own code, three found by probing it rather than reading it. All filed; the ones that belong here are fixed in 61223c5bc.

### Fixed here

**#958 — a malformed date poisons `mapping_set_version`, permanently.** `published_mapping_dates` accepted any non-empty string, and the header is `max()` over those:

```
prior row:  A:1  skos:exactMatch  B:1  not-a-date
max() ->    not-a-date        <- becomes mapping_set_version
```

Any non-digit string sorts above every `2026-…`. And this branch is what makes it stick: before it, every date was overwritten with the clock each run, so garbage could not survive. Now only `YYYY-MM-DD` is carried forward. Same function also resolved a repeated triple by **file order** — the earliest date wins now, since the field records when a mapping was first asserted.

**#959 — `open_deterministic_gzip` leaked the file handle** when the gzip stream or wrapper failed to construct. The helper exists precisely because it owns two handles instead of one. Also `published_path: Path = None` → `Optional[Path]`.

**#961 (half) — `mapping_tool` read `@unknown` under `--dry-run`.** The repo root came from the *output* path, and the dry-run exports to a temp dir, so `git rev-parse` failed there while the apply wrote a real SHA. A preview differing from the apply in a line no mapping had touched — the same class of defect as the gzip `FNAME` leak this PR already fixes. Now resolved from the script's own location.

**#960 — nothing proved the exporter used the dates it read.** The unit tests established `published_mapping_dates` was correct; the lookups could have been deleted and every one of them would still have passed. `tests/test_mapping_date_preservation.py` drives a real `export_unified_sssom` over a one-chemical consolidator — enough for an xref row, a canonical-name row and a synonym row — and asserts an existing triple keeps its date, a new one gets today, a first run dates everything today, `mapping_set_version` follows the newest row date (checked with a **future** date so a passing clock cannot fake it), a malformed date reaches neither rows nor header, identical inputs export identical bytes, and the tool SHA survives an export outside the repo.

### Still open

**#961 (other half)** — `rev-parse HEAD` names a commit whose script may differ from the one that ran. Observed here: regenerating from a modified tree and then committing stamps the header with the *parent* commit. Verified while checking the review fixes — a full re-run changed exactly one line:

```
< # mapping_tool: "...consolidate_chemical_mappings.py@27525a0f1c2b..."
> # mapping_tool: "...consolidate_chemical_mappings.py@d602e0a8585d..."
```

The comment above it calls this "Git SHA for reproducibility", which is the strongest claim you can make about a field that can be false. Needs a decision (`-dirty` suffix, or hash the script itself), so it stays open rather than being decided in passing here.

### Confirmed the review fixes move nothing

Every `mapping_date` in the artifact is already ISO and no triple repeats, so both parsing changes are no-ops on real data — as intended. The only line that moved on re-run is `mapping_tool`, which is #961.

Suite: **1,337 passed**, same two pre-existing failures (stale `bacdive/nodes.tsv`, stale `ontologies_stubs`), both still reproducing on master with this branch stashed.

PR #963, opened 2026-09-06
Raise the oaklib floor off the retired bbop-sqlite bucket (#962)

### LINKED ISSUE #962: oaklib 0.6.23 resolves sqlite:obo: against the retired bbop-sqlite S3 bucket (403); pin blocks the fix via llm ^0.11
## What happened

An LBNL network admin reported a developer laptop making hundreds of requests to
a retired BBOP endpoint from what looked like an old oaklib. The immediate source
was a sibling repo (CultureBotAI/culturebotai-claw#345), but this repo carries
the same exposure.

oaklib below 0.7.2 hardcodes the `sqlite:obo:` download URL as
`https://s3.amazonaws.com/bbop-sqlite/{prefix}.db.gz`. That bucket has been
retired (INCATools/semantic-sql#110, #112) and now answers **HTTP 403**. oaklib
0.7.2 moved the default to `https://semanticsql.berkeleybop.io` (oaklib#897),
which answers 200. There is no environment-variable escape hatch on old
versions — the URL is an f-string literal — so upgrading is the only fix.

This repo pins `oaklib = ">=0.5.25"` with **0.6.23** locked, the last pre-CDN
release.

## Exposure here

First-party transforms are mostly insulated: `kg_microbe/utils/ontology_utils.py`
builds SemSQL DBs from the OWL the repo ships (the #604 single-source rule; see
`_ensure_ncbitaxon_db`, which explicitly rejects `sqlite:obo:ncbitaxon`), and
`metatraits.py` / `ontologies_stubs_transform.py` call `get_adapter(f"sqlite:{db_path}")`
on local files. `ner_utils.py` uses the `llm:sqlite:` selector on a local path.

What the pin still governs: every `runoak -i sqlite:obo:...` and ad-hoc
`get_adapter("sqlite:obo:...")` run inside this venv — the normal way a
developer checks a term while working on a transform — resolves against the
dead bucket, and a warm `~/.data/oaklib` cache hides that until a cache miss.
The 0.6.23 lock is also what the reusable id↔label gate installs when it runs
from this repo's lock.

## Why the pin cannot simply move

oaklib >=0.7.3 requires `llm >=0.14,<1`. This repo pins `llm = "^0.11"`
(0.11.2 locked), so raising the oaklib floor alone is unresolvable. Nothing
first-party imports `llm`; it is only reached through OAK's `llm:` selector in
`kg_microbe/utils/ner_utils.py`, so the llm floor can move with it.

The existing `setuptools = "<81"` pin is compatible (oaklib wants `<82`) and
stays.

## Fix

- `oaklib = ">=0.7.3"`, `llm = ">=0.14,<1"`, with the rationale inline.
- Relock with the same Poetry that wrote the current lock (2.4.1, lock-version
  2.1) so the diff is the dependency change, not a format downgrade.
- Port MediaIngredientMech's `tests/test_oaklib_semsql_pin.py`, adapted to
  `[tool.poetry]` and `poetry.lock`, asserting the floor, the lock, and the
  resolved `SEMSQL_SQLITE_URL_BASE`.

realmarcin:
Closes #962.

## Why

oaklib below 0.7.2 hardcodes the `sqlite:obo:` download URL as `https://s3.amazonaws.com/bbop-sqlite/{prefix}.db.gz`. That bucket is retired (semantic-sql#110/#112) and answers **403**; 0.7.2 moved the default to `https://semanticsql.berkeleybop.io` (oaklib#897), which answers 200. This repo had `oaklib = ">=0.5.25"` with **0.6.23** locked — the last pre-CDN release. There is no env-var escape hatch on old versions (the URL is an f-string literal), so upgrading is the only fix. An LBNL admin noticed the traffic from a sibling repo (culturebotai-claw#345); this repo carried the same lock.

First-party transforms are mostly insulated — `ontology_utils.py` builds SemSQL from the OWL the repo ships (#604) and the other `get_adapter` calls take local `sqlite:{path}`s — but the pin still decides where every `runoak -i sqlite:obo:...` and ad-hoc lookup inside this venv goes, and a warm `~/.data/oaklib` cache hides the breakage until a cache miss.

## What

- `pyproject.toml`: `oaklib >=0.5.25` → `>=0.7.3`; `llm ^0.11` → `>=0.14,<1` because oaklib ≥0.7.3 requires it and nothing first-party imports `llm` directly (only OAK's `llm:` selector in `kg_microbe/utils/ner_utils.py`). `setuptools <81` untouched — oaklib wants `<82`, so it is compatible.
- `poetry.lock`: relocked with **Poetry 2.4.1** (the version that wrote the current lock, lock-version 2.1) so the diff is the dependency change, not a format downgrade. Moved: oaklib 0.6.23→0.7.4, llm 0.11.2→0.31.1, openai 0.28.1→2.54.0 (transitive via llm; nothing first-party imports openai). Added: condense-json, distro, jiter, puremagic, py-horned-owl, pyreadline3, regex, ruamel-yaml, sniffio. Removed: the aiohttp stack (aiohttp, aiosignal, aiohappyeyeballs, async-timeout, frozenlist, multidict, propcache, yarl), bcp47, funowl — none imported first-party. **Unchanged**: kgx 2.4.2, linkml 1.10.0, linkml-runtime 1.10.0, koza 2.1.1, biolink-model 4.4.2, semsql 0.4.0, pystow 0.8.14, setuptools 80.10.2.
- `tests/test_oaklib_semsql_pin.py`: ported from MediaIngredientMech, adapted to `[tool.poetry]` and `poetry.lock`, with a `tomli` fallback for 3.10. Asserts the declared floor, that the lock satisfies it, and that the resolved `SEMSQL_SQLITE_URL_BASE` is not the S3 host.

## Verification

- In the poetry venv after install: oaklib 0.7.4, `SEMSQL_SQLITE_URL_BASE == "https://semanticsql.berkeleybop.io"`, download kwargs use the `requests` backend with the `oaklib` User-Agent; `import oaklib`, `import pkg_resources`, and `from oaklib import get_adapter` all succeed.
- New test: 3 passed on the branch; 2 failed against `master`'s `pyproject.toml` + `poetry.lock` (floor and lock checks), as intended.
- `poetry check --lock` (Poetry 2.4.1, as CI): exit 0. `tox -e lint` and `tox -e codespell`: pass.
- Full suite: **1340 passed, 2 failed, 6 skipped**. The 2 failures (`test_no_deprecated_metpo_terms.py::test_transform_outputs_carry_no_unexpected_deprecated_term`, `test_ontologies_stubs.py::test_every_referenced_curie_has_stub_node`) reproduce identically after reinstalling `master`'s lock with oaklib 0.6.23 — pre-existing, about on-disk transform outputs, not oaklib.

## Not in this PR

- semsql 0.4.0's own `semsql download` still targets the bucket; nothing first-party calls it.
- The local poetry venv had drifted well ahead of the lock (linkml-runtime 1.11.1 vs 1.10.0 locked, and ~50 others); `poetry install` realigned it to the lock as a side effect.
## Review

Adversarial read-only pass over the diff:

- #964 — **left filed**: semsql 0.4.0 (still the latest release) hardcodes the same retired bucket in `semsql download`. Nothing first-party calls it (`ontology_utils.py` uses `semsql make`, which is local), and there is no upstream release to take, so the pin test here must not be read as covering semsql.
- Verified, no issue: the `llm:sqlite:` selector in `kg_microbe/utils/ner_utils.py` — the only first-party path that reaches `llm` — constructs against llm 0.31.1 (`LLMImplementation` wrapping `SqlImplementation`) and a label lookup through it succeeds. `tomli`, which the new test falls back to on 3.10, is in the lock's `main` group, so CI's `poetry install` provides it. No workflow, Dockerfile, `neo4j/requirements.txt`, or pre-commit hook installs oaklib outside the lock. Nothing first-party imports any package the relock removed (`aiohttp` stack, `funowl`, `bcp47`) or the one it moved a major version (`openai`).
- Not a repo issue: two stale venvs sit alongside the poetry one on the reporting machine (`venv/` with oaklib 0.5.25, `.venv-kgm/` with 0.6.23). Nothing in the repo references them, but any `runoak` from them still hits the bucket; that is a local cleanup, not a change here.

https://claude.ai/code/session_011Z358ozx6Un9XYF25MtetL

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,177 +13,6 @@ files = [
 ]
 
 [[package]]
-name = "aiohappyeyeballs"
-version = "2.6.2"
-description = "Happy Eyeballs for asyncio"
-optional = false
-python-versions = ">=3.10"
-groups = ["main"]
-files = [
-    {file = "aiohappyeyeballs-2.6.2-py3-none-any.whl", hash = "sha256:4708045e2d7a6c6bdf8aafa8ed39649eaf926a4543b54560659129e3365953c4"},
-    {file = "aiohappyeyeballs-2.6.2.tar.gz", hash = "sha256:e202810ee718bd01fc6ef49e8ea53d023d5cb6b581076d7925aa499fa55dbe64"},
-]
-
-[[package]]
-name = "aiohttp"
-version = "3.13.5"
-description = "Async http client/server framework (asyncio)"
-optional = false
-python-versions = ">=3.9"
-groups = ["main"]
-files = [
-    {file = "aiohttp-3.13.5-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:02222e7e233295f40e011c1b00e3b0bd451f22cf853a0304c3595633ee47da4b"},
-    {file = "aiohttp-3.13.5-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:bace460460ed20614fa6bc8cb09966c0b8517b8c58ad8046828c6078d25333b5"},
-    {file = "aiohttp-3.13.5-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:8f546a4dc1e6a5edbb9fd1fd6ad18134550e096a5a43f4ad74acfbd834fc6670"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.manylinux_2_28_aarch64.whl", hash = "sha256:c86969d012e51b8e415a8c6ce96f7857d6a87d6207303ab02d5d11ef0cad2274"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux2014_armv7l.manylinux_2_17_armv7l.manylinux_2_31_armv7l.whl", hash = "sha256:b6f6cd1560c5fa427e3b6074bb24d2c64e225afbb7165008903bd42e4e33e28a"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux2014_ppc64le.manylinux_2_17_ppc64le.manylinux_2_28_ppc64le.whl", hash = "sha256:636bc362f0c5bbc7372bc3ae49737f9e3030dbce469f0f422c8f38079780363d"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux2014_s390x.manylinux_2_17_s390x.manylinux_2_28_s390x.whl", hash = "sha256:6a7cbeb06d1070f1d14895eeeed4dac5913b22d7b456f2eb969f11f4b3993796"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_28_x86_64.whl", hash = "sha256:bca9ef7517fd7874a1a08970ae88f497bf5c984610caa0bf40bd7e8450852b95"},
-    {file = "aiohttp-3.13.5-cp310-cp310-manylinux_2_31_riscv64.manylinux_2_39_riscv64.whl", hash = "sha256:019a67772e034a0e6b9b17c13d0a8fe56ad9fb150fc724b7f3ffd3724288d9e5"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_aarch64.whl", hash = "sha256:f34ecee82858e41dd217734f0c41a532bd066bcaab636ad830f03a30b2a96f2a"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_armv7l.whl", hash = "sha256:4eac02d9af4813ee289cd63a361576da36dba57f5a1ab36377bc2600db0cbb73"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_ppc64le.whl", hash = "sha256:4beac52e9fe46d6abf98b0176a88154b742e878fdf209d2248e99fcdf73cd297"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_riscv64.whl", hash = "sha256:c180f480207a9b2475f2b8d8bd7204e47aec952d084b2a2be58a782ffcf96074"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_s390x.whl", hash = "sha256:2837fb92951564d6339cedae4a7231692aa9f73cbc4fb2e04263b96844e03b4e"},
-    {file = "aiohttp-3.13.5-cp310-cp310-musllinux_1_2_x86_64.whl", hash = "sha256:d9010032a0b9710f58012a1e9c222528763d860ba2ee1422c03473eab47703e7"},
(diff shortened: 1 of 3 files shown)


realmarcin:
## Adversarial review

Verified the claims independently rather than re-reading them. The PR holds up; one finding, fixed in d7952b4ee.

### Verified

**The fix works at runtime, not just in the lock.** Ran the real adapter path — the one every transform reaches through `ontology_utils.get_chebi_adapter`:

```
adapter: _LazyOntologyAdapter
CHEBI:15377 label: water
aliases: ['BOUND WATER', 'H(2)O', 'H2O']
ancestors: ['CHEBI:15377', 'CHEBI:15377', 'CHEBI:23367']
SEMSQL_SQLITE_URL_BASE = https://semanticsql.berkeleybop.io
```

**The `llm:` selector — the only first-party path to `llm` — constructs under 0.31.1** and resolves a label: `LLMImplementation`, `label('CHEBI:15377') == 'water'`.

**`funowl`'s removal is safe, and for a reason worth recording.** This looked like the real risk: `oaklib/implementations/__init__.py` and `selector.py` both do a *top-level* `from oaklib.implementations.funowl.funowl_implementation import ...`, and `funowl` is gone from the lock — which would break `import oaklib` on a fresh install. It does not, because in 0.7.4 that module is only *named* `funowl`; it imports `pyhornedowl`. That is what `py-horned-owl >=1.4.1` in the lock is for. Confirmed by blocking `funowl` from `sys.meta_path` and importing anyway:

```
import oaklib.get_adapter: OK without funowl
```

Worth flagging for anyone verifying this locally: `poetry install` does not prune, so `funowl`, `aiohttp` and `bcp47` are all still importable in an existing venv. Checking "is it gone?" by importing it gives the wrong answer.

**The lock diff is exactly what the description says** — enumerated mechanically: 3 versions changed (oaklib, llm, openai), 10 removed, 9 added. `kgx`, `linkml`, `linkml-runtime`, `koza`, `biolink-model`, `semsql`, `pystow` and `setuptools` all unchanged.

**Nothing first-party imports any removed or major-bumped package** — ignore-independent search across the tree for `openai`, `aiohttp`, `funowl`, `bcp47`, `yarl`, `multidict`: no hits. No `oaklib` reference in any workflow, Dockerfile, requirements or ini file, so the lock is the only thing deciding the version. No first-party `sqlite:obo:` usage at all, which confirms the insulation claim.

**The guard discriminates.** Against master's `pyproject.toml`: `'>=0.5.25'` → `(0, 5, 25)`, fails the `(0, 7, 3)` floor. `poetry check --lock` clean; `tomli` 2.4.1 is in the lock for the 3.10 fallback.

### Fixed here

**#965 — the floor regex matched only `>=`.** `oaklib = "^0.7.3"` would fail with *"no lower bound in '^0.7.3'"*, which is false — and this pyproject already uses caret pins for `kgx`, `biopython` and `requests-cache`, so it is a realistic edit. Now accepts `>=`, `^` and `~`, with a message naming them, plus a regression test over all three forms. The verdict was always safe; the message was not, and it is the part someone will read years from now.

### Noted, not filed

The new pin has no upper bound on a pre-1.0 dependency that just caused an incident. That matches what was there before (`>=0.5.25`) and the lock pins the resolved version, so nothing moves unreviewed — but `watchdog` is capped and `biolink-model` pinned exactly in the same file, so it is worth being a decision rather than an inheritance. Recorded in #965 rather than changed here.

#964 (semsql 0.4.0 still hardcodes the retired bucket) stays open and correctly scoped: nothing first-party calls `semsql download`.

Issue #955, opened 2026-09-06
mappings/README.md reports figures from a consolidator run that was not the one shipped


realmarcin:
From the adversarial review of #946. Fixing in that PR.

`mappings/README.md` (added there) says:

> A net row count hides the shape of a change: the #946 refresh was **+1,812** net, which was **2,091** added against 279 removed.

Measured against the artifact actually committed:

| | README / PR description | shipped artifact |
|---|---:|---:|
| rows before | 608,161 | 608,161 |
| rows after | 609,973 | **609,975** |
| net | +1,812 | **+1,814** |
| added | 2,091 | **2,093** |
| removed | 279 | 279 |

Rows and distinct `(subject, predicate, object)` triples coincide exactly — zero duplicates in either version — so the two levels cannot disagree and this is not a header-counting artifact.

**The cause is disclosed two paragraphs later in the same README:** step 1 seeds from the script's own previous output, so "run 2 added 2 rows over run 1 and run 3 was byte-identical to run 2". The figures were measured on run 1; the file that shipped is run 2. Every number is off by exactly the 2 rows that paragraph predicts.

The removal audit is unaffected and reproduces exactly — 279 = 177 MIM-subject (98 subject gone, 78 object changed, 1 flagged) + 102 non-MIM, and the flagged row is `MIM:Dry_Cow-manure skos:exactMatch ENVO:00003031`. Verified independently against the shipped artifact and MIM's current SSSOM. Only the added/net figures are stale.

It matters because those two numbers are the evidence for the claim that the removals were checked rather than assumed benign, and a reviewer who reaches for a calculator finds they do not reconcile with the file in front of them.

realmarcin:
Closing: fixed in `27525a0f1` (#946) — `mappings/README.md` was corrected to the figures measured on the shipped artifact (+1,814 net, 2,093 added, 279 removed). The 'Closes #A and #B' auto-close trap is why it stayed open.

PR #967, opened 2026-09-06
Put scripts/ in the lint gate, and make convergence visible (#948, #950)

### LINKED ISSUE #948: The unified mapping artifact is not a pure function of its inputs: the consolidator seeds from its own previous output
Found reviewing #946, by re-running the consolidator to confirm the refactor in
that PR reproduced its artifact. It did not — and the reason is structural, not a
bug in the refactor.

## Observation

Two consecutive runs, identical inputs (the vendored MIM SSSOM byte-identical
between them, MIM untouched), produced different unified artifacts:

```
run 1: 609,974 rows   sha256(decompressed)=dd3affe74196bafe
run 2: 609,976 rows   sha256(decompressed)=5dd185415b62dd62   +2 triples, -0
run 3: 609,976 rows   sha256(decompressed)=5dd185415b62dd62   fixed point
```

The two added rows are name-keyed entries derived from CAS numbers:

```
kgm.name:cas7365-82-4     -> CHEBI:65994
kgm.name:cas308066-66-2   -> CHEBI:28757
```

## Cause

`main()` seeds from the script's own prior output:

```python
if sssom_output_path.exists():
    consolidator.load_existing_unified(sssom_output_path)
```

So run *N*'s input includes run *N−1*'s output. Names that a run **creates**
(through ChEBI enrichment, xref harvesting, or cross-CURIE propagation) become
seeds on the next run, and can derive further rows. The artifact therefore
depends on how many times the script has been run, not only on the source data.

## Why this is worth recording rather than fixing blind

The seeding is load-bearing and documented — it is how the missing legacy
priority-1/2 inputs (`chemical_mappings.tsv`, `chebi_xrefs.tsv`) survive their
own absence, per `mappings/README.md`. Removing it would drop real data. So this
is not a defect to rip out.

But two consequences deserve to be explicit:

1. **A reviewer cannot verify the artifact by re-running.** The natural check —
   regenerate and confirm the diff is empty — fails legitimately, and looks
   exactly like non-determinism until you dig. I spent a cycle on this before
   identifying it.
2. **Committed output depends on run count.** Whoever regenerates decides,
   implicitly, whether the committed file is at the fixed point or one step
   short. Nothing surfaces which.

It does converge: run 3 was byte-identical to run 2, so the sequence terminates
quickly rather than accumulating without bound. #946 commits the fixed point.

## Suggested improvements

- Print whether the output changed from the seed, and by how many rows, so a run
  says "converged" or "added N rows; re-run to reach the fixed point."
- Optionally, iterate internally until stable, so one invocation always yields
  the fixed point and the artifact becomes reproducible from a given input set.
- Document the property (done for now in `mappings/README.md` on #946).

Filed rather than fixed in #946: that PR is a data refresh plus the #947 safety
fix, and changing the seeding contract wants its own review.

### LINKED ISSUE #950: scripts/ is excluded from every lint gate, including the 2,800-line consolidator that writes tracked artifacts
Noticed while fixing #947. Both quality gates in `tox.ini` cover `kg_microbe/`
and `tests/` only:

```ini
[testenv:lint]
commands = ruff check kg_microbe/ tests/

[testenv:codespell]
commands = codespell kg_microbe/ tests/ -S ...
```

`scripts/` is in neither, and `ruff format` covers the same two directories.

## Why it showed up

Working on #946 I changed roughly 150 lines of
`scripts/consolidate_chemical_mappings.py` and added one test file. CI failed
with **nine** ruff errors — every one of them in the test file. The script
changes were not examined at all.

That script is ~2,800 lines and is the writer for two tracked artifacts, one a
13 MB binary `.gz` that feeds every transform through
`kg_microbe.utils.chemical_mapping_utils`. It is arguably the highest-consequence
file in the repo per line, and it is the one file class nothing checks.

`scripts/` currently holds 40+ modules on similar footing.

## What this costs

- Style drifts silently, so the eventual cleanup is large and reviewable only as
  noise.
- Real lint classes that catch bugs — unused imports after a refactor, shadowed
  names, mutable default arguments, bare `except` — are not applied where the
  blast radius is largest. #925's shadowed loop variable was exactly this shape
  of defect.
- Contributors get inconsistent feedback: a one-line test change is linted
  harder than a rewrite of the artifact writer.

## Suggested approach

Adding `scripts/` to the existing commands will almost certainly produce a large
first error count, so a single flip would be disruptive. Two gentler options:

1. Extend the scope but grandfather the current state via
   `[tool.ruff.lint.per-file-ignores]`, so new code is checked and existing
   violations are recorded rather than blocking.
2. Add `scripts/` to `codespell` immediately — that is cheap and near-certainly
   clean — and stage `ruff check` behind a fixing pass.

Either way the goal is that a change to a script that rewrites tracked data gets
at least the scrutiny a change to a test file gets.

Not attempted in #946: that PR is a data refresh plus the #947 safety work, and
turning on a linter over 40+ unlinted modules is not something to bundle into it.

realmarcin:
## `scripts/` was in no quality gate (#950)

`tox -e lint` and `tox -e codespell` — the two things CI runs — covered `kg_microbe/` and `tests/` only. So the 2,800-line writer of two tracked artifacts, one a 13 MB binary every transform reads through `chemical_mapping_utils`, was the one file class nothing checked. **405 ruff findings** across 20 modules and 8,573 lines.

Three of them could bite, and they are the argument for the whole change:

**`scripts/validate_knowledge_sources.py:119` has been a SyntaxError on Python 3.10 and 3.11** — two of the three supported versions.

```
SyntaxError line 119: f-string: unmatched '['
```

It nests an f-string reusing the same quote character inside a replacement field: PEP 701 syntax, valid only on 3.12+. The file is unimportable and unrunnable on the repo's own minimum, and nothing noticed.

**`scripts/prego_validation/precision_assay_go.py` calls two functions it never imports.** `continuous_predicate` and `assert_non_empty` — a `NameError` on the first real line of work. Its three sibling scripts all do `from channel_compat import assert_non_empty, continuous_predicate`; this one was missed.

**`consolidate_chemical_mappings.py` annotates `Set[Tuple[str, str]]` with `Tuple` never imported.** Latent only because attribute annotations are not evaluated at runtime — confirmed by constructing the class — but it breaks `get_type_hints` and is an undefined name regardless.

### What was done with the rest

- **274 auto-fixed** (whitespace, f-strings without placeholders, import order).
- **Remaining bug-class findings fixed outright**: an unused `col` dict a later `dict(zip(...))` had superseded, an unused loop variable, an unused path, and three unchecked `zip()` lengths made explicit.
- **The rest grandfathered per rule** in `pyproject.toml` — style (`D*`, `E501`, `E701`, `E702`), bandit noise on local analysis scripts (`S108`, `S110`, `S311`, `S603`, `S607`), and `B023` closures that are all called inside their own iteration. Each with a reason, and an instruction not to add to the list. The gate protects against *new* findings today rather than after a cleanup that may never happen — which is what #950 asked for.
- **`ruff format` deliberately not extended** to `scripts/`. Over 8,573 lines it would be a diff reviewable only as noise; adopting it there is its own change, and `tox.ini` says so at the point where someone would otherwise assume an oversight.
- **`codespell-write` widened alongside the gate**, so the fixer and the checker scan the same tree. Its six findings were all `URE` in two exported CSV node dumps — excluded the same way the existing data-file exclusions are.

## The run now says which state its output is in (#948)

`main()` seeds from this script's own previous output, so run *N*'s input includes run *N−1*'s output, and a name one run derives can seed another row on the next. The seeding is **load-bearing** — it is how the absent legacy priority-1/2 inputs survive their own absence — so this reports the property rather than removing it.

A run now ends with one of:

```
  Converged: identical to the seed (609,975 triples). This is the fixed point.

  Not yet converged: N added, M removed (X -> Y). The seeding is intentional,
  but the output is one step ahead of its input -- re-run to reach the fixed
  point before committing.
```

That addresses both consequences the issue names: a reviewer regenerating to check a committed artifact previously saw a non-empty diff indistinguishable from non-determinism (the issue's author lost a cycle to exactly this), and nothing said whether what was committed was the fixed point or one step short.

**Not done:** iterating internally until stable. #948 marks it optional, and re-running the full pipeline in a loop changes the run contract rather than its reporting — worth its own decision.

## Verification

```
tox -e lint       OK   (kg_microbe/ tests/ scripts/)
tox -e codespell  OK   (same three)
--dry-run         triples 609,975 -> 609,975 (+0), added 0, removed 0
full run          Converged: identical to the seed (609,975 triples)
3.10 parse sweep  every file under scripts/ parses; none did before
```

Three tests for the reporter: a fixed point says so, a moved artifact reports both directions and the instruction to re-run, and a first run is not mistaken for a failure to converge.

Suite: **1,344 passed**. The two failures — `test_transform_outputs_carry_no_unexpected_deprecated_term` and `test_every_referenced_curie_has_stub_node` — are the known stale-transform-output pair, unrelated and cleared by the pending rerun.

The artifact itself is unchanged: this is not a data refresh, so the regenerated `.gz` was restored after using it to verify the convergence message.

Closes #948, closes #950.

### DIFF
--- a/mappings/README.md
+++ b/mappings/README.md
@@ -105,8 +105,16 @@ Note that step 1 seeds from the script's own previous output, so the artifact is
 not a pure function of its inputs: a run can derive names that the previous run
 created, and re-running immediately may add a handful of rows. This converges
 rather than growing without bound — in the #947 refresh, run 2 added 2 rows over
-run 1 and run 3 was byte-identical to run 2. If you need the fixed point, run
-until the output hash stops changing; `md5 mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz`
+run 1 and run 3 was byte-identical to run 2.
+
+**The run now tells you which state you are in** (#948). It ends with either
+`Converged: identical to the seed (N triples). This is the fixed point.` or
+`Not yet converged: N added, M removed ... re-run to reach the fixed point
+before committing.` Before that, regenerating to check a committed artifact
+produced a non-empty diff that looked exactly like non-determinism, and nothing
+said whether what had been committed was the fixed point or one step short.
+
+If you need the fixed point, run until the message says converged; `md5 mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz`
 is now a valid way to check that, and was not before (#953).
 
 Everything else in the artifact is a pure function of its inputs. Each row keeps
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,16 +114,28 @@ select = [
 "kg_microbe/query_utils/*" = ["S608"]  # DuckDB queries use database CURIEs, not user input
 # scripts/ entered the lint gate with 405 findings (#950). The three that could
 # bite were fixed outright -- two undefined names and an f-string that was a
-# SyntaxError on 3.10 and 3.11 -- and 274 more were auto-fixed. What is left is
-# style and bandit noise on local analysis scripts, grandfathered per rule so
-# the gate protects against new findings today rather than after a cleanup that
-# may never happen. Burn these down in place; do not add to the list.
-"scripts/*" = [
-    "B023",  # closures over loop variables, each called inside the same iteration
-    "D100", "D103", "D205", "D400", "D401", "D415",  # docstring style
-    "E501", "E701", "E702",  # long lines, compound statements
-    "S108", "S110", "S311", "S603", "S607",  # bandit: /tmp paths, subprocess, random
-]
+# SyntaxError on 3.10 and 3.11 -- and 274 more were auto-fixed. What remains is
+# style and bandit noise, grandfathered PER FILE rather than for the directory:
+# the other 18 scripts are already clean and stay strict, and so does every new
+# one (#968). Each entry disappears when its file is cleaned; do not add to the
+# list, and `tests/test_scripts_lint_debt.py` fails if an entry stops firing.
+"scripts/analyze_custom_metpo_mappings.py" = ["D401", "E501"]
+"scripts/analyze_gtdb_metatraits_overlap.py" = ["D401", "E501"]
+"scripts/consolidate_chemical_mappings.py" = ["B023", "D205", "D401", "E501", "S110", "S603", "S607"]
+"scripts/download_kegg_bulk.py" = ["D401"]
+"scripts/download_kegg_minimal.py" = ["D401"]
+"scripts/examples/visualization/create_1hop_full_labels.py" = ["D100", "D400", "D415", "E501"]
+"scripts/examples/visualization/create_1hop_subgraph.py" = ["D100", "D103", "D400", "D415", "E501"]
+"scripts/examples/visualization/create_2hop_full_labels.py" = ["D100", "D400", "D415"]
+"scripts/examples/visualization/create_full_label_visualizations.py" = ["D100", "D103", "D400", "D415", "E501"]
+"scripts/extract_metpo_proposals.py" = ["E501", "S603", "S607"]
+"scripts/mediadive_coverage_check.py" = ["E501"]
+"scripts/prego_validation/build_assay_gold.py" = ["S108"]
+"scripts/prego_validation/build_uniprot_gold.py" = ["D103", "E701", "E702", "S108"]
+"scripts/prego_validation/fold_enrichment_go.py" = ["E701", "E702", "S108"]
+"scripts/prego_validation/precision_assay_go.py" = ["D400", "E701", "E702", "S108"]
+"scripts/prego_validation/predicate_interaction.py" = ["D400", "E501", "S108", "S311"]
+"scripts/validate_knowledge_sources.py" = ["E501"]
 
 [tool.ruff.lint.mccabe]
 # Unlike Flake8, default to a complexity level of 10.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,6 +112,18 @@ select = [
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["S101"]  # Allow assert in test files
 "kg_microbe/query_utils/*" = ["S608"]  # DuckDB queries use database CURIEs, not user input
+# scripts/ entered the lint gate with 405 findings (#950). The three that could
+# bite were fixed outright -- two undefined names and an f-string that was a
+# SyntaxError on 3.10 and 3.11 -- and 274 more were auto-fixed. What is left is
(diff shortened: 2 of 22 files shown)


realmarcin:
## Adversarial review

Two findings in this PR's own work, both fixed in 6539f254d.

**#968 — the grandfathering covered the directory, not the debt.** `"scripts/*" = [15 rules]` reads like it freezes the current state. It does not: 17 of 35 scripts carry those findings, so the entry hands a blanket exemption to the **18 that are already clean**, and to every script added from now on. A new file could land with an undocumented public function, a 200-character line and a bare `except: pass` and pass the gate this PR exists to add. It also could never shrink — nothing distinguished a rule still needed from one fixed long ago.

Now listed **per file**: 17 entries, 51 `(file, rule)` pairs, each carrying only what its own file triggers. Clean scripts stay strict, new ones start strict, and an entry disappears when its file is cleaned.

`tests/test_scripts_lint_debt.py` keeps it honest — it fails on any exemption that no longer fires, on a `scripts/*` wildcard reappearing, and on an entry naming a file that is gone. It runs ruff with `--isolated`, because the repo config would otherwise suppress the very findings it asks about — the check is circular without it, and silently passes. Verified it catches a dead entry rather than merely passing: adding an unneeded `S311` to one file fails the guard; removing it passes.

**#969 — the preview read the 13 MB artifact twice.** `seed_triples` was read up front for the convergence report but only used on the apply path, while `_report_export_delta` re-read the same 609,975 rows to build its own `before`. Both share the one read now — faster, and it removes the possibility of the two reports disagreeing about what the seed was. That docstring also still carried the pre-#955 figures (+1,812 from 2,091); corrected.

### Verified in this pass

- **No dead entries.** All 15 rules fire, checked with `--isolated` so the config under test cannot mask the answer: `B023` 7, `D400` 9, `E701` 25, `E702` 21, `E501` 17, `S603`/`S607` 5 each, and so on.
- **The removed imports were genuinely unused.** The autofix dropped `import re` and `Set` from `analyze_custom_metpo_mappings.py`; neither appears anywhere in the file, and `F401`/`F811`/`F841` now report nothing across `scripts/`.
- **Every touched script still parses on 3.10** — the check that found the original SyntaxError, re-run over the whole diff.
- **The `channel_compat` import matches the sibling convention exactly** — same form and position as `fold_enrichment_envo.py`, `habitat_threshold_check.py` and `ubiquity_check.py`.
- **The glob really did cover subdirectories** (9 of the 17 debt files are nested), which is why the wildcard was as broad as it was.
- `tox -e lint` and `tox -e codespell` pass with `scripts/` included; dry-run still reports +0 triples; suite **1,347 passed**, same two pre-existing failures.

Issue #968, opened 2026-09-06
The scripts/ lint ignore is scoped to the directory, so 16 clean scripts and every future one are exempt too


realmarcin:
From the adversarial review of #967. Fixing there.

That PR puts `scripts/` in the lint gate and grandfathers the 118 findings it cannot fix outright, with:

```toml
"scripts/*" = ["B023", "D100", ..., "S607"]
```

The intent was to grandfather **the current debt**. What it actually does is exempt **the directory, permanently**:

- 19 of 35 scripts carry those findings. The other **16 are already clean**, and this hands them a blanket exemption from 15 rules they do not need.
- Every script added from now on inherits it. A new file can land with an undocumented public function, a 200-character line and a bare `except: pass`, and the gate will pass — in the directory the PR exists to protect.
- The list cannot shrink as debt is paid. Nothing distinguishes "still needed" from "long since fixed", so it will read as load-bearing forever and the burn-down has no scoreboard.

That is a weaker gate than #950 asked for, and weaker than the commit message claims.

**Fix:** scope the ignores per file, listing only the rules each file actually needs. 19 entries, verifiable mechanically, and each one disappears the moment its file is cleaned. New and already-clean scripts then start strict, which is the whole point of adding the directory to the gate.

Worth a guard too: a test asserting no entry in the list is dead — that every listed `(file, rule)` still fires — so the list cannot rot into permission nobody needs. It was checked by hand for this PR (all 15 rules fire; none dead), which is exactly the kind of check that stops being run.

realmarcin:
Fixed in #967 (6539f254d).

PR #970, opened 2026-09-07
Identify the exporter by its own hash, not the commit it sat on (#961)

### LINKED ISSUE #961: mapping_tool records a commit the running script may not match, and 'unknown' under --dry-run
From the adversarial review of #957. Fixing the `--dry-run` half there; the dirty-tree half needs a decision.

```python
# Git SHA for reproducibility in the mapping-set header.
git_sha = subprocess.check_output(
    ["git", "-C", str(sssom_output_path.parent.parent), "rev-parse", "HEAD"], ...
)
```

**1. The repo root is derived from the output path, so `--dry-run` gets `unknown`.**

The dry-run branch exports to a `tempfile.TemporaryDirectory()`, so `parent.parent` is somewhere under `/var/folders/...`, `git rev-parse` fails, and the `except` writes `mapping_tool: "...@unknown"`. The apply writes the real SHA. So the preview differs from the apply in a header line for a reason that has nothing to do with the mappings — the same class of defect as the gzip `FNAME` leak fixed in #957, where the archive depended on which path it was written to.

The provenance of the *script* has nothing to do with where its output lands. Deriving the root from `__file__` fixes it and is strictly more correct.

**2. `rev-parse HEAD` names a commit whose script may differ from the one that ran.**

There is no dirty check. Regenerating the artifact from a modified working tree and then committing it — the normal workflow, and exactly what #957 did — stamps the header with the *parent* commit, whose version of `consolidate_chemical_mappings.py` is not the version that produced the file. Observed while verifying #957: a re-run after committing changed exactly one line, and only that line:

```
< # mapping_tool: "...consolidate_chemical_mappings.py@27525a0f1c2b..."
> # mapping_tool: "...consolidate_chemical_mappings.py@d602e0a8585d..."
```

The comment calls this "Git SHA for reproducibility", which is the strongest possible claim to make about a field that can be false. It is the same failure #937 describes for the download manifest: recording a provenance the artifact does not have is worse than recording none, because it silences the question.

There is also a chicken-and-egg here that no SHA can escape — the run necessarily precedes the commit that contains its output — so the honest options are:

- append `-dirty` when `git status --porcelain` is non-empty, so a header produced from uncommitted work says so; and/or
- record the SHA-256 of `consolidate_chemical_mappings.py` itself, which is exact, needs no git, and is what the field is actually trying to convey.

The second is strictly better and would also make `mapping_tool` stable across commits that do not touch the script — removing the last field in the artifact that moves without a data change.

realmarcin:
Closes #961.

## The problem

`mapping_tool` in the unified SSSOM header recorded `git rev-parse HEAD`, under a comment reading "Git SHA for reproducibility". It names the commit the *checkout* was on, not the code that ran:

- The normal workflow — regenerate from a modified tree, then commit the artifact — stamps the **parent** commit, whose copy of `consolidate_chemical_mappings.py` is not the copy that produced the file. Observed while verifying #957: a re-run after committing changed exactly one line, that one.
- The run necessarily precedes the commit that contains its output, so no commit SHA escapes this on its own.
- It moved on every commit that touched anything, making it the last field in the artifact that changed without a data change — the thing #953 set out to remove.

Recording a provenance the artifact does not have is worse than recording none, because it silences the question (#937).

## The change

`mapping_tool` now carries `sha256:<digest>` of the exporter script itself:

```
# mapping_tool: "kg-microbe/scripts/consolidate_chemical_mappings.py@sha256:cad5fa1e41f6…"
```

Exact, checkable against the file on disk, needs no git, and stable across commits that leave the script alone. It covers that one script and not the modules it imports — `script_fingerprint`'s docstring states that boundary rather than letting the field imply a closure it does not have.

## Verification

Regenerated with `MEDIAINGREDIENTMECH_ROOT=/nonexistent-mim-root --allow-stale-vendored`, so a MIM data refresh could not confound the result.

- Dry-run canary first: `triples 609,975 -> 609,975 (+0)`, nothing written.
- Apply: `Converged: identical to the seed (609,975 triples). This is the fixed point.`
- `diff` of the uncompressed artifact against the committed copy: **one line**, the `mapping_tool` line.
- Its value equals `sha256(scripts/consolidate_chemical_mappings.py)` computed independently.
- gzip header still `mtime=0` with no `FNAME` (#957 holds).

Six tests in `ProvenanceHeaderTests`: the header equals the hash of the script that ran; the value is *not* a bare 40-hex commit (guards a revert to `rev-parse`); editing the script changes it; identical scripts at different paths agree; a missing script yields `unknown` rather than raising; and the export outside the repo still gets a real value.

Removing the `subprocess` call made the `S603`/`S607` exemptions for this file dead. `tests/test_scripts_lint_debt.py` failed on exactly that (#968 doing its job) and the entries are gone.


realmarcin:
## Adversarial review

Read-only pass over the diff, then five findings filed as issues. Three are addressed in `0cb92ab37`; two stay filed as decisions.

**Fixed here**

- **#971 — `mapping_tool` was carrying the tool *and* its version, joined by `@`.** `sssom_schema.MappingSet` declares `mapping_tool`, `mapping_tool_id` and `mapping_tool_version`; packed together, a spec-aware reader sees a tool named `...py@sha256:…`. Now two header lines. Done in this PR rather than later because the artifact is a 13 MB binary — splitting it afterwards costs a second permanent revision.
- **#972 — a content hash is exact but is not a lookup.** `git show <sha>:path` resolves a commit; nothing resolves a content hash, and it is not the git blob hash either (git hashes `blob <len>\0` + content), so `git cat-file` cannot find it. `mappings/README.md` now carries the loop that finds matching commits. Verified end to end: it recovers `84852d8ba`, the commit whose copy of the script produced the first regeneration in this PR.
- **#975 — the `OSError` fallback returned `"unknown"` silently**, leaving the only evidence inside the gzip it had just written — the same shape of silence #961 was about. It now says so on stdout; the test captures stdout rather than trusting the return value alone.

**Filed, not fixed**

- **#973 — nothing checks that the committed artifact was produced by the exporter currently in the tree.** #970 makes that check possible for the first time, but a test asserting it would fire on any edit to the 2,800-line script — a comment included — and the only way to green it is a ~5 minute regeneration adding another ~13 MB binary revision to history. That is a bad trade for edits that cannot change the output; the issue lays out three options.
- **#974 — the fingerprint covers the script, not the modules it imports.** A change in `read_predicate_semantics` can alter the artifact while `mapping_tool_version` stays byte-identical. Narrower than the commit SHA was, and the fix is not obvious — an import-closure hash re-introduces churn from unrelated edits.

**Verification after the review fixes**

Dry-run canary, then apply, both with `MEDIAINGREDIENTMECH_ROOT=/nonexistent-mim-root` so a MIM refresh could not confound the result. Canary: `+0` triples and SSSOM validation passing with the new header key. Apply: `Converged: identical to the seed (609,975 triples)`. Against the artifact as it stood on master the entire diff is the one provenance line becoming two, and `mapping_tool_version` equals `sha256(scripts/consolidate_chemical_mappings.py)` computed independently.

Issue #540, opened 2026-04-17
Parameterize the IN (...) list in get_media_composition() to remove SQL-injection shape


turbomam:
## Context

Surfaced in Copilot's 2026-04-16 review of #531.

`kg_microbe/query_utils/organism_queries.py::get_media_composition()` builds its SQL by interpolating a list of medium IDs directly into an `IN ('...')` clause. Upstream inputs are currently trusted, so this is not presently exploitable — but it is a brittle pattern that:

- becomes an injection vector the moment the function is reused against unvalidated input
- silently breaks if upstream data ever contains single quotes or other SQL metacharacters

## Suggested fix

Parameterize the query. Two DuckDB-friendly options:

- pass the list as a Python list bound to a `?` placeholder with `UNNEST` on the SQL side, or
- load the IDs into a temp table and `JOIN` against it.

## File involved

- `kg_microbe/query_utils/organism_queries.py` (`get_media_composition`)

## References

- [PR #531](https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/531)
- Copilot review at commit `1de973d`, 2026-04-16T23:15Z

PR #977, opened 2026-09-07
Refresh the unified chemical mappings from MIM ab6cc8c3, and make the consolidator reach a fixed point (#976)

### LINKED ISSUE #976: consolidate_chemical_mappings never reaches a fixed point: the cobalamin name retraction oscillates with a 2-cycle
## What happens

Running `scripts/consolidate_chemical_mappings.py` repeatedly on unchanged inputs does not converge. It alternates between two artifacts one triple apart:

| run | triples | sha256 (12) | `name-level groundings removed` |
|---|---:|---|---:|
| 1 (seed = published `ff53900f1`) | 610,249 | – | 4 |
| 2 | 610,248 | `5e4dbdd75ce6` | 5 |
| 3 | 610,249 | `fd4a1f1567a7` | 4 |
| 4 | 610,248 | `5e4dbdd75ce6` | 5 |

Runs 2 and 4 are byte-identical; so are 1 and 3 modulo the seed. The convergence report added by #967 says `Not yet converged` on every run, alternating `0 added, 1 removed` and `1 added, 0 removed`, and the README's instruction to "re-run to reach the fixed point before committing" cannot be satisfied.

The single flapping triple is

```
kgm.name:cobalamin  skos:closeMatch  CHEBI:28911   (cob(III)alamin)
```

which is exactly the grounding `mappings/kgm_retracted_groundings.tsv` retracts at name level (#784, #786): `stale_object=CHEBI:28911`, `now_asserted=CHEBI:30411`, `retract_names=Cobalamine|COBALAMIN|Cobalamin (B12)`.

Environment: lock-faithful `poetry install` (sssom 0.4.21, oaklib 0.7.4), `MEDIAINGREDIENTMECH_ROOT` pointed at MIM `ab6cc8c3` (SSSOM sha `18aca34ae013`, 2,999 rows). The inputs did not change between runs; only the seed did.

## Why

`_retract_names` registers a name in `self._retracted_names` only inside `if hit:` — i.e. only when the record *currently carries* the name and something was actually removed. `propagate_synonyms_via_xrefs` reads that set as its guard against handing a retracted name back across the `CHEBI:28911 ↔ CHEBI:30411` xref.

So the guard depends on the seed:

- **Seed carries the name** (even run): retraction hits, registers the guard, propagation is blocked → output lacks the name.
- **Seed lacks the name** (odd run): nothing to remove, `hit` is False, the guard is never registered, propagation copies `COBALAMIN` back from the xref-equivalent `CHEBI:30411` record → output carries the name again.

Each run's output is the opposite phase of its seed. The existing test `test_propagation_cannot_hand_a_retracted_name_back` only covers the seeded phase.

## Fix

Register the retraction whenever the row's `now_asserted` target carries the name (the existing orphan check), whether or not the stale record carries it at that moment. The retraction is a standing rule about the object, not a removal event. Regression test: a record that does *not* carry the retracted name, with an xref-equivalent record that does; after `apply_retractions` + `propagate_synonyms_via_xrefs` the name must still be absent.

Found while refreshing the vendored MIM SSSOM (2,995 → 2,999 rows); the fix rides in that PR so the committed artifact can be a real fixed point.

https://claude.ai/code/session_01FyveZVuXAT9bytKrGKUcY6

realmarcin:
Normal consolidation run (sync from the MediaIngredientMech sibling, no `--allow-stale-vendored`) picking up everything that landed in MIM since #946, plus the one-line fix without which the artifact cannot be committed at a fixed point.

## 1. Data refresh (`491a6f2ca`)

| file | before | after |
|---|---|---|
| `mappings/ingredient_mappings.sssom.tsv` | 2,995 rows, sha `410b19fa186b` | **2,999 rows, sha `18aca34ae013`** (MIM `ab6cc8c3`) |
| `mappings/culturebotai_reviewed_ingredients.tsv` | 2026-08-14 snapshot | synced from `UNIFIED_INGREDIENT_MAPPING.tsv` sha `832dcda804ef` (MIM#556) |
| `mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz` | 609,975 triples | **610,248** (+274 / −1) |

**Vendored MIM SSSOM, +8/−4 triples**, all identity moves MIM made deliberately:
- `MIM:Air-dried_Garden_Soil` — was `exactMatch ENVO:00002263`; now a registry record `kgmicrobe.ingredient:air-dried_garden_soil` with `narrowMatch ENVO:00002263` (MIM#544).
- `MIM:Trace_Element_Solution`, `…_SL-10`, `MIM:Zeikus_Trace_Element_Solution` — were `exactMatch NCIT:C896`; now registry records with `closeMatch NCIT:C896` (MIM `b892fd4a`).
- ~2,300 rows differ only in `source` / `mapping_date` / `other` / `validation_method`: MIM#549 restored CultureMech grounding evidence on 110 records, and #540 republished with the curation-history-derived `mapping_date`.

**Unified artifact.** The single removal is the retired garden-soil `exactMatch`. The 274 additions are 4 MIM subject rows, 256 `kgm.name:*` name rows, 7 `cas:*`, 3 `CHEBI:*` xrefs, 3 `kgmicrobe.ingredient:*`, 1 `kgmicrobe.compound:*`; all carry `mapping_date 2026-09-06`. Rows changed in place: 15,269 in `source` only (MIM's last-curator tag moved), 33 in `subject_label` casing (from the refreshed reviewed-ingredients labels), 18 in `mapping_date`. Every other row is byte-identical to the published artifact, which #957 made checkable.

## 2. The consolidator did not converge (`0b0ccc9ca`, closes #976)

The first four runs alternated between two artifacts one triple apart:

| run | triples | sha256 |
|---|---:|---|
| 1 | 610,249 | – |
| 2 | 610,248 | `5e4dbdd75ce6` |
| 3 | 610,249 | `fd4a1f1567a7` |
| 4 | 610,248 | `5e4dbdd75ce6` |

The flapping triple is `kgm.name:cobalamin closeMatch CHEBI:28911` — the exact grounding `kgm_retracted_groundings.tsv` retracts at name level (#784/#786). `_retract_names` registered the propagation guard only inside `if hit:`, so on a run whose seed lacked the name the guard was never set and `propagate_synonyms_via_xrefs` copied `COBALAMIN` back from `CHEBI:30411`. The fix registers the retraction as soon as the row's own `now_asserted` target is shown to carry the name, whether or not the stale record carries it right now. Regression test added; it fails with the fix reverted (checked).

After the fix:

```
run 5  sha256 b024b16224bd   Converged: identical to the seed (610,248 triples). This is the fixed point.
run 6  sha256 b024b16224bd   Converged: identical to the seed (610,248 triples). This is the fixed point.
```

Runs 4 and 6 have identical triple sets; the only header difference is `mapping_tool_version`, which is the exporter's own hash (#961) and moved because the script changed.

## Environment note

The shared checkout's `venv` is from 2024 (sssom 0.4.4, oaklib 0.5.25) and fails the consolidator's own SSSOM validation with `validate() got an unexpected keyword argument 'fail_on_error'`. Ran everything in a fresh `poetry install` from `poetry.lock` (sssom 0.4.21, oaklib 0.7.4, Python 3.10), which is what CI installs. `--dry-run` first: it reported `+274 / −1` with the same single removal before anything was written.

## Gates run locally

- `ruff check kg_microbe/ tests/ scripts/` — clean (`scripts/` is deliberately outside `ruff format`, see `tox.ini` / #950)
- `codespell` on the two changed source files — clean
- `pytest` on the retraction, reproducibility, source-of-truth, asymmetric-direction, dispatch/data-input, and negative-cache suites — 75 passed
- Full suite: running, will report on the PR.

https://claude.ai/code/session_01FyveZVuXAT9bytKrGKUcY6


realmarcin:
Full suite on this branch, lock-faithful env (sssom 0.4.21, oaklib 0.7.4, Python 3.10), `KG_MICROBE_TESTS_NO_NETWORK=1`:

```
1316 passed, 46 skipped, 27 warnings, 21 subtests passed in 134.90s
```

No failures. The two stale-transform failures #957 reported on its branch did not reproduce here.

realmarcin:
## Adversarial review

Read-only pass over both commits, then verification in a separate worktree with the lock-faithful venv. Three findings filed; one fixed here in `3f544f7f6`.

**The fix is right.** Registration now sits after the `_name_is_carried_by` orphan check and before the `hit` branch — exactly the condition #976 proposed. The new test fails on the old line (checked by stashing the script) and passes with it.

**Convergence claim reproduced under complete inputs:** dry-run on the branch against MIM `ab6cc8c3` → `610,248 -> 610,248 (+0)`. First attempt was invalid — the worktree had no `data/raw`, so `hydrate_pairs=0` and two retractions vanished (`-14`); with the raw inputs linked it converges. Worth knowing for anyone verifying in a fresh checkout.

**Artifact provenance:** `mapping_tool_version` equals the script's sha256, gzip `mtime=0`/no FNAME, +274/−1 triples as stated, all 274 additions dated 2026-09-06.

### Findings

- **#978 — 18 common rows changed `mapping_date` with nothing else about them changing.** All `skos:narrowMatch` parent relations. The parent-relation branch let MIM's re-stamped date win over the published one — the inverse of every other row class, pre-existing since #957. **Fixed here**, and regenerated **seeded from master's artifact** (seeding from this branch would only have preserved the re-stamped dates). Result vs master: 274 added, 1 removed, **0** date changes; vs `0b0ccc9ca`: identical triples, 18 dates restored; dry-run after: `+0`.
- **#979 — 33 `kgm.name` rows relabelled** (`'matamycin' → 'Matamycin'`, `'[Fe(edta)](-)' → 'Fe·EDTA'`, `'Pyridoxine HCl' → 'Pyridoxine . HCl'`). The exporter takes `sorted(synonyms)[0]`, so a newly imported capitalised variant wins under ASCII order. Deterministic, cosmetic, left filed. Two of the new MIM surface forms look like upstream regressions (a spurious ` . `, an unbalanced parenthesis).
- **#980 — the unit suite writes into `data/raw`.** After one `pytest` run the fresh worktree had `data/raw/nlp/{stopwords,output}/…` and `exclusion_branches.tsv`. `Transform.__init__` does `rmtree` + `mkdir` + writes `stopWords.txt` under `input_base_dir/nlp` on construction. Not this PR's doing; filed.

One correction to the comment above: the two stale-transform failures did not "not reproduce" — in an environment without `data/transformed` they **skip** (`stub transform output not generated yet`). Same suite here: 1,315 passed, 47 skipped, both among the skips.

The 15,291 `source`-column changes are MIM curator-provenance tags arriving with the refresh and look right.

PR #981, opened 2026-09-07
Rebuild prep: LPSN deposit nodes, atomic LPSN writes, honest output summary, schema in the fingerprint (#932, #820, #949, #943)

### LINKED ISSUE #820: lpsn/nodes.tsv is written non-atomically, so a crashed run leaves a truncated file consumers trust
Found by the external review of #817; pre-existing, not introduced there.

`lpsn.py` writes `data/transformed/lpsn/nodes.tsv` with a plain `open(..., "w")` — no `atomic_write`, no completion marker. A run that dies mid-write leaves a **truncated but non-empty** file.

That shape is worse than an absent file, because the fail-safes are all keyed on absence:

* `microbedecoder._lpsn_supplied_ids()` (#817) treats an empty set as "lpsn has not run" and emits no stubs. A *truncated* file yields a partial set instead, so the ids that got cut off look unsupplied and are stubbed — minting duplicate nodes for taxa the lpsn transform really does own.
* `bacdive` and `microbedecoder` both resolve LPSN synonyms through `resolve_accepted_records`, which reads the same tree; a partial file silently under-resolves.

Narrow window, and the repo already has the tool for it: `kg_microbe/utils/atomic_io.py:atomic_write` (temp file + `os.replace`) exists precisely because "a failed run leaves no truncated file for the next run to accept as complete", per CLAUDE.md. It is used for derived caches guarded by a bare `path.exists()`; transform outputs consumed by *other* transforms are the same hazard with a wider blast radius.

Worth auditing beyond lpsn: any transform output another transform reads should be written atomically. `ontologies/ncbitaxon_nodes.tsv` is the other obvious one — `gold` refuses to run without it and `bacdive` builds its label map from it, and `gold._trimmed_taxa()` already raises on an *empty* file for exactly this reason, which shows the failure mode was anticipated but only guarded at one call site.

### LINKED ISSUE #932: LPSN references 4,233 deposit CURIEs it never writes a node for

### LINKED ISSUE #943: A Biolink version change marks no transform stale, so a rerun can mix schema versions
From the review of #941.

`data/raw/biolink-model.yaml` is a real pipeline input: `biolink_hierarchy.py` loads it, and `prepare_kgx()` makes it the default schema for every `Toolkit()` KGX constructs, so category and predicate decisions in transform output depend on it.

It is not tracked by freshness. `kg_microbe/utils/transform_fingerprint.py` contains no reference to it, and no transform lists it in `DATA_INPUTS`. `code_fingerprint`, `upstream_fingerprint` and `data_fingerprint` between them cover source code, upstream transform outputs and declared curation files — not the schema.

So swapping the pinned model, as #941 just did from 4.3.6 to the declared 4.4.2, marks **nothing** stale. A partial rerun then leaves some sources' outputs built against one Biolink version and the rest against another, in one merged graph, with nothing recording that it happened.

This is the same shape as #922 — a real input the fingerprint cannot see — but pointing the other way: #922 is about spurious staleness, this is about staleness that never fires.

Not urgent right now, because the rerun currently in progress rebuilds every source. It matters the next time the pin moves and someone reruns only part of the pipeline.

**Options:**
- add `data/raw/biolink-model.yaml` and its sibling imports to a shared input set every transform's fingerprint includes, or
- record the loaded model version in each `source_fingerprint.json` and treat a change as stale, which is cheaper and gives the version a place in the artifact record.

Either way the schema version should be visible in the fingerprint, not implicit.

### LINKED ISSUE #949: _describe_output reports a successful multi-file transform as having written nothing, inverting the #813 guard
Observed on a real `kg transform -s ontologies` run that succeeded:

```
[transform] ontologies: done — wrote no nodes.tsv/edges.tsv in data/transformed/ontologies
```

It wrote **415 MB across 28 TSVs**, including `ncbitaxon_nodes.tsv` at 925,301 rows and `ncbitaxon_edges.tsv` at 925,267 — the extract `gold` and the stub transform both depend on.

`_describe_output` in `kg_microbe/transform.py` looks for exactly two filenames:

```python
for name in ("nodes.tsv", "edges.tsv"):
```

The ontologies transform writes `<ontology>_nodes.tsv` / `<ontology>_edges.tsv`, so neither literal name ever exists and the function falls through to its "wrote nothing" branch.

**This inverts the guard it was added for.** Its own docstring says it reports row counts "because the failure this guards against is a run that completes without producing anything (#813)". A run that produced 415 MB now prints the same line a run that produced nothing would, so the signal is not merely absent — it is actively misleading, and on the most expensive transform in the pipeline, where a wrong "produced nothing" is most likely to send someone re-running an hour of work.

**Affected sources**, by inspecting `data/transformed/`:

| source | files written | reported as |
|---|---|---|
| `ontologies` | 28 TSVs, 415 MB | wrote nothing |
| `ontologies_stubs` | 9 TSVs | wrote nothing |

The other 14 sources write plain `nodes.tsv`/`edges.tsv` and report correctly.

**Fix:** summarise whatever `*_nodes.tsv` / `*_edges.tsv` the output directory actually contains, falling back to the plain pair, and reserve the "wrote nothing" wording for a directory with no node or edge TSV at all. The count of files plus total rows would be more useful than two row counts for these sources anyway.

**Separate observation, not part of this issue:** `data/transformed/bakta/` and `data/transformed/kegg/` are entirely empty (0 bytes, no files). Those two do write plain `nodes.tsv`/`edges.tsv` when they produce anything, so their "wrote nothing" line would be accurate — but whether they *should* be empty is worth confirming during the pending full rerun.

realmarcin:
Closes #932. Closes #820. Closes #949. Closes #943.

Four small fixes from the triage sequence, all cheaper to land before the transform rebuild than to find in the merged graph afterwards. `lpsn` feeds `lpsn_api` and `microbedecoder`, so #932/#820 need to be in before it reruns.

## #932 — LPSN declares the deposits it references

LPSN emitted `lpsn:<record> --close_match--> kgmicrobe.strain:<deposit>` for every culture-collection deposit in `nomenclatural_type` and never a node row, so KGX invented a bare `biolink:NamedThing` with no label for each — 4,233 in the `20260815` graph, all LPSN's. Every referenced deposit now gets a node row of the same shape BacDive emits for the deposits *it* cites (`biolink:OrganismTaxon`, name = the code as written, `provided_by infores:lpsn`), so the merge collapses shared deposits onto one node. Buffered across records: one row per deposit, not per citation.

## #820 — LPSN output is written atomically

`nodes.tsv`/`edges.tsv` went through a plain `open()`. A run that died mid-write left a truncated file, and every downstream fail-safe keys on *absence*: `microbedecoder._lpsn_supplied_ids` read a partial set as "lpsn has run" and treated the cut-off ids as unsupplied. Both files now go through `atomic_write`; the test crashes the run on the second record and asserts no output and no `.partial` remains.

## #949 — the completion line counts what was actually written

`_describe_output` checked two literal filenames, so `kg transform -s ontologies` printed `wrote no nodes.tsv/edges.tsv` after writing 28 files / 415 MB — the inverse of the #813 guard on the most expensive transform. It now counts every `*nodes.tsv` / `*edges.tsv` (`14 nodes files: 1,234,567 rows; …`); the plain pair keeps its familiar line; nothing written still says so.

## #943 — the Biolink schema is in the fingerprint

`source_fingerprint.json` now records `schema: {version, digest}` over `biolink-model.yaml`, `attributes.yaml` and `predicate_mapping.yaml`, and `kgm-freshness-check` reports **`STALE_VS_SCHEMA`** (`built against Biolink 4.3.6, pinned model is now 4.4.2`) when a recorded schema differs from the pinned one. A marker written before the field existed is not judged on it — unknown provenance is not wrong provenance (#911) — so nothing flips spuriously today; verified by running the check on the current tree (`ontologies` stays FRESH, `STALE_VS_SCHEMA 0`).

## Verification

- `tests/test_lpsn_transform.py`: every cited deposit has a node row; one row per deposit; a mid-run crash leaves no output file.
- `tests/test_transform_output_summary.py` (new): plain pair, per-ontology files, nothing written, header-only.
- `tests/test_transform_fingerprint.py`: version parsed; same version / different bytes → different digest; no schema on disk → `None`; the marker carries it.
- `tests/test_transform_dispatch_and_data_inputs.py`: moved schema → `STALE_VS_SCHEMA`; matching → FRESH; absent field → not judged.
- `ruff check kg_microbe/ tests/ scripts/` clean.

### DIFF
--- a/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
+++ b/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
-"""Freshness check for KG-Microbe transform / merge outputs vs origin/master.
+"""
+Freshness check for KG-Microbe transform / merge outputs vs origin/master.
 
 For every transform directory under kg_microbe/transform_utils/<source>/,
 compare:
@@ -13,6 +14,7 @@ Also report:
 
 Per-source status codes:
   FRESH             — output mtime > latest code commit AND no local changes
+  STALE_VS_SCHEMA   — the recorded Biolink schema differs from the pinned one (#943)
   STALE_VS_CODE     — output mtime < latest code commit on origin/master
   LOCAL_CHANGES     — working tree diverges from origin/master for this dir
                      (in that case, "current relative to master" is not
@@ -38,11 +40,10 @@ from __future__ import annotations
 
 import argparse
 import json
-import os
+import re
 import subprocess
 import sys
-from dataclasses import dataclass, asdict, field
-import re
+from dataclasses import asdict, dataclass, field
 from pathlib import Path
 from typing import Iterable, Optional
 
@@ -329,6 +330,7 @@ def _fingerprint_verdict(source: str, code_dir: Path) -> Optional[tuple]:
             code_fingerprint,
             data_fingerprint,
             read_fingerprint,
+            schema_fingerprint,
             upstream_fingerprint,
         )
     except ImportError:
@@ -343,6 +345,17 @@ def _fingerprint_verdict(source: str, code_dir: Path) -> Optional[tuple]:
         upstream_stale = recorded.get("upstream") != upstream_fingerprint(
             TRANSFORMED_DIR, _declared_transform_inputs(source)
         )
+        # Only judged when the marker recorded a schema. A marker written before
+        # the field existed says nothing about the schema, and unknown provenance
+        # is not wrong provenance (#911, #943).
+        recorded_schema = recorded.get("schema")
+        if recorded_schema is not None and recorded_schema != schema_fingerprint(REPO):
+            current = schema_fingerprint(REPO) or {}
+            return (
+                "STALE_VS_SCHEMA",
+                f"built against Biolink {recorded_schema.get('version', '?')}, pinned model is now "
+                f"{current.get('version', 'absent')}; rerun `poetry run kg transform -s {source}`",
+            )
         if upstream_stale and not (code_stale or data_stale):
             upstreams = ", ".join(_declared_transform_inputs(source)) or "an upstream"
             return (
@@ -641,6 +654,7 @@ def main() -> int:
             "STALE_VS_CODE",
             "STALE_VS_DATA",
             "STALE_VS_CODE_AND_DATA",
+            "STALE_VS_SCHEMA",
             "LOCAL_CHANGES",
             "MISSING_OUTPUT",
             "NO_CODE",
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -184,7 +184,9 @@ three move together — half a schema at one version is worse than either versio
 alone — and `tests/test_biolink_schema_pin.py` checks that `download.yaml`
 fetches every file the shipped model imports, at one pinned tag. Updating the
 Biolink version requires changing `download.yaml`, the lock file, fixtures, and
-tests together.
+tests together. Each transform's `source_fingerprint.json` records the schema
+it was built against, and `kgm-freshness-check` reports `STALE_VS_SCHEMA` when
+the pinned model has moved since (#943).
 
 A pin bump alone does not refresh `data/raw`. The 4.4.2 bump landed in August
 2026 and the 4.3.6 file already on disk was never replaced, so every run since
--- a/kg_microbe/transform.py
+++ b/kg_microbe/transform.py
@@ -206,16 +206,27 @@ def _describe_output(transform_obj, source: str) -> str:
     out_dir = getattr(transform_obj, "output_dir", None)
     if out_dir is None:
         return f"no output_dir attribute on {source} transform"
+    # Every `*nodes.tsv` / `*edges.tsv`, not the two literal names: the
+    # ontologies transform writes `<ontology>_nodes.tsv` per ontology and the
+    # stub transform does the same, so the literal check reported a successful
+    # 415 MB, 28-file run as "wrote no nodes.tsv/edges.tsv" -- the inverse of
+    # what the guard is for (#949).
     parts = []
-    for name in ("nodes.tsv", "edges.tsv"):
-        path = Path(out_dir) / name
-        if not path.is_file():
+    for kind in ("nodes", "edges"):
+        files = sorted(Path(out_dir).glob(f"*{kind}.tsv"))
+        if not files:
             continue
-        try:
-            with path.open("r", encoding="utf-8") as handle:
-                rows = max(sum(1 for _ in handle) - 1, 0)
-        except OSError as exc:  # pragma: no cover - unreadable output is rare
-            parts.append(f"{name} unreadable ({exc})")
-            continue
-        parts.append(f"{name}: {rows:,} rows")
-    return "; ".join(parts) if parts else f"wrote no nodes.tsv/edges.tsv in {out_dir}"
+        counted = []
+        for path in files:
+            try:
+                with path.open("r", encoding="utf-8") as handle:
+                    counted.append((path.name, max(sum(1 for _ in handle) - 1, 0)))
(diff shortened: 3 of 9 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff, then four probes: how `_hash_files` treats paths, whether microbedecoder's `_lpsn_supplied_ids` now sees deposit rows, what KGX does when two sources disagree on a deposit's `name`, and remaining callers of `_extract_strain_curies`. Four findings filed; one fixed here in `5c672bddd`.

**Fixed**

- **#982 — the schema digest folded in the absolute checkout path.** `_hash_files` hashes `path.as_posix()` before content, by design for rename detection, but with `repo_root / rel` the same bytes under two checkouts hashed differently — a marker carried with its output directory would have read `STALE_VS_SCHEMA` for a schema that had not moved. Now hashed by repo-relative name + content; `test_the_same_schema_in_two_checkouts_is_one_schema` pins it.

**Filed, not fixed**

- **#983 — `data_fingerprint` has the same property**, pre-existing since #844. Changing its scheme invalidates every recorded `data` field, so it should ride with a `FINGERPRINT_VERSION` bump or land just before the full rerun — not in this PR.
- **#984 — LPSN and BacDive can disagree on a deposit's `name`.** Where the strings differ, KGX's `prepare_data_dict` (preserve=True) turns the single-valued `name` into a list. Not measurable until both transforms rerun; the check is written in the issue.
- **#985 — the paired atomic writes can still tear if the *second* `os.replace` fails.** The #820 case (exception in the body) is covered; a rename failure after the first file is published is not. No two-file transaction exists in the repo; recorded as a boundary.

**Checked and fine**

- `_lpsn_supplied_ids` filters on `LPSN_PREFIX`, so deposit rows do not change what microbedecoder treats as "lpsn-supplied" taxa.
- `_extract_strain_curies` keeps its contract (one test in `test_bacdive_deposit_parents.py` still uses it).
- The freshness check on the current tree: `ontologies` stays FRESH, `STALE_VS_SCHEMA 0` — no marker has a `schema` field yet, and unrecorded is not judged.
- The `*nodes.tsv` glob matches nothing but transform outputs in any current output directory (`unmapped_associations.tsv`, `bacdive_media_links.txt` and the MIREOT side-files do not match).

PR #987, opened 2026-09-08
ontologies_stubs: write the MIREOT module as RDF/XML, which ROBOT can read back (#986)

### LINKED ISSUE #986: ontologies_stubs: ROBOT cannot re-read the functional-syntax MIREOT module it just wrote, so the MICRO stubs fail on the current mapping set
Hit during the rebuild (`kg transform -s ontologies_stubs`, master `6566b0717`), after NCIT, mesh, BTO and PO had written:

```
INVALID ONTOLOGY FILE ERROR Could not load a valid ontology from file: micro_mireot_module.ofn
subprocess.CalledProcessError: ['/usr/local/bin/robot', 'convert', '--input', 'data/transformed/ontologies_stubs/micro_mireot_module.ofn', ... '-f', 'json'] returned non-zero exit status 1.
```

`_run_mireot_extract` writes the MIREOT module as OWL functional syntax (`.ofn`) and immediately asks ROBOT to convert it to OBO-JSON. ROBOT 1.9.6 (Java 20) writes the file but cannot parse it back: the `-vvv` log shows every OWLAPI parser being tried and rejecting it. The module now carries MICRO annotation literals with embedded newlines and `™` (`"Difco™ Nutrient Broth\n…pH 6.8 ± 0.2"`), pulled in by 20 MICRO terms the current mapping set references that the 2026-08-15 run did not (`MICRO_0000094`, `MICRO_0000456`, `MICRO_0000520`…).

**Control that isolates the format:** extract the same module to RDF/XML → `robot convert` to JSON succeeds (114 nodes / 103 edges). Convert *that* RDF/XML to `.ofn` with ROBOT and convert it back → the same `INVALID ONTOLOGY FILE ERROR`. So ROBOT's own functional-syntax output does not round-trip for this content; RDF/XML does.

Consequence: the stub transform is dead on the current inputs, and it gates `gold` (`TRANSFORM_INPUTS`) and the merge. The 2026-08-15 outputs on disk are from a term set that no longer matches the mappings.

Fix: use RDF/XML (`.owl`) as the intermediate. Nothing else reads the `.ofn` (checked `kg_microbe/`, `tests/`, `scripts/`, `.claude/skills/`).

realmarcin:
Closes #986.

## What broke

During the rebuild, `kg transform -s ontologies_stubs` died on the MICRO step after NCIT/mesh/BTO/PO had written:

```
INVALID ONTOLOGY FILE ERROR Could not load a valid ontology from file: micro_mireot_module.ofn
```

`_run_mireot_extract` writes the MIREOT module in OWL functional syntax and immediately asks ROBOT to convert it to OBO-JSON. ROBOT 1.9.6 writes the `.ofn` but cannot parse it back once the module carries MICRO annotation literals with embedded newlines and a `™` (`"Difco™ Nutrient Broth … pH 6.8 ± 0.2"`). Twenty MICRO terms the current mapping set references — and the 2026-08-15 run did not — brought those literals in. `gold` depends on this transform, and the merge on both.

## Why the fix is the format, not the terms

- Extract the same module to **RDF/XML** → `robot convert` to JSON succeeds (114 nodes / 103 edges).
- Convert that RDF/XML to `.ofn` **with ROBOT itself** → converting it back fails with the same error.

So ROBOT's own functional-syntax output does not round-trip for this content; RDF/XML does. The intermediate is now `{prefix}_mireot_module.owl`. Nothing else reads the `.ofn` (checked `kg_microbe/`, `tests/`, `scripts/`, `.claude/skills/`).

## Verification

- Real run through the same driver the rebuild uses: **MICRO wrote 90 nodes / 89 edges**; all five stub ontologies landed; the completion line reads `5 nodes files: 755 rows; 4 edges files: 841 rows` (the #949 fix showing its first real output).
- `test_mireot_intermediate_is_rdfxml_not_functional_syntax`: mocks ROBOT, asserts the extract writes `.owl`, the convert reads that same file, and the JSON path is returned. **Fails on the old line, passes on the new** (checked by stashing the source).
- Ruff clean.

Stale `micro_mireot_module.ofn` / `po_mireot_module.ofn` files from earlier runs are not removed by the transform; they are inert.

### DIFF
--- a/kg_microbe/transform_utils/ontologies_stubs/ontologies_stubs_transform.py
+++ b/kg_microbe/transform_utils/ontologies_stubs/ontologies_stubs_transform.py
@@ -481,7 +481,7 @@ class OntologiesStubsTransform(Transform):
         """
         Run ``robot extract --method MIREOT`` and convert the result to OBO Graph JSON.
 
-        Writes intermediate ``.ofn`` and ``.json`` files into ``self.output_dir``
+        Writes intermediate ``.owl`` (RDF/XML) and ``.json`` files into ``self.output_dir``
         so they're available for inspection if the module needs auditing.
         """
         robot_bin = shutil.which("robot")
@@ -499,7 +499,12 @@ class OntologiesStubsTransform(Transform):
 
         lower_terms_file = self.output_dir / f"{prefix.lower()}_mireot_lower.txt"
         lower_terms_file.write_text("\n".join(lower_iris) + "\n", encoding="utf-8")
-        module_ofn = self.output_dir / f"{prefix.lower()}_mireot_module.ofn"
+        # RDF/XML, not functional syntax. ROBOT 1.9.6 writes an `.ofn` it then
+        # cannot parse back once the module carries MICRO literals with
+        # embedded newlines and a `™` -- every OWLAPI parser rejects it and the
+        # convert step fails with INVALID ONTOLOGY FILE ERROR. RDF/XML round-
+        # trips the same content (#986).
+        module_owl = self.output_dir / f"{prefix.lower()}_mireot_module.owl"
         module_json = self.output_dir / f"{prefix.lower()}_mireot_module.json"
 
         extract_cmd: List[str] = [
@@ -514,7 +519,7 @@ class OntologiesStubsTransform(Transform):
         ]
         for iri in upper_iris:
             extract_cmd += ["--upper-term", iri]
-        extract_cmd += ["--output", str(module_ofn)]
+        extract_cmd += ["--output", str(module_owl)]
         subprocess.run(extract_cmd, check=True)  # noqa: S603
 
         subprocess.run(  # noqa: S603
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass, then three probes: other readers of the `.ofn` (none, across `kg_microbe/`, `tests/`, `scripts/`, docs and skills), what `tests/test_robot_atomic_output.py` actually guards, and what the transform leaves behind.

**Fixed here** — **#988**: the superseded `*_mireot_module.ofn` stayed beside the new `.owl`/`.json` in every existing output dir, with nothing saying which was current. Now unlinked when the `.owl` is written; the regression test plants one and asserts it is gone.

**Filed, not fixed** — **#989**: `_run_mireot_extract` calls ROBOT via `subprocess.run` straight into `output_dir`, bypassing `robot_utils`' atomic-output guarantee. Harmless today because every run re-extracts before reading, so a partial `.json` is overwritten rather than trusted; it becomes real the day a skip-if-present shortcut is added. Out of scope for a rebuild-unblocking fix.

**Checked**: the extract/convert argument order and the returned JSON path are unchanged; the mutation check (stash the source, run the new test) fails on the old line.

PR #990, opened 2026-09-08
bacdive: name the knowledge source as a CURIE on isolation-source and quality edges (#432)

### LINKED ISSUE #432: some bacdive edges have non-CURIE primary_knowledge_source values
## Summary

169,632 BacDive edges (10.2% of total) have non-CURIE values in the `primary_knowledge_source` field, violating KGX format requirements.

## Invalid Values

- `bacdive` - 169,130 edges (99.7%)
- `bacdive_mappings.tsv` - 502 edges (0.3%)

## Expected Format

According to [Biolink Model knowledge source guidelines](https://biolink.github.io/biolink-model/docs/knowledge_source.html), should use InfoRes CURIEs:

- `infores:bacdive`

## Impact

- Violates KGX format specification
- Will fail RDF serialization
- Breaks CURIE expansion
- Inconsistent with other kg-microbe data sources

realmarcin:
Closes #432.

## Measured on the rebuilt bacdive output (FRESH, 2026-09-08)

| `primary_knowledge_source` | edges |
|---|---:|
| `bacdive` (bare, not a CURIE) | **228,738** |
| … of which `biolink:location_of` / `RO:0001015` | 222,880 |
| … `METPO:2000067` | 5,056 |
| … `METPO:2000068` | 802 |

Every other edge in the file says `infores:bacdive`, or `['infores:bacdive', 'bacdive:<id>']` where a strain is an endpoint. Two sites wrote `self.source_name` instead of `self.knowledge_source` into the edge row: the isolation-source edge (`bacdive.py` ~3472) and the quality-aware PATO emit (~3370).

## Change

Both sites pass `self.knowledge_source`. The existing `StrainProvenanceWriter` then rewrites it to the list form where the organism is a strain, as it already does for the other edges (`tests/test_bacdive_provenance.py` covers that path for `location_of`).

## Guard

No fixture drives `BacDiveTransform.run()` end to end, so `tests/test_bacdive_edge_provenance_is_a_curie.py` parses the module and fails if any `edge_writer.writerow([...])` carries `self.source_name`; a second test pins that the guard actually finds edge rows. **Fails on the old file, passes on the new** (checked by swapping the file, not with `git stash`).

Rerunning `bacdive` (~11 min) after merge so the graph about to be merged does not carry 228,738 edges with a non-CURIE source.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -3367,7 +3367,7 @@ class BacDiveTransform(Transform):
                                         predicate_override,
                                         subject_id,
                                         predicate_override,
-                                        self.source_name,
+                                        self.knowledge_source,
                                         knowledge_level,
                                         agent_type,
                                     ]
@@ -3460,6 +3460,9 @@ class BacDiveTransform(Transform):
                                 )
                             )
                         # Write edge from the isolation source to organism
+                        # knowledge_source, not source_name: the bare "bacdive" that used
+                        # to land here is not a CURIE, and 228,738 location_of edges
+                        # shipped with it while every other edge said infores:bacdive (#432).
                         knowledge_level, agent_type = self._add_edge_metadata(
                             NCBI_TO_ISOLATION_SOURCE_EDGE, LOCATION_OF, organism_id
                         )
@@ -3469,7 +3472,7 @@ class BacDiveTransform(Transform):
                                 NCBI_TO_ISOLATION_SOURCE_EDGE,
                                 organism_id,
                                 LOCATION_OF,
-                                self.source_name,
+                                self.knowledge_source,
                                 knowledge_level,
                                 agent_type,
                             ]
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass over the two-line change and the guard. Probes:

- **Does the list-form rewrite now apply to these edges?** Yes, and it should: `StrainProvenanceWriter` rewrites `infores:bacdive` to `['infores:bacdive', 'bacdive:<id>']` when either endpoint is a strain, which is what `test_strain_object_with_infores_rewrites_to_list_form` already asserts for a `location_of` edge with a strain object. The isolation-source edge has the organism as its object, so it takes that path — consistent with every other edge on the same strain.
- **mediadive's byte-matching twin rows** (`mediadive.py:993-1000`) are medium edges, not isolation-source edges, so nothing there depends on the old bare value.
- **Guard scope**: the AST walk finds `edge_writer.writerow([...])` list literals only. Edges written through other names (`raw_edge_writer`, buffered `ar_edges_data_to_write`) are not inspected; those already pass `self.knowledge_source` today, but the guard would not notice a regression there. Narrow by construction; the real fix for the gap is a fixture-driven end-to-end run, which #859 already owns.

No new issues from this diff. Rerunning `bacdive` after merge.

Issue #997, opened 2026-09-08
kghub-downloader 0.5.0 changes download_from_yaml's signature; kg download would break, and the unit tests mock the call so CI stays green


realmarcin:
From the adversarial review of dependabot #966 (python-minor-patch group, 15 updates), verified in a fresh venv built from that PR's lock.

`kghub_downloader.download_utils.download_from_yaml`:

```
0.3.11: (yaml_file, output_dir, ignore_cache: bool = False, snippet_only: bool = False, tags=None, mirror=None) -> None
0.5.0:  (yaml_file, output_dir, download_options: Optional[DownloadOptions] = None, tags=None, mirror=None) -> DownloadReport
DownloadOptions(*, snippet_only=False, ignore_cache=False, progress=False, fail_on_error=True, verbose=False)
DownloadReport(downloaded, skipped, failed)
```

`kg_microbe/download.py:77` calls it with `snippet_only=` and `ignore_cache=` keyword arguments. Under 0.5.0 that is a `TypeError` on every `kg download`. CI on #966 is green because every test in `tests/test_download_tags.py` patches `download_from_yaml` (`:159, :199, :264, :277`) — the mock accepts any kwargs, so the one call that matters is never exercised against the real signature.

Two consequences:

1. **#966 cannot merge as-is.** The downloader call has to move to `DownloadOptions(...)` in the same change that raises the range.
2. **A mocked boundary needs one unmocked contract test.** The cheapest guard is a test that binds our kwargs against the *real* `inspect.signature(download_from_yaml)` — no network, no download — so a signature change fails at test time instead of at the next `kg download`.

Worth taking while there: 0.5.0's `fail_on_error=False` plus the returned `DownloadReport` is exactly the behaviour #938 wanted ("kghub_downloader aborts the whole run on the first failure … fetched nothing"), and the report gives the #911 manifest an authoritative list of what actually landed instead of inferring it from the filesystem.

realmarcin:
Fixed in #998: download.py builds a DownloadOptions, the floor is kghub-downloader >=0.5.0, and tests/test_download_contract.py binds the kwargs we pass against the installed signature (fails with TypeError on the old call shape).

Issue #938, opened 2026-09-02
PREGO's TLS certificate expired 2026-08-26, so the literature channel cannot be downloaded


realmarcin:
`kg download -t prego` fails:

```
ERROR:root:Failed to download: https://prego.hcmr.gr/download/literature.tar.gz
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED]
    certificate verify failed: certificate has expired
```

Not a local trust-store problem — the certificate really has lapsed:

```
subject = C=GR, ST=Attica, L=Anavyssos, O=Hellenic Centre for Marine Research, CN=prego.hcmr.gr
notAfter = Aug 26 09:21:57 2026 GMT
```

Seven days expired as of 2026-09-02.

## Impact is smaller than it looks

`data/raw/prego/` already holds `annotated_genomes_isolates.tar.gz` and `environmental_samples.tar.gz`, both fetched 2026-08-04/05 before the lapse. `PregoTransform.run` ingests every `*.tar.gz` in that directory and its docstring is explicit that "any subset works" — so the transform runs on two of three channels, which is what the current `data/transformed/prego/` was built from anyway. No regression, but the literature channel stays absent.

## It also blocks unrelated downloads

`kghub_downloader` aborts the whole run on the first failure, so `kg download -t schema -t mappings -t prego` fetched **nothing** — the PREGO entry was attempted at 2/13 and killed the run before `predicate_mapping.yaml`, `metpo_sheet.tsv` and `metpo-properties.tsv`. Any tag combination including `prego` is currently unusable.

Worth noting the manifest from #911 behaved correctly under this: 26 artifacts recorded despite the abort (thanks to #930's `finally`), and the three files that never arrived stayed unrecorded rather than gaining a false provenance.

## What not to do

Disabling certificate verification. It would work and it is not worth it — a source we re-download unattended is exactly the wrong place to stop authenticating the server.

## Options

1. **Wait.** Expired certs are usually renewed within days. Cheapest if the literature channel is not urgent.
2. **Mark it pending-hosting.** `download.yaml` already has the `REPLACE_ME_` mechanism for a source whose URL does not work: the entry keeps its tag and starts working when a real URL is pasted in, but is skipped at download time with a printed notice instead of aborting the run. That restores `-t prego` for the two working channels and unblocks any tag combination including it.
3. **Ask HCMR to renew**, and note the contact in the entry's comment.

[...]

PR #998, opened 2026-09-08
deps: python-minor-patch group without the Biolink bump; adapt download.py to kghub-downloader 0.5 (supersedes #966)

realmarcin:
Supersedes dependabot #966, which (a) bumped `biolink-model` to 4.4.3 against the standing decision to stay on 4.4.2, (b) would have broken every `kg download` (#997), and (c) conflicts on `poetry.lock` since #809/#996 merged.

## What moves

kghub-downloader 0.3.11→0.5.0, kgx 2.4.2→2.7.0, koza 2.1.1→2.6.2, curies 0.13.11→0.14.6, requests-cache, biopython, python-dotenv, duckdb, pytest, pre-commit, parameterized, notebook, ruff 0.15.14→0.16.6, boto3 — the #966 set. **`biolink-model` stays 4.4.2**: the schema pin is 4.4.2 in `download.yaml`, the fixtures, CLAUDE.md and every transform's `source_fingerprint.json` (#943); a package at 4.4.3 beside a data pin at 4.4.2 is the half-schema mismatch #939 was about.

## #997 — the downloader API changed and CI could not see it

`download_from_yaml` in 0.5.0 takes `download_options: DownloadOptions` and returns a `DownloadReport`; the `snippet_only`/`ignore_cache` keywords are gone. `kg_microbe/download.py` passed the old keywords → `TypeError` on every `kg download`, invisible because every unit test mocks the function. Fixed: the call builds `DownloadOptions`, the floor is `>=0.5.0`, and `tests/test_download_contract.py` binds the kwargs we actually pass against the **installed** signature. Verified in a venv built from this lock: the old call shape fails the contract test with `TypeError`; the new one passes (44 download tests green). `fail_on_error` stays default so a bad URL still aborts; #938 owns changing that.

## Lock

Updated with Poetry 2.4.3 scoped to the fifteen packages; `lock-version` unchanged. (A Poetry 1.8 rewrite of this lock silently downgrades it and strips the `groups` tables — see #996.)

## kgx 2.7 verification

Not just the suite: a real `kg merge -y merge.minimal.yaml` (bacdive + madin_etal + metatraits, 9.0M edges) under the new lock — result posted below when it finishes.


realmarcin:
kgx 2.7.0 verification, beyond the suite: `kg merge -y merge.minimal.yaml` (bacdive + madin_etal + metatraits) in a venv built from this lock — **363,214 nodes / 9,031,963 edges**, 690 s, exit 0. The merge invariants ran (21,858 invented endpoints across 14 prefixes, expected for a three-source merge with no ontologies), and `merge-cleanup` normalised the TSVs (`dropped=['id', 'knowledge_source']`) exactly as under 2.4.2. Full suite under the new lock: see below; `ruff check` with 0.16.6 is clean.

Issue #643, opened 2026-08-03
Biolink domain/range: has_phenotype with OrganismTaxon subject / ontology-class endpoints


realmarcin:
KGXVal biolink validation on the merged KG (see \`.claude/skills/kg-model-review/reviews/20260802_193254_merged.md\`) surfaces:

- 2 signatures with \`biolink:has_phenotype\` **subject** = \`organism taxon\`
- 2 signatures with \`biolink:has_phenotype\` **object** = \`ontology class|phenotypic quality\`
- 1 signature with subject = \`named thing\`
- 1 signature with object = \`biological process|ontology class|phenotypic quality\`

**Owner**: bacdive / metatraits (both emit \`NCBITaxon:X --biolink:has_phenotype--> METPO:Y\`)

Example:
\`\`\`
NCBITaxon:100    biolink:has_phenotype    METPO:1000443
NCBITaxon:100    biolink:has_phenotype    METPO:1000465
NCBITaxon:100    biolink:has_phenotype    METPO:1000457
\`\`\`

**Why biolink flags it:** \`has_phenotype\` domain is \`biolink:BiologicalEntity\`; \`OrganismTaxon\` is not under that (it's under \`OntologyClass\`). Historic KG-Microbe convention. Two options:
1. Widen: emit taxa as \`OrganismalEntity\` in addition to \`OrganismTaxon\` (\`biolink:OrganismTaxon|biolink:OrganismalEntity\` multi-cat).
2. Or use a predicate whose domain includes \`OrganismTaxon\` (a METPO-native predicate).

**Severity:** latent. Not release-blocking.

Surfaced during PR #639 post-merge review.

PR #999, opened 2026-09-08
Derive METPO predicates from the properties template's RANGE, not a column that no longer exists (#568)

### LINKED ISSUE #568: mapping_file_utils.load_metpo_mappings reads non-existent 'biolink equivalent' column from METPO sheet
`kg_microbe/utils/mapping_file_utils.py:load_metpo_mappings` (around L270, L315) reads:

```python
biolink_equivalent = row.get("biolink equivalent", "").strip()
```

and uses tree-traversal logic to find the closest ancestor that has a `biolink_equivalent` value, then maps that to a Biolink predicate via `range_to_predicate`.

The METPO classes-tab template (Google Sheet `1_Lr-9_5QHi8QLvRyTZFSciUhzGKD4DbUObyTpJ16_RU`, exported to `berkeleybop/metpo:src/templates/metpo_sheet.tsv`) does **not** have a column called `biolink equivalent`. The Biolink-mapping columns that exist are:

- `biolink close match` (col 9, ROBOT directive `AI skos:closeMatch`)
- `biolink broad match` (col 24, added today; ROBOT directive `AI skos:broadMatch`)

Result: `biolink_equivalent` is the empty string for every row, the tree-traversal never finds a non-empty parent value, and downstream predicate selection silently falls back to the default (`"has phenotype"`) and to the `subject_label`/`notes` heuristics in `_resolve_biolink_predicate`. The METPO sheet's SKOS mappings are not consumed at all.

Suggested fix: read from `biolink close match` (and optionally `biolink broad match`) instead of `biolink equivalent`. METPO maintainers use SKOS predicates deliberately; this is the canonical signal, not a missing parallel column.

Discovered while investigating berkeleybop/metpo#423.

realmarcin:
Closes #568.

## The no-op

`load_metpo_mappings` walked a term's ancestors for one with a class-level `biolink equivalent` and only then consulted the properties template's RANGE table. The pinned classes tab has no such column (its Biolink URIs moved to `biolink close match` / `biolink broad match`), so the walk always came back empty and **all 393 synonyms across four columns resolved to the default `has phenotype` with no category**. `madin_etal`, `bactotraits` and `metatraits_gtdb` therefore asserted `biolink:has_phenotype` for processes and qualities alike — the `has_phenotype`-on-process-endpoints half of #643.

## The change

One resolver (`_resolve_metpo_predicate`, now used by the main loop and the alias overrides), two walks:

- **predicate** — nearest ancestor whose *label* is a RANGE in the properties template: under `biological process` → `capable of`; under `phenotype` → `has phenotype`; under `quality` → `has quality`. Biolink predicate from the property's `biolink equivalent`, else the shared METPO→Biolink map (`biolink:has_attribute` for `has quality`).
- **category** — nearest ancestor declaring a Biolink category URI, read from `biolink equivalent` → `biolink close match` → `biolink broad match`; the ROBOT directive row (`AI skos:closeMatch`) is not a URI.

Also fixed a collision in the RANGE table: "quality" is claimed by three properties and "chemical entity" by eleven; last-row-wins handed every quality to `isolated from environment with quality`. The generic `has <RANGE>` wins, then Biolink-equivalent, then family root, then lowest id (ids alone fail: `isolated from host with quality` is METPO:2000067, `has quality` METPO:2000101).

## Measured on the pinned templates

| column | before | after |
|---|---|---|
| metatraits synonym (102) | 102 `has phenotype`, 0 categories | 75 `has phenotype` / 21 `has quality`→`biolink:has_attribute` / 6 `capable of`; 75 with a category |
| bactotraits related synonym (159) | 159 / 0 | 135 / 23 / 1; 135 with a category |

## Tests

`tests/test_metpo_predicate_derivation.py` drives a template shaped like the pinned one (no `biolink equivalent`, categories in `biolink close match`, ROBOT row second): process → `capable of`, phenotype → `has phenotype`, quality → the generic property not the colliding one, category from the close-match column, orphan → default, directive row ignored. A skip-unless-present test on the real sheet pins that it no longer collapses. Existing madin/metatraits/category-alignment tests: 62 passed.

## Follow-through

`madin_etal`, `bactotraits`, `metatraits_gtdb` must rerun for this to reach the graph, then the merge — that is the targeted rerun the rebuild decision deferred.

### DIFF
--- a/kg_microbe/utils/mapping_file_utils.py
+++ b/kg_microbe/utils/mapping_file_utils.py
@@ -185,6 +185,25 @@ class MetpoTreeNode:
         return None
 
 
+#: Classes-tab columns that can carry a Biolink category URI, in precedence order.
+_CLASS_CATEGORY_COLUMNS = ("biolink equivalent", "biolink close match", "biolink broad match")
+
+
+def _class_biolink_category(row: Dict[str, str]) -> str:
+    """
+    Return the Biolink category URI a classes-tab row declares, if any.
+
+    :param row: One row of the METPO classes template.
+    :return: The first URI found across :data:`_CLASS_CATEGORY_COLUMNS`, or "".
+    """
+    for column in _CLASS_CATEGORY_COLUMNS:
+        value = (row.get(column) or "").strip()
+        # The ROBOT template row carries `AI skos:closeMatch`, not a URI.
+        if value.startswith("http"):
+            return value
+    return ""
+
+
 def _build_metpo_tree() -> Dict[str, MetpoTreeNode]:
     """
     Build a tree structure from METPO classes based on parent-child relationships.
@@ -221,7 +240,11 @@ def _build_metpo_tree() -> Dict[str, MetpoTreeNode]:
             label = row.get("label", "").strip()
             madin_synonym = row.get("madin synonym or field", "").strip()
             bacdive_synonym = row.get("bacdive keyword synonym", "").strip()
-            biolink_equivalent = row.get("biolink equivalent", "").strip()
+            # The 2026-06-12 classes tab has no `biolink equivalent` column any more;
+            # the Biolink category now travels in `biolink close match` (and, for a
+            # handful of rows, `biolink broad match`). Reading only the old column
+            # made every lookup below fall to its defaults for months (#568).
+            biolink_equivalent = _class_biolink_category(row)
 
             if iri and label:
                 # handle pipe-separated synonyms from madin column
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass, four probes on the pinned templates:

1. **Does any synonym-bearing class inherit a wrong category?** The classes tab declares four: `phenotype` → PhenotypicQuality, `microbe` → OrganismTaxon, `chemical entity` → ChemicalEntity, `growth medium` → ProcessedMaterial. Only `phenotype` has synonym-bearing descendants; across all four synonym columns the inferred categories are PhenotypicQuality or none (metatraits 75/27, bactotraits 135/24, madin 113/25, bacdive 133/27). No term lands on OrganismTaxon or ProcessedMaterial.
2. **Transforms with an empty category still get their defaults** — madin uses `uri_to_curie(...) or <DEFAULT>`, bactotraits `if category_url else`; an empty `inferred_category` cannot produce an empty node category.
3. **Other consumers of the walk**: the alias overrides already went through `_resolve_metpo_predicate` and now get the same answer as the main loop; `MetpoTreeNode.find_biolink_equivalent_parent` has no callers outside the module. Metabolite/enzyme loaders do not use the RANGE table.
4. **The collision rule** is exercised by the test on `quality`; on the real template it picks `has quality`, `organism interacts with chemical`, `enzyme activity analyzed`, `grows in`, `has phenotype`, `capable of` — the generic property in every family.

**Filed, not fixed** — **#1000**: 44 quality classes will now reach the graph as `biolink:has_attribute` (METPO's own map for `has quality`) instead of today's `has_phenotype`. Spec-faithful, but a visible change for phenotype consumers; it should be a chosen mapping, not an inherited one.

The three consuming transforms rerun after merge, then the merge.

Issue #985, opened 2026-09-07
LPSN's paired atomic writes can still publish a torn pair if the second os.replace fails


realmarcin:
From the adversarial review of #981 (#820 fix). Low, filed so the boundary is recorded.

`lpsn.py` now writes `nodes.tsv` and `edges.tsv` through two nested `atomic_write` contexts. An exception inside the body unwinds both and neither is published — that is the #820 case and it is covered by a test. The remaining window is narrower: the body completes, the inner `os.replace` (edges) succeeds, and the outer `os.replace` (nodes) then fails (disk full at rename time, permissions changed). Result: new `edges.tsv`, old or absent `nodes.tsv`.

`atomic_write` is per-file by design and the repo has no two-file transaction. Options if it ever matters: write the pair into a staging directory and rename the directory, or have consumers cross-check the two files' fingerprints. Not worth doing for LPSN alone.

realmarcin:
Same shape one file over: `lpsn_api.py` still writes `nodes.tsv`/`edges.tsv` with a plain `open()` (`run()`, ~line 197), so a crash mid-run leaves a truncated pair that the merge would ingest as complete. lpsn itself moved to `atomic_write` in #981; noted here from the #1003 review rather than as a new issue — it is the same one-line change when someone next touches the file.

PR #1003, opened 2026-09-08
lpsn_api declares the records it links to; METPO processes are typed as processes (#991, #438)

### LINKED ISSUE #438: biolink:capable_of used with wrong object type (PhenotypicQuality instead of Occurrent)
## Issue

186,197 edges use `biolink:capable_of` predicate with objects categorized as `biolink:PhenotypicQuality`, violating Biolink Model's range constraint requiring `Occurrent` (biological processes).

## Biolink Model Constraint

**Predicate:** [`biolink:capable_of`](https://biolink.github.io/biolink-model/capable_of/)
- **Domain:** NamedThing  
- **Range:** Occurrent (biological processes/activities)
- **Definition:** "a physical entity and process or function, where the continuant alone has the ability to carry out the process or function"

## Actual Usage

```
biolink:OrganismTaxon --capable_of--> biolink:PhenotypicQuality (EC enzyme codes)
```

**Examples:**
```
NCBITaxon:1000562 --capable_of--> EC:3.1.3.1 (alkaline phosphatase)
NCBITaxon:1000562 --capable_of--> EC:1.11.1.6 (catalase)
```

## Impact

- **186,197 edges** across bacdive (186,014) and madin_etal (183) datasets
- Violates Biolink Model semantics
- Breaks reasoning/validation tools expecting Occurrent objects
- Affects ~11% of all BacDive edges

## Root Cause

EC enzyme codes (which represent catalytic activities/processes) are being categorized as `biolink:PhenotypicQuality` (see #437 for duplicate node issue). The predicate is correct, but the object categorization is wrong.

### Option 1: Recategorize EC Nodes (Recommended)
Change EC enzyme codes to `biolink:MolecularActivity` or `biolink:BiologicalProcess`:
- EC codes represent enzymatic activities, not qualities
- `MolecularActivity` is a subclass of `Occurrent`
- Fixes both this issue and #437

### Option 2: Change Predicate
Use `biolink:has_phenotype` instead:
- Keeps EC codes as `PhenotypicQuality`
- Semantically: organism has phenotype of "having alkaline phosphatase activity"
- Less precise than capable_of

## Evidence

```bash
# Check object categories for capable_of edges
$ awk -F'$' 'NR==1 {for(i=1;i<=NF;i++)h[$i]=i} 
  NR>1 && $h["predicate"]=="biolink:capable_of" {
    print $h["object"]
  }' data/transformed/bacdive/edges.tsv | head -5 | \
  xargs -I {} grep "^{}" data/transformed/bacdive/nodes.tsv | cut -f1-2

EC:3.1.3.1	biolink:PhenotypicQuality
EC:1.11.1.6	biolink:PhenotypicQuality
EC:1.11.1.6	biolink:PhenotypicQuality
EC:3.4.11.1	biolink:PhenotypicQuality
EC:3.1.3.2	biolink:PhenotypicQuality
```

## Biolink Model References

- [`Occurrent` class](https://biolink.github.io/biolink-model/Occurrent/): "A processual entity" (includes processes, activities, events)
- [`MolecularActivity` class](https://biolink.github.io/biolink-model/MolecularActivity/): "An execution of a molecular function"
- [`capable_of` predicate](https://biolink.github.io/biolink-model/capable_of/)

## Related

- #437 - EC nodes have duplicate entries with conflicting categories
- #30 - Broader KGX format compliance

## Analysis Date

2025-11-20 (build from 2025-11-11, commit 77c42d8)

### LINKED ISSUE #991: lpsn_api references 1,344 LPSN records that neither it nor lpsn declares, so the merge invents them
Found in the 2026-09-08 rebuild (`merge.noprego.yaml`, all 15 sources FRESH). `merged_stub_nodes.tsv`:

```
prefix  count  expected
lpsn    1344   no       lpsn:1002|lpsn:1005|lpsn:101
```

Attribution, by checking every `lpsn:*` endpoint in each source's edges against the union of `lpsn/nodes.tsv` and `lpsn_api/nodes.tsv`: **all 1,344 come from `lpsn_api`** — 4,559 `biolink:subclass_of` edges and 1,113 `biolink:same_as` edges point at LPSN records no node row declares. `bacdive`, `microbedecoder`, `metatraits*`, `gold`, `madin_etal`: 0.

```
lpsn:517203  biolink:subclass_of  lpsn:1002  rdfs:subClassOf  infores:lpsn
lpsn:517304  biolink:subclass_of  lpsn:1002  rdfs:subClassOf  infores:lpsn
```

`lpsn:1002` is not in `lpsn_gss.csv` (the GSS covers genus and below); it is an above-genus record that `lpsn_api`'s taxonomy enrichment links to. So the API layer asserts the parent and never emits the parent — same shape as #932 (LPSN citing deposits it never declared), one level up. KGX fills each with a bare `biolink:NamedThing`, no label, no rank, which is what the #918 report exists to catch.

Fix shape: `lpsn_api` has the JSON for the records it links to (name, rank, status), so it can write the node row beside the edge, buffered per record as #981 did for deposits. `same_as` targets (1,113) need the same treatment or a check that the target is a current name.

Rerun cost: `lpsn_api` alone (~11 s), then the merge.

realmarcin:
Closes #991. Closes #438.

## #991 — 1,344 invented `lpsn:*` nodes

`lpsn_api` emits `subclass_of` edges to the parent chain above genus and `same_as` edges to basonyms, but the GSS covers genus and below and the API layer only fetched GSS records — so nothing declared those endpoints and KGX invented 1,344 bare NamedThings in the 2026-09-08 merge.

Every linked record is now declared, once: fetched through the existing cache so the node carries LPSN's `full_name` and status, its own parent/basonym edges emitted, and the walk continued up to the domain (depth-capped at 12). When the API has nothing, a stub in our own namespace still declares the endpoint with a description saying so — an unnamed node that records its origin beats one KGX invents with nothing on it. The completion line reports `linked_records_declared` / `linked_records_stubbed`.

## #438 — four processes typed as qualities

All 148,357 `capable_of → PhenotypicQuality` edges in metatraits_gtdb point at **four** METPO classes: Fermentation, nitrification, nitrogen fixation, denitrification. The predicate is right; the category was the transform default, because the METPO `biological process` class declares no Biolink category. The resolver now falls back to the category the RANGE ancestor implies (`biological process → BiologicalProcess`, `chemical entity → ChemicalEntity`, `phenotype`/`quality → PhenotypicQuality`) when nothing on the path declares one. A declared category still wins (tested).

## Tests

- lpsn_api: every linked target has a node row; a fetchable genus is named and walks up; unfetchable family/basonym get declared stubs; a shared parent is declared once. **All four fail on the old code.**
- predicate derivation: process with no declared category → `biolink:BiologicalProcess`; declared category beats the fallback. **Fails on the old resolver.**
- 125 tests across lpsn, lpsn_api, microbedecoder, madin and category alignment pass; ruff clean.

## Follow-through

Per the decisions on record: rerun `lpsn_api`, `madin_etal`, `bactotraits`, `metatraits_gtdb`, then the **canonical `merge.yaml`** (habitat PREGO), which refreshes `merged-kg.tar.gz` and `merged_graph_stats.yaml`.

### DIFF
--- a/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
+++ b/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
@@ -153,7 +153,7 @@ class LPSNAPITransform(Transform):
         super().__init__(LPSN_API_SOURCE, input_dir, output_dir)
         self.knowledge_source = LPSN_KNOWLEDGE_SOURCE
         self._client_override = client
-        self._stats = {"fetched": 0, "from_cache": 0, "errors": 0}
+        self._stats = {"fetched": 0, "from_cache": 0, "errors": 0, "linked_declared": 0, "linked_stubbed": 0}
 
     # ------------------------------------------------------------------
     # public entry point
@@ -202,16 +202,26 @@ class LPSNAPITransform(Transform):
             node_writer.writerow(self.node_header)
             edge_writer.writerow(self.edge_header)
 
-            for record_no in self._iter_record_nos(gss_nodes):
+            # Every record the GSS declares, plus every record this file links
+            # to. The parent chain above genus and the basonym targets are not
+            # in the GSS, and without a node row KGX invents a bare NamedThing
+            # for each at merge time -- 1,344 of them in the 2026-09-08 graph
+            # (#991). Same rule LPSN applies to deposits (#932).
+            self._declared = set(self._iter_record_nos(gss_nodes))
+            for record_no in sorted(self._declared):
                 record = self._fetch(client, record_no, cache_dir)
                 if record is None:
                     continue
                 self._emit(record_no, record, node_writer, edge_writer)
+                for ref in self._linked_record_nos(record_no, record):
+                    self._ensure_declared(client, ref, cache_dir, node_writer, edge_writer)
 
         print(
             f"[lpsn_api] fetched={self._stats['fetched']:,} "
             f"cached={self._stats['from_cache']:,} "
-            f"errors={self._stats['errors']:,}"
+            f"errors={self._stats['errors']:,} "
+            f"linked_records_declared={self._stats.get('linked_declared', 0):,} "
+            f"linked_records_stubbed={self._stats.get('linked_stubbed', 0):,}"
         )
 
     # ------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


realmarcin:
## Adversarial review

Read-only pass over both changes, probes against the real cache and templates.

**#991**
- **Fetch volume is bounded and known.** Of 34,301 cached records, the parent/basonym ids outside the GSS number exactly **1,344** — the invariant's count — and none are cached, so the first rerun makes ~1.3k API calls plus the short chains above them, all through the existing cache. Credentials are in `.env`; the fetch path already prints and counts failures per record rather than aborting.
- **Linked nodes get real labels.** A real cached record carries `full_name` and `category` (rank): `Corynebacterium`, `genus`, parent 346. The rank is not written today; it would be a cheap addition to the description if wanted.
- **No double declaration**: GSS ids are pre-seeded into `_declared`, so a basonym or parent that *is* a GSS record is never re-emitted; the test for a shared genus pins one row.
- **Not addressed, pre-existing**: `lpsn_api` writes its TSVs with a plain `open()`, the #820 shape; lpsn itself was fixed in #981. Filed as a note on #985's family rather than a new issue — it is one line when someone touches the file next.

**#438**
- On the pinned templates the fallback changes exactly the terms it should: 6 + 1 + 6 + 1 synonyms (metatraits / bactotraits / madin / bacdive columns) move to `biolink:BiologicalProcess`; the 21/23/19/26 quality terms and everything under `phenotype` stay `PhenotypicQuality`; **no term lands on `ChemicalEntity`** (no synonym-bearing class sits under `chemical entity`).
- The resolver now returns a mix of URI (from a declared close match) and CURIE (from the fallback). Cosmetic: every consumer passes it through `uri_to_curie`, which returns CURIEs unchanged. Left as is.

No new issues. Reruns per the decisions after merge: `lpsn_api`, `madin_etal`, `bactotraits`, `metatraits_gtdb`, then the canonical `merge.yaml`.

PR #1005, opened 2026-09-08
Fingerprint the shared code too, hash by relative name, and migrate the markers that still hold (#1002, #983)

### LINKED ISSUE #983: data_fingerprint folds absolute paths into the digest, so a marker read from another checkout reports STALE_VS_DATA
Observation from the adversarial review of #981; pre-existing since #844.

`data_fingerprint(repo_root, data_inputs)` calls `_hash_files(repo_root / rel for rel in data_inputs)`, and `_hash_files` folds `path.as_posix()` into the digest before the content. With absolute paths, the same curation files under `/Users/a/kg-microbe` and `/home/ci/kg-microbe` hash differently.

Where it bites: a `data/transformed/<source>/` directory copied between machines (release bundles ship transformed outputs; a worktree or CI checkout lives at a different root) carries a `source_fingerprint.json` whose `data` field can never match, so `kgm-freshness-check` reports `STALE_VS_DATA` for inputs that are byte-identical. The rename-detection rationale in `_hash_files` only needs the *relative* name.

Low today — markers are normally compared in the checkout that wrote them — but it is the same trap #981 just removed from `schema_fingerprint`. Fix shape: hash `rel.as_posix()` + content, as `schema_fingerprint` now does. Note that changing the digest scheme invalidates every existing `data` field, so it should ride with a `FINGERPRINT_VERSION` bump or land just before a full rerun.

### LINKED ISSUE #1002: A change under kg_microbe/utils/ marks no transform stale: code_fingerprint hashes only the transform's own package
Found while landing #999 (#568).

`kg_microbe/utils/mapping_file_utils.py` changed how `madin_etal`, `bactotraits` and `metatraits_gtdb` derive every METPO predicate — measured afterwards as 7,741 new `biolink:has_attribute` edges in bactotraits, 917 `capable_of` in madin_etal, 227,927 `capable_of` in metatraits_gtdb. Immediately after the merge, `kgm-freshness-check` reported all three **FRESH**. Nothing was stale because `code_fingerprint(code_dir)` hashes `kg_microbe/transform_utils/<source>/` only; shared code in `kg_microbe/utils/` (mapping loaders, ontology adapters, chemical-mapping utils, atomic IO) and `transform_utils/constants.py` are invisible to it.

This is the #943 shape again — a real input the fingerprint cannot see — and it is the more common one: most behaviour-changing PRs in this repository touch `utils/`. A rebuild planned from the freshness table would have skipped the three transforms whose output the change was for; I only reran them because I knew.

Options:
- fold the behaviour digest of `kg_microbe/utils/` and `constants.py` into every transform's `code` fingerprint (simple; every utils edit then marks everything stale, which is the honest answer but a blunt one);
- per-transform import closure: hash the set of first-party modules the transform actually imports (precise; a little work to compute, `modulefinder` or a static import walk);
- record `git rev-parse HEAD` beside the fingerprint for the report only, without treating it as staleness (does not fix the gate, but makes the situation visible).

realmarcin:
Closes #1002. Closes #983.

## What changes (scheme 3 of `source_fingerprint.json`)

- **`shared` digest (#1002).** `code_fingerprint` saw only the transform's package; a change in `kg_microbe/utils/` marked nothing stale — #999 changed three transforms' predicates while every source stayed FRESH. Every marker now records a behaviour digest over `kg_microbe/utils/`, `transform_utils/constants.py` and `transform_utils/transform.py`, and `kgm-freshness-check` says which of the two moved (`code differs…` vs `shared code (utils/, constants.py, transform.py) differs…`). Blunt by design.
- **Repo-relative names (#983).** Digests folded absolute paths in, so the same files under two checkouts hashed differently; a marker read from a scratch worktree produced four spurious failures yesterday. Names are folded in relative to the repo root; a rename is still a change (tested).
- **Migration.** `FINGERPRINT_VERSION` 2 → 3 makes every existing marker unreadable, which would push today's fully rebuilt tree back to timestamp comparison. `scripts/migrate_fingerprints.py` rewrites a scheme-2 marker as scheme 3 **only when its scheme-2 digests still match the inputs** (it vouched for the output then and still does); stale ones are left for a real rerun. Two phases so downstreams are rewritten after their upstreams and see the new upstream markers.

## Tests

`tests/test_transform_fingerprint.py`: package and data digests are path-independent; renames still change them; shared-code edits move `shared` and not `code`; the marker carries `shared`; a holding marker migrates to scheme 3 with the right digests; a stale one is left (and reads as absent); a downstream is rewritten after its upstream. `FreshnessSchemaTest` extended for the new key. 47 passed; ruff clean.

## Follow-through

After merge: `poetry run python scripts/migrate_fingerprints.py` on the rebuilt tree, then `kgm-freshness-check` should still read 15 FRESH.

### DIFF
--- a/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
+++ b/.claude/skills/kgm-freshness-check/kgm_freshness_check.py
@@ -331,6 +331,7 @@ def _fingerprint_verdict(source: str, code_dir: Path) -> Optional[tuple]:
             data_fingerprint,
             read_fingerprint,
             schema_fingerprint,
+            shared_code_fingerprint,
             upstream_fingerprint,
         )
     except ImportError:
@@ -340,7 +341,12 @@ def _fingerprint_verdict(source: str, code_dir: Path) -> Optional[tuple]:
         recorded = read_fingerprint(TRANSFORMED_DIR / name)
         if recorded is None:
             continue
-        code_stale = recorded.get("code") != code_fingerprint(code_dir)
+        # The package, then the first-party code every transform shares (#1002):
+        # either moving means the output may differ, and the note says which.
+        package_stale = recorded.get("code") != code_fingerprint(code_dir, REPO)
+        shared_stale = recorded.get("shared") != shared_code_fingerprint(REPO)
+        code_stale = package_stale or shared_stale
+        code_what = "code" if package_stale else "shared code (utils/, constants.py, transform.py)"
         data_stale = recorded.get("data") != data_fingerprint(REPO, _declared_data_inputs(source))
         upstream_stale = recorded.get("upstream") != upstream_fingerprint(
             TRANSFORMED_DIR, _declared_transform_inputs(source)
@@ -363,9 +369,15 @@ def _fingerprint_verdict(source: str, code_dir: Path) -> Optional[tuple]:
                 f"{upstreams} rebuilt since this output; rerun `poetry run kg transform -s {source}`",
             )
         if code_stale and data_stale:
-            return "STALE_VS_CODE_AND_DATA", f"code and data differ from the recorded build; rerun `poetry run kg transform -s {source}`"
+            return (
+                "STALE_VS_CODE_AND_DATA",
+                f"{code_what} and data differ from the recorded build; rerun `poetry run kg transform -s {source}`",
+            )
         if code_stale:
-            return "STALE_VS_CODE", f"code differs from the recorded build; rerun `poetry run kg transform -s {source}`"
+            return (
+                "STALE_VS_CODE",
+                f"{code_what} differs from the recorded build; rerun `poetry run kg transform -s {source}`",
+            )
         if data_stale:
             return "STALE_VS_DATA", f"a declared data input differs from the recorded build; rerun `poetry run kg transform -s {source}`"
         return "FRESH", "verified by content fingerprint"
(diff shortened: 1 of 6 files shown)


realmarcin:
## Adversarial review

Read-only pass over the scheme change and the migration. Probes:

- **Symlinked inputs.** `_folded_name` relativises after `resolve()`. A curation file reached through a symlink that points *outside* the checkout (the scratch-worktree layout I used yesterday) does not relativise and falls back to the absolute path — so that one layout stays path-dependent. Real checkouts have no such links; recorded here, not filed.
- **`prego`'s marker lives in `prego_habitat/`.** The migration script carries the same alias table the freshness check uses, so it finds the marker; `upstream_fingerprint` would not, but nothing declares `prego` as a `TRANSFORM_INPUT`, so that pre-existing gap has no consumer today.
- **Migration cannot launder staleness.** Phase 1 judges every scheme-2 marker against its inputs *before* any is rewritten, and a marker whose package, data or upstream digest has moved is left in place (test: it then reads as absent, i.e. timestamp fallback, not FRESH).
- **Downstream ordering.** Topological on `TRANSFORM_INPUTS`; a cycle falls back to registration order rather than looping. Tested with `gtdb → lpsn`.
- **Bluntness of `shared`.** Any edit under `kg_microbe/utils/` now marks all 15 sources stale — the option chosen, and the honest one; the note names "shared code" so a reader can decide whether the edit could have touched their transform.
- Added in `246def29b`: a freshness test that a shared-code change reads `STALE_VS_CODE` and the note says "shared code".

No new issues. After merge: run `scripts/migrate_fingerprints.py` on the rebuilt tree and confirm 15 FRESH.

Issue #1001, opened 2026-09-08
FreshnessDataStalenessTest reads the real data/transformed markers, so it fails on any checkout whose transforms have run


realmarcin:
Found while verifying #999; reproduced on master `1d4d1eaca` with the same result.

`tests/test_transform_dispatch_and_data_inputs.py::FreshnessDataStalenessTest` mocks `_latest_commit`, `_has_local_diff`, `_output_mtime` and `_latest_data_input_commit` and then calls `check_source("bacdive", ...)`. Those mocks cover the **timestamp** path only. `check_source` consults `_fingerprint_verdict` first, which reads the real `data/transformed/bacdive/source_fingerprint.json` — and since the 2026-09-08 rebuild, bacdive has one. The content verdict wins, the mocks never matter, and two of the four tests fail:

```
FAILED …::test_an_output_older_than_its_data_input_is_stale
  'mappings/isolation_source_to_ontology.tsv' not found in 'a declared data input differs from the recorded build …'
FAILED …::test_stale_code_and_stale_data_are_reported_together
  'FRESH' != 'STALE_VS_CODE_AND_DATA'
```

CI is green because CI has no `data/transformed`. So the suite's result depends on whether the developer has run the pipeline — the #855 hermeticity concern in one more place, and the mirror image of #980's worry.

Fix: patch `read_fingerprint` (or `_fingerprint_verdict`) to return `None` in these tests so the timestamp path under test is the one exercised, and add one test that drives `_fingerprint_verdict` with a synthetic marker — `FreshnessSchemaTest` (#981) already shows the pattern.

PR #1007, opened 2026-09-08
metatraits_gtdb: pick the NCBITaxon id by genome count, not by set iteration order (#1006)

### LINKED ISSUE #1006: metatraits_gtdb is nondeterministic: ~22% of its edges and ~19% of its NCBITaxon subjects change between identical runs
Found while re-running `metatraits_gtdb` after #1003, on master `b94367e99`/`e5e213982`, with no change to its inputs, its package, the shared code paths it uses, or the worker count (9 every time).

| run | edges | nodes |
|---|---:|---:|
| 05:47 | 4,271,769 | 35,444 |
| 08:29 | 4,239,399 | 35,474 |
| run 1 | 4,224,916 | 35,437 |
| run 2 | 4,223,579 | 35,335 |
| run 3 | 4,250,151 | 35,479 |

Totals wobble by ~1%; the *content* does not. Diffing run 2 against run 3 by node id and by `(subject, predicate, object)`:

```
nodes: only in run2 = 6,626   only in run3 = 6,770   (all NCBITaxon)
edges: only in run2 = 944,512  only in run3 = 971,084
   195,975  NCBITaxon has_phenotype METPO
   109,502  NCBITaxon METPO:2000038 CHEBI
    69,235  NCBITaxon METPO:2000034 CHEBI   ...
```

So roughly a fifth of the taxon subjects are assigned to a *different NCBITaxon id* on each run, and every edge hanging off them moves with it. Every merged graph this transform has fed — including today's canonical one — contains one arbitrary draw of those assignments. Nothing reports it: the transform exits 0, the fingerprint says FRESH, and the merge invariants only see dangling endpoints.

Where it comes from, by reading: `metatraits.py` resolves a taxon name with `search_by_label(self._get_ncbitaxon_impl(), name, limit=1)` and takes `results[0]` (species name, then GTDB species, then genus). `search_by_label` is `list(oi.basic_search(label, config(limit=...)))` with no ordering of our own. When a name has several candidates (homonyms, synonyms, ranks), which one comes first is whatever oaklib yields first — and with nine worker processes each carrying its own hash seed, a set-backed candidate list yields a different "first" per process. Chunking is ruled out (same 9 chunks every run; results recombined in order with global dedup).

Severity: silent wrongness in the shipped graph, at scale, and it undermines everything built on reproducibility (#953-style fixed points, the fingerprints in #844/#1005, release diffs).

Fix shape: never take `[0]` of an unordered candidate list. Ask for all candidates, prefer an exact label match, then rank deterministically (e.g. rank-aware, then lowest CURIE), and log the ambiguous names to a report so they can be curated — the LPSN transform already follows a single-hit-only rule for the same reason (#484/#590). Check every other `search_by_label(..., limit=1)` / `[0]` site while there.

realmarcin:
Closes #1006.

## The defect

`metatraits_gtdb` builds `gtdb_to_ncbi` as a **set** of every member genome's `ncbi_taxid` per GTDB name and resolves with `list(ncbi_ids)[0]`. Set order is per-process hash order, and the map is pickled into nine workers — so a species' NCBITaxon id was a random draw on every run.

Evidence (all on identical code and inputs):
- four unpinned runs → four different outputs (4,239,399 / 4,224,916 / 4,223,579 / 4,250,151 edges);
- run 2 vs run 3: **4,665 species names resolved to different ids**, 944,512 edges only in one, 971,084 only in the other (≈22%); e.g. *Streptococcus phocae* `NCBITaxon:1000562` (a subspecies) vs `NCBITaxon:119224` (the species);
- **40,138 of 199,923 GTDB species (20.1%)** carry more than one taxid in the metadata the transform reads;
- two runs with `PYTHONHASHSEED=0` → **byte-identical** (sorted-edges md5 `cff5939f…` both).

Every merged graph this transform has fed carried one arbitrary draw of those assignments, silently.

## The fix

The map counts genomes per taxid; the pick is the taxid most genomes carry (in practice the species record — subspecies/strain ids are usually one genome each), ties to the lowest numeric id. The load step prints how many names were ambiguous. `bacdive.py`'s `search_by_label(..., limit=1)` was checked and is deterministic; not part of this.

## Tests

`tests/test_metatraits_gtdb_taxid_pick.py`: majority wins (the *S. phocae* case); a tie does not depend on insertion order; the resolver goes through the rule; and an AST guard that no `list(...)[i]` over a collection remains in the module. **All fail on the old file** (checked by swapping it). 63 metatraits/gtdb tests pass; ruff clean.

## Follow-through

After merge: rerun `metatraits_gtdb` **twice unpinned** and confirm identical output; then the canonical `merge.yaml` again, since today's `merged-kg.tar.gz` carries one draw of the old behaviour.

### DIFF
--- a/kg_microbe/transform_utils/metatraits_gtdb/metatraits_gtdb.py
+++ b/kg_microbe/transform_utils/metatraits_gtdb/metatraits_gtdb.py
@@ -6,9 +6,9 @@ Uses GTDB metadata files to map GTDB species names to NCBITaxon IDs.
 """
 
 import gzip
-from collections import defaultdict
+from collections import Counter, defaultdict
 from pathlib import Path
-from typing import Dict, Optional, Set, Union
+from typing import Dict, Optional, Union
 
 from kg_microbe.transform_utils.constants import METATRAITS_GTDB, RAW_DATA_DIR
 from kg_microbe.transform_utils.gtdb.utils import clean_taxon_name
@@ -98,8 +98,12 @@ class MetaTraitsGTDBTransform(MetaTraitsTransform):
         self.unresolved_taxa_file = self.output_dir / "unresolved_taxa.tsv"
         self.measurement_traits_file = self.output_dir / "measurement_traits.tsv"
 
-        # GTDB species name -> set of NCBITaxon IDs mapping
-        self.gtdb_to_ncbi: Dict[str, Set[str]] = defaultdict(set)
+        # GTDB name -> how many member genomes carry each NCBITaxon id. A set
+        # here, read back with list(...)[0], picked a taxid in hash order: a
+        # fifth of GTDB species carry more than one (species record plus
+        # subspecies/strain ids), so ~22% of this transform's edges changed
+        # subject on every run (#1006).
+        self.gtdb_to_ncbi: Dict[str, Counter] = defaultdict(Counter)
         # Genome accession -> NCBITaxon ID mapping (for equivalence links)
         self.accession_to_ncbi: Dict[str, str] = {}
         # Genome accession -> current GTDB species name (for hierarchical links)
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass, plus a measurement against the real metadata.

- **Is majority the right answer?** Deterministic, yes; optimal, not always. Of the 40,138 ambiguous species, the majority pick's label equals the GTDB species name for 4,445, and for **1,611** a species-labelled id was in the set but lost to a strain id with more genomes. Filed as **#1008** (rank-aware pick; needs labels in the parent, workers stay adapter-free). Not a blocker: the old code picked arbitrarily among the same candidates, so this is strictly better and, for the first time, stable.
- **Other unordered picks?** `accession_to_ncbi` is one taxid per accession (last-wins over a file read in order — deterministic). The base `metatraits` transform resolves through the label search, shown deterministic under three hash seeds. `bacdive.py:779` same. No other `list(...)[0]` in the module — the AST guard now enforces that.
- **Worker sharing**: `Counter`s pickle like sets did; `_init_from_shared_data` rebuilds `defaultdict(Counter, ...)`.
- Full suite: 1,404 passed, only the two #1001 failures.

Follow-through after merge: two unpinned runs of `metatraits_gtdb` must be byte-identical (the seed-pinned pair already is), then the canonical `merge.yaml` again — today's `merged-kg.tar.gz` carries one draw of the old behaviour.

PR #1017, opened 2026-09-10
metatraits_gtdb: prefer the NCBITaxon record labelled with the GTDB name over the genome-count majority (#1008)

### LINKED ISSUE #1008: metatraits_gtdb taxid pick: 'most genomes' is deterministic but not rank-aware; prefer the species-rank record when the set has one
From the adversarial review of #1007. Follow-up, not a blocker.

#1007 makes the GTDB-name → NCBITaxon pick deterministic: the taxid most member genomes carry, ties to the lowest id. That is the right first cut and it removes the run-to-run drift. It is not the *best* answer for every species: a species whose only species-level genome sits beside several strain-level genomes (each with its own taxid) still gets a strain id, and the tie rule ("lowest numeric id") is a convention rather than a taxonomic fact.

The rank-aware rule — prefer the id whose NCBITaxon record is the species itself, else the majority — needs NCBITaxon labels/ranks inside the workers, which the GTDB transform deliberately switches off (`ncbitaxon_name_to_id.clear()`, `_ncbi_adapter = "DISABLED"`). It could be done from `ncbitaxon_nodes.tsv` at load time in the parent (a label → id dict is small and picklable), keeping workers adapter-free.

Measurement on the current metadata is in the comment below once run; the size of the residual decides whether this is worth doing before the next release.

realmarcin:
Closes #1008.

## What

#1007 made the GTDB→NCBITaxon pick deterministic (most member genomes, ties to lowest id) but not rank-aware: **1,611** GTDB species had, among their candidates, the id whose NCBITaxon label *is* the species name, and a strain id with more genomes won.

The parent process now computes `gtdb_preferred_ncbi` — for each ambiguous GTDB name, the candidate whose NCBITaxon label is that name — and the pick consults it before the majority rule. Only that dict travels to the workers (6,056 entries), not the 925k-entry name map; the workers stay adapter-free as before.

One thing the first cut got wrong and the measurement caught: this class re-implements the parent's `__init__` by hand and never loads the name map, so the first version found **0** preferred ids on real data. It now loads the map explicitly in the parent for this step and drops it again.

## Measured on the current GTDB metadata

`60,222` ambiguous names · `6,056` have a species-labelled candidate · `1,611` of those the majority rule had chosen differently (the #1008 residual, exactly as estimated).

## Tests

`tests/test_metatraits_gtdb_taxid_pick.py` (7): preferred beats the majority; a preferred id that isn't a candidate is ignored; the resolver consults the dict; plus the #1006 tests unchanged. Ruff clean.

## Follow-through

`metatraits_gtdb` rerun (6 min, then the two-run determinism check as before) and the canonical merge.

### DIFF
--- a/kg_microbe/transform_utils/metatraits_gtdb/metatraits_gtdb.py
+++ b/kg_microbe/transform_utils/metatraits_gtdb/metatraits_gtdb.py
@@ -112,6 +112,27 @@ class MetaTraitsGTDBTransform(MetaTraitsTransform):
         self.synthetic_nodes_metadata: Dict[str, Dict[str, str]] = {}
         self._load_gtdb_to_ncbi_mapping()
         self._load_gtdb_taxonomy()
+        # Computed here, while the parent's name -> id map is still loaded:
+        # for a GTDB name whose genomes carry several taxids, the id whose
+        # NCBITaxon label *is* that name is the species record, and beats the
+        # genome-count majority (#1008). Only this small dict travels to the
+        # workers; the 925k-entry name map does not.
+        # This class re-implements the parent's __init__ by hand and never
+        # calls _load_ncbitaxon_labels (its workers never use OAK), so the map
+        # is empty here; load it once, in the parent process only, and drop it
+        # again below.
+        MetaTraitsTransform._load_ncbitaxon_labels(self)
+        self.gtdb_preferred_ncbi: Dict[str, str] = {}
+        for name, counts in self.gtdb_to_ncbi.items():
+            if len(counts) > 1:
+                labelled = self.ncbitaxon_name_to_id.get(name.lower())
+                if labelled in counts:
+                    self.gtdb_preferred_ncbi[name] = labelled
+        ambiguous = sum(1 for counts in self.gtdb_to_ncbi.values() if len(counts) > 1)
+        print(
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass, then a probe over the real preferred map (6,056 entries, current GTDB metadata):

- **Every preferred id is an exact-label species record** — 6,056 of 6,056 have an NCBITaxon label equal to the GTDB species name byte-for-byte; 0 match only case-insensitively; 0 are genus/family keys. So the homonym/rank risk the rule could carry (a same-named record at another rank) does not occur on this data.
- **849 of them override a strain majority** — e.g. *Bifidobacterium animalis* → `NCBITaxon:28025` (41 genomes) over a subspecies with 43; *Stenotrophomonas muris* → the species (31) over a strain with 84. That is the correction the issue asked for; the remaining ~760 of the 1,611 differed from #1007 only by its lowest-id tie-break.
- **Worker footprint**: only the 6,056-entry dict is pickled; the parent's 925k-entry name map is loaded for this step and cleared (verified `len == 0` after init).
- **The first cut was a no-op on real data** (0 preferred) because this class hand-rolls the parent's `__init__` and never loaded the name map; the measurement caught it before the PR opened, and the commit says so.

No new issues. Follow-through after merge: two unpinned `metatraits_gtdb` runs (must be byte-identical), then the canonical merge.

PR #1016, opened 2026-09-09
kg-model-review: read domain/range from the pinned Biolink model; house rules are labelled and counted (#1011)

realmarcin:
Closes #1011.

## What was wrong

The DomainRange check judged every biolink predicate against a hand-written table. Its `located_in` row — domain `{BiologicalEntity, Protein, Gene}` — contradicts the pinned 4.4.2 model (`named thing → named thing`) and produced **228,762 false violations, 97% of the full-cap total** in the #810 re-measure, pointing a release reviewer at gold for edges that conform.

## What it does now

- **`biolink:*` predicates are judged against the pinned model** through one BMT toolkit built after `prepare_kgx()`. (Without that, `bmt.Toolkit()` reaches for a remote schema, the offline test env refuses, the exception was swallowed, and every predicate silently read as unconstrained — a second latent defect.)
- **House conventions are explicit, labelled and counted.** `HOUSE_ALLOWANCES` holds the deliberate looser-than-model rules with a reason each; an edge that passes only by one is reported as **INFO under that rule with a count**, so the model's verdict stays visible instead of being absorbed:
  - `#834` strain `subclass_of` taxon (Biolink: OntologyClass domain/range) — gold: 471,702 each side;
  - `#643/#1000` taxa as phenotype bearers, qualities as Attributes — madin: 112,781 each side;
  - `#645` EC nodes typed Protein enabling GO activities — `Protein` is **not** a physical entity in 4.4.2, so this was never a model pass;
  - the CHEBI-macromolecule-as-MacromolecularComplex convention.
- **METPO properties** (no machine-readable domain/range) and stricter-than-model house rows stay in a hand table, and their findings are labelled `[table]`; model findings `[model]`.
- **Verbose prints every constraint**, not the first five (the #810 run hid three of six).

## Verification

- `tests/test_kg_model_review_domain_range.py` (6): organism `located_in` ecosystem is not a violation; a bare `MolecularActivity` enabling something is, labelled `[model]`; EC typed `Protein` passes only by the named house rule; `has_phenotype → PhenotypicQuality` is counted under #643 not hidden; a METPO predicate is judged by the table and labelled; verbose lists all constraints. The tests run against the **real pinned model** — the conftest's minimal fixture defines neither `enables` nor `has_phenotype`.
- Real reviews on `gold` and `madin_etal` at `--max-rows 0`: `located_in` gone; the only INFO lines are the named house rules above; "Domain/range OK" across the rest.
[...]

Issue #1011, opened 2026-09-09
kg-model-review's hand-written PREDICATE_DOMAIN_RANGE contradicts the pinned Biolink model; 97% of the full-cap 'violations' are its own located_in rule


realmarcin:
From the 2026-09-09 triage, re-measuring #810 on the canonical merge (3,330,088 / 15,299,158) at `--max-rows 20000000 --verbose`:

```
[DomainRange] WARNING: 235975 edges violate predicate domain/range across 6 distinct constraints
  biolink:located_in (subject=biolink:OrganismTaxon): 228762 edges, e.g. NCBITaxon:1000569 → gold.ecosystem:3381
  biolink:enables (subject=biolink:MolecularActivity): 6957 edges, e.g. RHEA:10000 → GO:0050168
  biolink:consumes (object=biolink:NamedThing): 225 edges, e.g. lpsn:774937 → kgmicrobe.ingredient:pyg
```

The checker does not consult the model. `kg_model_review.py:162` holds a hand-written table:

```python
PREDICATE_DOMAIN_RANGE = {
    "biolink:located_in": ({"biolink:BiologicalEntity", "biolink:Protein", "biolink:Gene"}, {"biolink:NamedThing"}),
    ...
```

Against the pinned `data/raw/biolink-model.yaml` (4.4.2) via BMT: `located in` has **domain `named thing`, range `named thing`**. `OrganismTaxon` is a `named thing`. gold's 228,762 organism → ecosystem edges are conformant; the table says otherwise, and `gold.py:76` even records that "neither constrains domain/range" when the predicate was chosen. So **97% of the reported total is the reviewer's own rule**, and a release review reading this artifact would send someone to fix gold.

What survives against the real model: `enables` with a `MolecularActivity` subject (6,957 — #645) and `consumes` with a `NamedThing` object (225, microbedecoder placeholder ingredients — #650's family), plus three constraints the excerpt does not print.

Fix shape: derive domain/range from BMT on the pinned model (the skill already loads it for category checks), keep a hand-written table only for house conventions and label those rows as such; and print all constraints, not the first three, in verbose mode. Until then the DomainRange line should say it is a house table, the way the KGXVal section says NOT RUN.

Issue #810, opened 2026-08-16
kg-model-review: default merged sampling hides real domain/range violations, and the artifact does not say so


realmarcin:
The merged review samples edges, and at the sampling used for the last release review it reported **0 warnings** on a graph that has **28,238 domain/range violations**.

Reproduced on the same merged KG (2,852,793 nodes / 14,664,554 edges), only the row cap differs:

    --max-rows  3000000  ->  "Domain/range OK across 985,894 constrained edges"   0 WARNINGs
    --max-rows 20000000  ->  "28238 edges violate predicate domain/range"         1 WARNING

The violations, all pre-existing rather than new:

| predicate | subject category | edges | example |
|---|---|---|---|
| `biolink:capable_of` | `biolink:NamedThing` | 21,196 | `lpsn:20002 -> kgmicrobe.pathway:phototrophy` |
| `biolink:enables` | `biolink:MolecularActivity` | 6,957 | `RHEA:10000 -> GO:0050168` |
| `biolink:has_phenotype` | `biolink:NamedThing` | 85 | `NCBITaxon:2169973 -> METPO:1000615` |

Two separate problems.

**1. The default is too low to gate a release.** `.claude/skills/kg-model-review/reviews/20260813_122843_merged.md` — the artifact behind the last release-adjacent review — sampled 3,000,000 of 14,664,554 edges, i.e. **20%**, and the violations happen to fall outside it. Its "Total WARNINGs: 0" is therefore a much weaker claim than it reads as.

**2. The artifact does not disclose the limitation.** The skill already does exactly the right thing for KGXVal — an artifact built without it says `**NOT RUN**` in bold, precisely so a weaker review cannot be mistaken for a full one (that was #774). Sampling coverage gets no equivalent treatment: it appears only as `(3,000,000 rows sampled)` in an INFO line, which reads like a fact rather than a caveat.

Suggested, mirroring the KGXVal precedent:

* State coverage as a percentage in the summary, and mark anything below 100% as a partial review in the same bold style as `**NOT RUN**`.
* Have `kg-release`'s gate refuse an artifact whose edge coverage is < 100%.
* Consider making full coverage the default for `--merged` and requiring an explicit opt-out. The guard that rejects `--max-rows 0` for merged reviews suggests memory was the concern, but 14.6M edges completed here without trouble, so the cap may be more conservative than necessary.

[...]

PR #1018, opened 2026-09-10
Review renderer prints every example the check returned (#1011 follow-up)

realmarcin:
Follow-up to #1016. `check_domain_range` returns every violated constraint, but `Finding.__str__` capped output at `examples[:3]`, so a verbose report of six constraints printed three. The renderer now prints everything the check returned; the checks decide how many examples are worth showing.

- `Finding.__str__`: drop the `[:3]` slice, with a comment recording why.
- `tests/test_kg_model_review_domain_range.py`: a six-example `Finding` renders all six.

Ruff reports 57 pre-existing findings in `kg_model_review.py` on `master` (the `.claude/` tree is outside the tox lint gate); this diff adds none and the test file is clean.

Refs #1011.

### DIFF
--- a/.claude/skills/kg-model-review/kg_model_review.py
+++ b/.claude/skills/kg-model-review/kg_model_review.py
@@ -305,7 +305,10 @@ class Finding:
     def __str__(self):
         icon = {"ERROR": "❌", "WARNING": "⚠️ ", "INFO": "ℹ️ "}.get(self.severity, "  ")
         s = f"    [{self.check:<8}] {icon} {self.severity}: {self.message}"
-        for ex in self.examples[:3]:
+        # Every example the check chose to return. A cap here hid three of
+        # six domain/range constraints from the #810 review; the checks decide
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff and the rest of the renderer.

- `examples[:3]` was the only cap in the rendering path (`Finding.__str__`); `format_text` adds none. The ten remaining `[:5]` slices all sit inside individual checks, which is the division of labour the comment states — a check chooses its examples, the renderer prints them.
- Consequence to be aware of: a check that returns a long list now prints all of it. Today every check that can return more than five already truncates itself, so the report length is unchanged for existing checks; `check_domain_range` is the one that deliberately returns everything (up to the number of violated constraints, 17 in the #810 measurement).
- The test constructs a `Finding` directly and asserts on `str()`; it does not depend on the pinned model, so it runs where the model-dependent tests skip.

No new issues.

PR #1019, opened 2026-09-10
lpsn_api: label the 665 API-less basonyms from LPSN's web page (#1004)

### LINKED ISSUE #1004: lpsn_api: 665 of 1,789 linked above-genus LPSN records return nothing from search(id=…), so they are declared as unnamed stubs
From the first rerun after #1003 (#991), master `b94367e99`:

```
[lpsn_api] fetched=1,124 cached=34,301 errors=665 linked_records_declared=1,124 linked_records_stubbed=665
```

Every one of the 665 is the same failure — `[lpsn_api] no record returned for N` (e.g. 4359, 4120): `client.search(id=N)` followed by `client.retrieve()` yields an empty list. The ids are low-numbered above-genus records (families, orders, classes) reached by walking `lpsn_parent_id`. The 1,124 that did resolve carry `full_name` and `category`, so the API can serve this rank in general; the 665 look like records the `id=` search does not index, or that need a different endpoint (`advanced_search` with `category`, or the record page URL in `lpsn_address`).

Consequence today: the endpoints are declared (no more KGX-invented nodes; `merged_stub_nodes.tsv` will show 0 `lpsn`), but 665 taxonomy nodes above genus have no label — the description says why. Better than an invented NamedThing, still 665 unnamed nodes in our namespace.

To settle: try one of the failing ids against the API by hand (`lpsn` package, `search(id=4359)` vs `advanced_search(...)`), and check whether the GSS export or LPSN's website exposes names for these records so a stub can at least be labelled from a second source.

realmarcin:
Closes #1004.

## What the 665 are

Measured on the 35,425-record API cache and the live API (details on the issue): every one of the 665 `no record returned` ids is a `basonym_id` target of a species/subspecies record — none is an `lpsn_parent_id` — and they are pre-Approved-Lists names the JSON API does not serve by any route (`fetch/<id>` → `count: 0`; `flexible_search` → 0; `advanced_search?id=` → 400). LPSN's website serves each at `https://lpsn.dsmz.de/taxon/<id>` without auth, with `<title>Species: Bacterium sonnei</title>`.

## Change

- `lpsn_api.py`: on an API miss for a linked record, `_fetch_from_web` fetches the taxon page (`fetch_taxon_page`, `requests`, 30 s timeout), `parse_taxon_page` takes rank + name from the title and the two status lines from the body, the parsed record is cached under `data/raw/lpsn/web_cache/<id>.json` (apart from `api_cache/`, marked `kgmicrobe_source: lpsn-web-page`), and `_make_web_linked_node` writes a labelled node whose description says the label came from the page. No edges are minted from a page. A failed fetch keeps the unnamed stub; a page with no usable title is printed and counted. New summary field `linked_records_from_web`.
- `README.md`: the web-cache paragraph.
- `tests/test_lpsn_api_transform.py`: the fixture injects a fake page fetcher (`page_fetch=`), so the suite stays offline; new tests for the web-labelled node, the separate cache (second run never calls the fetcher), and the parser.

## Verification

- `pytest tests/test_lpsn_api_transform.py`: 19 passed; ruff clean.
- Canary through the real path: `_fetch_from_web("4387")` with the default fetcher wrote `data/raw/lpsn/web_cache/4387.json` (294 bytes) for *Bacterium sonnei*, with both statuses. The remaining 664 fetches run in the `lpsn_api` rerun after merge; the summary line should read `linked_records_from_web=665 linked_records_stubbed=0`, and `merged_stub_nodes.tsv` should stay at 0 `lpsn`.

### DIFF
--- a/kg_microbe/transform_utils/lpsn_api/README.md
+++ b/kg_microbe/transform_utils/lpsn_api/README.md
@@ -64,6 +64,15 @@ Every API response is cached to `data/raw/lpsn/api_cache/<record_no>.json`
 (gitignored) so partial runs resume cleanly and re-runs skip
 already-fetched records.
 
+Records the API does not serve at all — every basonym that predates the
+Approved Lists (665 in the 2026-09-10 graph, #1004) — are labelled from
+LPSN's public web page `https://lpsn.dsmz.de/taxon/<record_no>` instead:
+rank and name from the page title, nomenclatural and taxonomic status from
+the page body. Those are cached apart, under `data/raw/lpsn/web_cache/`,
+and the node's description says the label came from the page. No edges are
+minted from a page. If the page fetch fails too, the node is an unnamed
+stub in our namespace, as before.
+
 ## Expected wall-clock
 
 - **Cold cache** — 6–10 hours for 34,300 records (LPSN's empirical
--- a/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
+++ b/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
@@ -54,11 +54,14 @@ The merge step then combines both nodes.tsv / edges.tsv files.
 """
 
 import csv
+import html
 import json
 import os
+import re
 from pathlib import Path
-from typing import Any, Dict, Iterator, Optional, Union
+from typing import Any, Callable, Dict, Iterator, Optional, Union
 
+import requests
 from dotenv import load_dotenv
 
 from kg_microbe.transform_utils.constants import (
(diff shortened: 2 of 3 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus two live probes.

- **A bogus record number cannot acquire a bogus label.** `https://lpsn.dsmz.de/taxon/999999999` answers HTTP 404 with `<title>Slim Application Error</title>`; the status guard rejects it before the parser, and the parser rejects the title anyway (no `rank: name` shape) — `parse_taxon_page` returns `None` on that body.
- **Web records cannot leak into the API cache or be mistaken for one.** Separate directory, `kgmicrobe_source: lpsn-web-page` in the dict, and `_fetch` never reads `web_cache/`; the second-run test pins that the fetcher is not called again for a cached page.
- **Failure modes stay loud and non-fatal.** Transport errors, non-200s, and unparsable titles each print and count; none aborts the run, and none is cached, so a transient failure retries next run instead of freezing a stub.
- **One wording gap** (fixed in this PR): the stub's description still says "not retrievable from the LPSN API" — after this change a stub means the web page failed too, and the text should say so.
- Not exercised by the canary: 665 sequential page fetches against the website. No delay is inserted; the API pull runs at ~1–2 req/s without one, and `requests` is synchronous. If the site throttles, the failures surface as `linked_records_stubbed>0` in the summary line rather than silently.

No new issues to file; the wording fix is the only change.

PR #1022, opened 2026-09-10
kg transform: isolate a failed source, skip its dependents, preflight inputs, accept one ontology (#685, #690)

### LINKED ISSUE #685: kg transform: one transform's hard failure aborts all remaining sources
Raised by the adversarial review of #680.

## Problem

`kg_microbe/transform.py:108-115` dispatches transforms in a bare loop with no `try/except`:

```python
for source in sources:
    if source in DATA_SOURCES:
        t = DATA_SOURCES[source](input_dir, output_dir)
        ...
        t.run(show_status=show_status)
```

Any exception from one transform kills the whole run.

## Why this now matters

#680 makes `mediadive` raise `FileNotFoundError` when its bulk data is missing (previously it degraded silently to years-old caches — see #681). `MEDIADIVE` is the **10th of 17** entries in `DATA_SOURCES`.

Scenario: a user runs bare `poetry run kg transform` without MediaDive bulk data. Hours of upstream work (`ontologies`, `bacdive`, `gtdb`) complete, then MediaDive aborts and `madin_etal`, `metatraits`, `metatraits_gtdb`, `rhea_mappings`, `bactotraits`, `microbedecoder`, and `prego` never run.

Failing is still better than the silently-stale graph it replaces, but failing *late* is avoidable.

## Suggested fix

Either:
1. **Pre-flight validation** — check each selected source's inputs before running any of them, so a missing prerequisite fails in seconds rather than hours. This is the better fix and generalizes beyond mediadive.
2. Per-source `try/except` that records the failure, continues, and exits non-zero at the end with a summary of which sources failed.

Option 1 is preferable; option 2 risks masking failures unless the non-zero exit and summary are unmissable.

### LINKED ISSUE #690: kg transform -s ontologies is all-or-nothing; the per-ontology branch in transform.py:112 is unreachable
## Problem

`poetry run kg transform -s ontologies` always processes **all 14 ontologies**. There is no way from the CLI to re-run just one, even though `OntologiesTransform.run()` accepts a `data_file` argument specifically to scope to a single ontology.

`kg_microbe/transform.py:108-115` looks like it supports this:

```python
for source in sources:
    if source in DATA_SOURCES:
        t = DATA_SOURCES[source](input_dir, output_dir)
        if source in ONTOLOGIES_MAP.keys():   # <-- line 112, never true
            t.run(ONTOLOGIES_MAP[source])
        else:
            t.run(show_status=show_status)
```

…but the branch is **unreachable**. The outer guard requires `source in DATA_SOURCES`, and the two key sets are disjoint:

```
ONTOLOGIES_MAP keys : chebi, ec, envo, foodon, go, hp, metpo, mondo,
                      ncbitaxon, pato, ro, taxrank, uberon, upa
DATA_SOURCES keys   : ontologies, ontologies_stubs, bacdive, bakta, cog,
                      gtdb, kegg, lpsn, lpsn_api, mediadive, madin_etal,
                      metatraits, metatraits_gtdb, rhea_mappings,
                      bactotraits, microbedecoder, prego

intersection: EMPTY
```

So `-s ec` is silently ignored (not in `DATA_SOURCES`), and `-s ontologies` falls to the `else` branch → `run(data_file=None)` → loops all of `ONTOLOGIES_MAP`.

## Not a regression — never worked

`git log -S "if source in ONTOLOGIES_MAP"` points at `9b35088` ("ontologies_map fix", 2024-08-08). `DATA_SOURCES` at that commit already keyed on `ONTOLOGIES` only, with no per-ontology entries. This is dormant intent from the start rather than something that broke.

## Motivating case

#680 fixed a duplicated header row in the EC output. Repairing `data/transformed/ontologies/ec_edges.tsv` (12,023 rows) requires re-running only `ec`, but the CLI forces a full pass including `ncbitaxon_edges.tsv` (925,268 rows) and `chebi_edges.tsv` (417,074 rows).

## Workaround that works today

```bash
poetry run python -c "
from kg_microbe.transform_utils.ontologies.ontologies_transform import OntologiesTransform
OntologiesTransform(input_dir='data/raw', output_dir='data/transformed').run(data_file='ec.owl.gz')
"
```

The explicit dirs are required: `Transform.DEFAULT_INPUT_DIR` resolves to `kg_microbe/transform_utils/data/raw`, which does not exist.

## Suggested fix

Options, roughly in order of preference:

1. **Accept a sub-source selector**, e.g. `kg transform -s ontologies:ec` (or `--ontology ec`), and pass it through as `data_file`. Keeps `DATA_SOURCES` clean and makes the existing `run(data_file=...)` reachable.
2. **Register per-ontology keys in `DATA_SOURCES`** so the existing line-112 branch becomes live. Simplest diff, but inflates the `-s` namespace by 14 entries and makes `kg transform` with no `-s` ambiguous (would it run each ontology twice — once via `ontologies`, once individually?). Needs care.
3. If neither is wanted, **delete the unreachable branch** so it stops implying a capability that does not exist.

Whichever route, worth a test asserting the selector actually scopes — the current dead branch is exactly the failure mode an assertion would have caught.

## Related

- #533 (open umbrella, ergonomics) states *"Transform already supports per-source execution (`-s` flag); download does not."* That is true at the source level and false at the sub-source level; this issue is the gap. #533's "expose `--tags` in `kg download`" item also looks already done — CLAUDE.md documents `kg download -t ontologies -t gtdb` — so that issue may need a re-read.
- #685 — a transform's hard failure aborting the remaining sources. Different axis (failure blast radius vs. selection granularity), same underlying theme of coarse pipeline control.
- #683 — the post-#680 re-run/re-merge that surfaced this.

realmarcin:
Closes #685. Closes #690.

## #685 — one failure no longer aborts the batch

- `transform()` catches `Exception` per source, prints the traceback, continues, and raises `TransformBatchError` at the end with a per-source summary (`failed:` / `skipped:` lines). `run.py` maps it to `click.ClickException` → exit 1 without a second traceback.
- A source whose `TRANSFORM_INPUTS` name a failed (or skipped) source is **skipped**, not run on stale upstream output — `gtdb` failing skips `lpsn`, `lpsn_api`, `microbedecoder`; `gold` still runs.
- `BaseException` (`FatalOntologyError`, Ctrl-C) still aborts immediately (CLAUDE.md rule).
- Preflight: every selected source's declared `DATA_INPUTS` is checked before anything runs; a missing curation file raises `FileNotFoundError` in seconds and nothing runs. (Raw downloads are not declared per transform, so option 1 from the issue is possible only for curation inputs; option 2 covers the rest.)

## #690 — `-s ec` runs one ontology

The unreachable branch tested the source name against `ONTOLOGIES_MAP`, whose keys never overlap `DATA_SOURCES`. An ontology name is now accepted in `-s`, runs `OntologiesTransform.run(<file>)` alone, reports only `<ontology>_nodes/edges.tsv`, and **does not** rewrite the directory-wide `ontologies` fingerprint (one refreshed ontology does not make the other thirteen current; the old marker stays and reads STALE if the code moved). The unknown-source error now lists both kinds of accepted name.

## Tests

`tests/test_transform_dispatch_and_data_inputs.py` — 8 new tests with fake registries: failure isolated and reported; dependents skipped with the chain named; `BaseException` propagates; clean batch raises nothing; missing `DATA_INPUTS` fails before any run; ontology name runs only that file and leaves no fingerprint; completion line counts only that ontology; unknown name lists ontologies. The two `FreshnessDataStalenessTest` failures in this file are #1001 (pre-existing on this checkout, next in the queue).

## Canary (real path)

`poetry run kg transform -s ec`: rc 0, 11.6 min, `[transform] ec: done — ec_nodes.tsv: 11,871 rows; ec_edges.tsv: 12,021 rows (single ontology; ontologies fingerprint not updated)`; `source_fingerprint.json` and `ec_edges.tsv` byte-identical before/after. `ec_nodes.tsv` differs in exactly one row — `dcterms:license` vs `dct:license` for the same IRI — a pre-existing prefix nondeterminism in the ontologies transform, filed as a separate issue.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,9 +32,16 @@ poetry run kg download -t ontologies -t gtdb
 poetry run kg download -i -t mediadive
 poetry run kg transform
 poetry run kg transform -s bacdive -s mediadive
+poetry run kg transform -s ec          # one ontology; leaves the ontologies fingerprint alone
 poetry run kg merge -y merge.yaml
 make run-summary
 ```
+
+`kg transform` checks every selected source's declared `DATA_INPUTS` before
+running anything, isolates a failure to its source (later sources still run;
+those that declare the failed one in `TRANSFORM_INPUTS` are skipped rather
+than built on stale upstream output), and exits non-zero with a per-source
+summary. `FatalOntologyError` and Ctrl-C still abort at once. See #685, #690.
 
 `kg download` skips existing files, but a pin change now takes effect on its own: the URL behind each artifact is recorded in `data/raw/.download_manifest.json`, and a file whose declared URL has moved is re-fetched. A file with no record is left alone — unknown provenance is not wrong provenance (#911). `-i` still invalidates every selected entry, so always combine it with one or more `-t` tags. MediaDive invalidation also clears
 its response cache and can trigger an approximately one-hour crawl.
--- a/kg_microbe/run.py
+++ b/kg_microbe/run.py
@@ -84,12 +84,17 @@ def transform(*args, **kwargs) -> None:
     :param sources: A list of sources to transform.
     :return: None
     """
+    from kg_microbe.transform import TransformBatchError
     from kg_microbe.transform import transform as kg_transform
 
     try:
         kg_transform(*args, **kwargs)
     except ValueError as e:
         raise click.BadParameter(str(e), param_hint="--sources") from e
+    except (FileNotFoundError, TransformBatchError) as e:
+        # Per-source tracebacks were already printed; end with the summary
+        # and a non-zero exit, not a second traceback (#685).
+        raise click.ClickException(str(e)) from e
 
     return None
 
--- a/kg_microbe/transform.py
+++ b/kg_microbe/transform.py
@@ -233,10 +233,15 @@ def transform(
 
     failed: dict = {}
     skipped: dict = {}
+    # Names a later source may declare in TRANSFORM_INPUTS that did not
+    # complete. A single ontology failing counts as `ontologies` failing:
+    # gold and prego declare the directory, not the ontology inside it.
+    unavailable: set = set()
     for source in sources:
-        upstream = [u for u in getattr(DATA_SOURCES.get(source), "TRANSFORM_INPUTS", ()) if u in failed or u in skipped]
+        upstream = [u for u in getattr(DATA_SOURCES.get(source), "TRANSFORM_INPUTS", ()) if u in unavailable]
         if upstream:
             skipped[source] = upstream
+            unavailable.add(source)
             print(f"[transform] {source}: skipped — upstream {', '.join(upstream)} did not complete", flush=True)
             continue
         # print, not logging.info: the CLI does not configure a handler that
@@ -247,6 +252,9 @@ def transform(
             _run_one(source, input_dir, output_dir, show_status)
         except Exception as exc:  # noqa: BLE001 - isolate one source; BaseException still aborts the batch
             failed[source] = exc
+            unavailable.add(source)
+            if source not in DATA_SOURCES:
+                unavailable.add(ONTOLOGIES)
             print(f"[transform] {source}: FAILED — {type(exc).__name__}: {exc}", flush=True)
             traceback.print_exc()
 
--- a/kg_microbe/transform.py
+++ b/kg_microbe/transform.py
@@ -1,6 +1,7 @@
 """Transform module."""
 
 import inspect
+import traceback
 from functools import cached_property
 from importlib import import_module
 from pathlib import Path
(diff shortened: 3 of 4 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus probes against the registry.

- **Dependency naming gap (fixed in this PR).** `gold` and `prego` declare `ontologies` in `TRANSFORM_INPUTS`; a failed `-s ec` was recorded as `ec`, so `-s ec -s gold` would have run gold on the half-refreshed directory. A failed single ontology now also marks `ontologies` unavailable; test added.
- **Skip is transitive.** A skipped source is itself unavailable, so `gtdb` → `lpsn` (skipped) → `lpsn_api` (skipped) chains; the test pins the chain and the message names the immediate upstream.
- **`run()` signatures.** Every registered transform's `run` accepts `show_status=` as a keyword (`ontologies_stubs` via `**kwargs`); the isolation path calls it exactly as before, so no signature drift.
- **Preflight resolves `DATA_INPUTS` against the same repo root the fingerprint uses** (`Path(__file__).resolve().parent.parent`), so the two cannot disagree about where a curation file lives.
- **`BaseException` passthrough verified by test**, and `KeyboardInterrupt` is one, so Ctrl-C still stops a batch at once rather than being logged as a per-source failure.
- **Not covered:** raw-download prerequisites are undeclared, so the preflight cannot catch the MediaDive-bulk case from the issue body before the run — that still surfaces as an isolated failure with the rest of the batch completing. Declaring raw inputs per transform is a separate change.

Filed from the canary: #1023 (`dcterms:` vs `dct:` prefix nondeterminism in `ec_nodes.tsv`).

PR #1024, opened 2026-09-10
kg merge: stamp merged_graph_stats.yaml with provenance and count every predicate as written (#1013, #993)

### LINKED ISSUE #993: merged_graph_stats cannot name any METPO-predicate edge: ~10M edges appear as predicate None and are absent from count_by_predicates
Observed on the 2026-09-08 rebuild's `merged_graph_stats_no_prego.yaml` (15,331,182 edges); identical shape in the committed `merged_graph_stats.yaml` at HEAD (0 `METPO:` keys there too), so pre-existing.

- `count_by_predicates` lists only `biolink:*` predicates. `METPO:2000028`, `METPO:2000517`, `METPO:2000511` (706,765 edges, the #909 predicate) — none appear.
- `count_by_spo` renders them as `<subject category>-None-<object category>`: **10,185,324 of 20,120,253** category-pair counts (50.6%) sit under a `None` predicate.
- The `predicates:` list has 0 METPO entries.

Scale from the fresh transform outputs: **9,995,090 edges** carry a `METPO:` predicate (bacdive 3.58M, metatraits 3.31M, metatraits_gtdb 3.01M, mediadive 55k, madin_etal 42k) — roughly two thirds of the merged graph. `kgx.graph_operations.summarize_graph` resolves predicates through the Biolink toolkit and records anything it cannot resolve as `None`; METPO predicates are deliberately not Biolink (CLAUDE.md, "METPO overrides biolink:located_in").

Consequences: the stats file the README points readers at, the #689-style "did predicate X disappear" checks, and any release diff built on it are blind to the largest block of edges. A predicate could vanish or double and the stats would not move.

Fix shape: post-process the stats with a direct `cut -f2 | sort | uniq -c` over the merged edges (the `kg-release-diff` skill already does this from the TSVs), or emit a `count_by_raw_predicate` section beside KGX's. Not a KGX bug to fix upstream — it is doing what its Biolink-only summarizer says.

### LINKED ISSUE #1013: merged_graph_stats.yaml records no provenance: nothing in the file says which merge, commit or config produced it
From the adversarial review of #1012 (refreshing the stats file for #689).

The file's top-level keys are `edge_stats`, `graph_name`, `node_stats`. There is no merge date, no commit, no config name, no input list. So the committed copy cannot be told apart from a stale one except by diffing counts against a graph one already trusts — which is how the 2026-08-02 copy sat beside a 2026-09-08 graph for a day, and how a locally modified copy of unknown origin sat in the working tree for two days before this refresh (#689 comments).

KGX's `generate_graph_stats` writes only what it computes. The merge step already knows the rest: `merge.yaml`'s path, `git rev-parse HEAD`, the set of input TSVs and their `source_fingerprint.json` digests (#1005), and the wall-clock date. Writing a small `provenance:` block into the YAML after KGX produces it — or a sibling `merged_graph_stats.provenance.json` — would make the artifact self-describing, the way `mapping_tool_version` did for the mappings SSSOM (#961/#970).

Same request applies to `merged_graph_stats_no_prego.yaml` and the other variant outputs.

realmarcin:
Closes #1013. Closes #993.

## Change

`kg_microbe/merge_utils/stats_provenance.py`, hooked into `_cleanup_merged_outputs` right after the invariants check (same isolation: a failure prints and the archive still ships):

- **`provenance`** block: `generated_at`, `commit` (short HEAD; `-dirty` only when a *tracked file other than the stats file* differs — KGX has just rewritten the stats file, so `git describe --dirty` would always say dirty), `merge_config`, `edges_file`, `python`, `kgx`, and per-source `source_fingerprint.json` digests (`no marker in <dir>` when absent).
- **`edge_stats.count_by_raw_predicate`**: the predicate column counted as written, no Biolink resolution, so `METPO:*` predicates sit next to `biolink:*`. Refuses to write if its total ≠ KGX's `total_edges`.
- The stats filename is read from the config's `generate_graph_stats` operation, so every generated variant (`merged_graph_stats_no_prego.yaml`, …) is annotated the same way.
- CLAUDE.md: one paragraph under Data flow.

## Canary (real archive, scratch copy of the stats)

15,392,517 edges: raw total == KGX `total_edges`; `METPO:2000511` = 706,765; 67 METPO predicates carrying 7,656,831 edges; 88 predicates in all; 166 s.

The tracked `merged_graph_stats.yaml` is **not** rewritten in this PR: recording today's HEAD against an archive that was merged on `4fcea53b1` would be the false provenance the issue is about. The block lands with the next canonical merge (scheduled at the end of this batch).

## Tests

`tests/test_merge_stats_provenance.py` — 8 tests: raw counts include METPO and sum to the total; KGX's block kept; provenance fields and source markers; idempotent re-annotation; mismatched edges file refused; filename from config (incl. the real `merge.yaml`); stats file never makes the commit dirty; non-checkout records `unknown`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,13 @@ transform classes -> data/transformed/<source>/{nodes,edges}.tsv
 merge.yaml -> data/merged/merged-kg.tar.gz + merged_graph_stats.yaml
 ```
 
+`merged_graph_stats.yaml` carries a `provenance` block (when, commit, merge
+config, each source's fingerprint digest) and
+`edge_stats.count_by_raw_predicate`, the predicate column counted as
+written. KGX's own `count_by_predicates` resolves through Biolink and records
+every METPO predicate as `None`, so it cannot see two thirds of the edges;
+compare predicates with the raw block. See #993 and #1013.
+
 - `download.yaml` owns upstream URLs and pinned versions.
 - `kg_microbe/transform_utils/<source>/` owns source parsing and normalization.
 - `kg_microbe/transform_utils/transform.py` owns the standard TSV headers.
--- a/kg_microbe/merge_utils/merge_kg.py
+++ b/kg_microbe/merge_utils/merge_kg.py
@@ -12,6 +12,7 @@ import networkx as nx  # type: ignore
 import yaml
 
 from kg_microbe.merge_utils.invariants import check_merged_invariants
+from kg_microbe.merge_utils.stats_provenance import annotate_graph_stats, stats_filename_from_config
 from kg_microbe.transform_utils.constants import (
     AGENT_TYPE_COLUMN,
     CATEGORY_COLUMN,
@@ -142,6 +143,11 @@ def load_and_merge(
     return merged_graph
 
 
+def _repo_root() -> Path:
+    """Return the checkout root; merge configs and stats paths are relative to it."""
+    return Path(__file__).resolve().parents[2]
+
+
 def _cleanup_merged_outputs(yaml_file: str) -> None:
     r"""
     Defensive post-merge normalization of the KGX-merged TSVs.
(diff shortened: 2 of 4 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff, plus the canary numbers.

- **The count cannot silently describe the wrong file.** `annotate_graph_stats` refuses when its raw total differs from KGX's `total_edges` (tested), and it is called while the loose `merged-kg_edges.tsv` from *this* run exists, before the re-tar. A stale sibling cannot be counted by accident because the path comes from the same `dest` the invariants check uses.
- **Dirty flag is meaningful, not constant.** `git describe --dirty` would have said `-dirty` on every merge (KGX rewrites the stats file first); `git_commit` ignores exactly that path and counts nothing untracked. The test checks the ignore can only make the verdict cleaner.
- **Variants covered without a name list.** The filename is read from the config's `generate_graph_stats` operation, which `scripts/generate_merge_configs.py` rewrites per variant, so `merge.noprego.yaml` etc. get the block with no further wiring.
- **Isolation matches the invariants check.** A failure prints `[merge-stats] annotation skipped:` and the archive still ships — a bookkeeping error cannot decide whether the artifact exists.
- **Nits, not blocking:** the 14 ontology sources each repeat the same `ontologies` marker digest (accurate, verbose); `edges_file` names the loose TSV, which the cleanup then folds back into the archive — the path is what was counted, not where it is now. `generated_at` makes every re-merge a one-line diff in the committed file, which is the intent.
- **Not exercised:** the hook inside a real `kg merge` (only the function on the real archive). The end-of-batch canonical merge is the first run through the hook; its `[merge-stats]` line is the acceptance check.

No new issues.

Issue #681, opened 2026-08-06
mediadive: YAML medium cache and requests_cache have no expiry or invalidation


realmarcin:
Split out from #680, which added a guard but did not fix the underlying cache design.

## Problem

Two caches in the MediaDive transform are permanent and unversioned:

1. `kg_microbe/transform_utils/mediadive/mediadive.py:139` — `requests_cache.install_cache("mediadive_cache")` is called with **no `expire_after`**. `mediadive_cache.sqlite` on this machine is 33 MB and holds `rest/solution/*` responses created in 2025 that are still served verbatim today.
2. `mediadive.py:759-762` — `get_json_object()` falls back to the per-medium YAML cache under `tmp/medium_yaml/`, guarded only by `if not fn.is_file():`. Freshness is never checked. Files there are dated **Oct 30 2023**.

## Impact (observed, not hypothetical)

`data/transformed/mediadive/edges.tsv` (Jul 24 15:38) was produced before `data/raw/mediadive/*.json` finished downloading (Jul 25 00:14), so the bulk path was unavailable and both caches took over. Result:

- `tmp/medium_yaml/916.yaml` (2023) records `solution_id: 1885` named `Solution F`.
- Current MediaDive has restructured this: `1885` is "Ferrous chloride solution (5.2%)", wrapped by a new solution `6570` named "Solution F".
- So the shipped output asserts `solution:1878 -has_part-> 1885`, skipping the wrapper, and emits a node `mediadive.solution:1885` labelled "Solution F". `mediadive.solution:6570` has no node row and no edges at all.

Drift scale: 22 KG solution subjects and 24 KG solution objects reference IDs absent from today's `solutions.json`; 13 raw solutions never appear as a KG subject.

## Suggested fix

- Give `requests_cache` an explicit `expire_after`, or key the cache on the MediaDive release/download date.
- Add invalidation (or at minimum an age check + warning) to the YAML medium cache.
- Consider dropping the YAML cache entirely now that bulk download is the supported path.

#680 makes the silent-fallback path refuse to run, which prevents recurrence of the *silent* failure, but the caches themselves are still unbounded and will serve stale data whenever they are used deliberately.

PR #1025, opened 2026-09-10
mediadive: own the API fallback's HTTP cache instead of patching requests.Session process-wide (#624)

### LINKED ISSUE #624: MediaDiveTransform still installs the global requests cache

realmarcin:
Closes #624.

`MediaDiveTransform.__init__` called `requests_cache.install_cache("mediadive_cache")`, which monkeypatches `requests.Session` for the whole process from a constructor. #612 removed the same call from the download path; this does it for the transform.

## Change

- `__init__` no longer touches `requests`. The transform opens **one `CachedSession` it owns** on the first API call (`_http_session`), backed by `<bulk_data_dir>/mediadive_transform_cache.sqlite`; `run()` closes it in a `finally` (`_close_http`).
- The legacy `mediadive_cache.sqlite` the old code left in the working directory is adopted (moved) on first use so cached responses are not re-downloaded; a failed move is reported and a fresh cache used.
- `_get_mediadive_json` uses the owned session; module-level `requests.get` is gone from the transform.
- The stale-cache guard's docstring now names the transform's cache rather than `requests_cache`.

The transform does not thread, so one session is the right scope; the downloader's per-thread sessions stay as they are.

## Tests

`tests/test_mediadive_http_cache.py` — 5 tests: constructing the transform leaves `requests.Session` untouched and opens no session; the session is a `CachedSession` beside the bulk data and is closed; the API helper never calls `requests.get`; the legacy CWD cache is adopted; `run()` closes the session even when the transform raises.

The issue's order-dependent repro passes in the failing order: `pytest tests/test_transform_category_alignment.py tests/test_mediadive_bulk_download.py` → 34 passed, 4 skipped.

No rerun needed: the bulk data path is unchanged; the API fallback only runs for media the bulk JSONs lack.

### DIFF
--- a/kg_microbe/transform_utils/mediadive/mediadive.py
+++ b/kg_microbe/transform_utils/mediadive/mediadive.py
@@ -19,6 +19,7 @@ import csv
 import json
 import math
 import os
+import shutil
 import time
 from pathlib import Path
 from typing import Dict, Optional, Union
@@ -26,8 +27,9 @@ from urllib.parse import urlparse
 
 import pandas as pd
 import requests
-import requests_cache
 import yaml
+from requests_cache import CachedSession
+from requests_cache.backends.sqlite import SQLiteCache
 from tqdm import tqdm
 
 from kg_microbe.transform_utils.constants import (
@@ -124,6 +126,12 @@ from kg_microbe.utils.pandas_utils import (
     drop_duplicates,
 )
 
+#: HTTP response cache for the API fallback, kept beside the bulk JSONs. The
+#: old name is the file `requests_cache.install_cache("mediadive_cache")` left
+#: in the working directory; it is adopted on first use so nothing re-downloads.
+HTTP_CACHE_FILENAME = "mediadive_transform_cache.sqlite"
+LEGACY_HTTP_CACHE_FILENAME = "mediadive_cache.sqlite"
+
 
 class MediaDiveTransform(Transform):
     """Template for how the transform class would be designed."""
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass plus three probes.

- **No other caller depended on the global patch.** `rg "requests\.(get|post)"` over the mediadive transform, the bulk downloader and `chemical_mapping_utils` finds none besides the one call site now routed through the owned session, so removing `install_cache()` cannot silently un-cache anything else in the run.
- **Expiry is unchanged (`expire_after=-1`, never).** That is the behaviour the stale-cache guard (`KG_MEDIADIVE_ALLOW_STALE_CACHE`) exists to gate; this PR moves the cache, it does not change its lifetime. #681 stays the issue for expiry.
- **The `finally` path is safe on a half-built instance.** `_close_http` uses `getattr(self, "_http", None)` because the stale-guard tests construct via `__new__`; without it the guard's `FileNotFoundError` was being replaced by an `AttributeError` from the `finally` — caught by the existing test, fixed before pushing.
- **Legacy adoption moves a 33 MB file out of the repo root on the next real run.** Intended (it is gitignored and was only ever there by accident), and the message says where it went.
- **Not exercised:** a live API fallback through the new session. The bulk JSONs are complete on this checkout, so the fallback path does not fire; the unit test covers the wiring with a fake session.

No new issues.

PR #1026, opened 2026-09-10
Growth-media edges: docs and query match what the transforms emit (#539)

### LINKED ISSUE #539: Growth-media edges (METPO:2000517/2000518) encoded inconsistently across constants, queries, and docs
## Context

Surfaced in Copilot's 2026-04-16 review of #531.

`kg_microbe/transform_utils/constants.py` now models the growth-media edges ("grows in" / "does not grow in") as **Biolink predicates**, with `METPO:2000517` / `METPO:2000518` placed in the `relation` column. But two other parts of the codebase still treat those METPO IDs as the **predicate**:

1. `kg_microbe/query_utils/organism_queries.py::get_media_preferences()` filters on:
   ```python
   edges.predicate IN ('METPO:2000517', 'METPO:2000518')
   ```
   With the new encoding this query will silently return zero rows even when growth-media edges exist.

2. `.claude/skills/kg-query/SKILL.md` documents these IDs as predicates.

This is a self-inconsistency: the transform emits one shape; queries and docs expect another. Any downstream consumer that queries growth media will be silently wrong.

## Suggested fix

Pick one representation and make all three sites agree:

- **Option A** — keep Biolink predicates in `constants.py`. Update `get_media_preferences()` to filter on `e.relation IN ('METPO:2000517', 'METPO:2000518')` and use the relation value to split `grows_in` vs `no_growth`. Update SKILL.md accordingly.
- **Option B** — revert `constants.py` so the METPO IDs remain predicates, leave queries and docs unchanged.

## Files involved

- `kg_microbe/transform_utils/constants.py`
- `kg_microbe/query_utils/organism_queries.py` (`get_media_preferences`)
- `.claude/skills/kg-query/SKILL.md`

## References

- [PR #531](https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/531)
- Copilot review at commit `1de973d`, 2026-04-16T23:15Z

realmarcin:
Closes #539.

## Measured, not assumed

On the FRESH 2026-09-10 outputs, growth-media edges look like this in **both** sources:

| source | predicate | relation | edges |
|---|---|---|---:|
| bacdive | `METPO:2000517` | `METPO:2000517` | 36,596 |
| mediadive | `METPO:2000517` | `METPO:2000517` | 55,251 |
| mediadive | `METPO:2000518` | `METPO:2000518` | 3 |

No edge carries `biolink:located_in`. The docs (`kg-query/SKILL.md`, the `get_media_preferences` docstring) described a Biolink-predicate encoding that does not exist, and `tests/test_duckdb_query_cli.py`'s fixture encoded that fiction too — so the docs, the fixture and the graph disagreed while the query, filtering on `relation`, happened to work.

## Change

- `get_media_preferences`: match `predicate` first, fall back to `relation` (so a graph built before the METPO predicate still answers); classify by whichever carried the term. Docstring states the measured shape.
- `.claude/skills/kg-query/SKILL.md`: the encoding paragraph, and the node/edge counts (1.5M / 6.1M were from an old graph → 3.38M / 15.39M with a pointer to `merged_graph_stats.yaml`'s new `provenance` block).
- Fixture uses the real shape; new test covers grows-in, does-not-grow-in, and the legacy relation-only edge.

No rerun needed — query and docs only.

### DIFF
--- a/.claude/skills/kg-query/SKILL.md
+++ b/.claude/skills/kg-query/SKILL.md
@@ -31,13 +31,12 @@ poetry run kg query-organism "Corynebacterium glutamicum" -o report.md
 1. **Name Resolution**: Fuzzy matching on organism names and synonyms
 2. **Taxonomy**: Species classification hierarchy
 3. **Phenotypic Traits**: Oxygen preference, metabolism, morphology, Gram stain
-4. **Media Preferences**: Growth media edges use a Biolink predicate (e.g. `biolink:located_in`) and carry the METPO semantics in the `relation` column — `METPO:2000517` (grows in) and `METPO:2000518` (doesn't grow in). Filter on `relation` when querying.
+4. **Media Preferences**: Growth media edges carry the METPO term as the **predicate** — `METPO:2000517` (grows in) and `METPO:2000518` (does not grow in) — with the same term repeated in `relation`; there is no `biolink:located_in` on these edges (measured 2026-09-10: 36,596 BacDive + 55,251 MediaDive edges, all `predicate == relation`). Filter on `predicate`; `get_media_preferences()` also accepts the term in `relation` so a graph built before the METPO predicate still answers. See #539.
 5. **Media Composition**: 2-hop traversal (organism → media → solutions → chemicals)
 6. **Strain Information**: All strain records linked to species
 
 ## Data Sources
-- Nodes: 1.5M entries (883K organism taxa)
-- Edges: 6.1M relationships
+- Nodes: 3.38M entries; Edges: 15.39M relationships (canonical `merge.yaml`, 2026-09-10 — see `merged_graph_stats.yaml`, whose `provenance` block names the merge that produced it)
 - Primary sources: BacDive, MediaDive, MadinEtal, BactoTraits, GTDB, UniProt
 
 ## Example Organisms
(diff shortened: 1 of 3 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff.

- **The fixture was part of the bug.** `tests/test_duckdb_query_cli.py` wrote `biolink:located_in` + METPO relation — the shape the docs described, not the one the transforms emit — so the suite could never have caught the docs being wrong. It now writes the measured shape, and the same-size rebuild test (#866) is re-keyed on the relation digit (`2000517` → `2000518`, equal length) so its byte-for-byte premise still holds. The first push had that test red for one commit; fixed in `65a8d1844` before CI finished.
- **Fallback cannot double-count.** An edge with the METPO term in both columns matches the predicate branch of the `CASE` and is classified once; the legacy row (`biolink:located_in` + relation) is classified by relation. The test has one of each plus a no-growth edge and asserts exact lists.
- **`rg "located_in"` over `kg_microbe/query_utils/`** finds no other growth-media filter; the CLI (`kg query-organism`) goes through `get_media_preferences`.
- **Stale counts in the skill were a second, unrelated fiction** (1.5M / 6.1M); replaced with the 2026-09-10 numbers and a pointer to the stats file's provenance block rather than a number that will drift again.

No new issues.

PR #1027, opened 2026-09-10
tests: freshness timestamp-path tests no longer depend on the pipeline having run (#1001)

### LINKED ISSUE #1001: FreshnessDataStalenessTest reads the real data/transformed markers, so it fails on any checkout whose transforms have run

realmarcin:
Closes #1001.

`FreshnessDataStalenessTest` mocks `_latest_commit`, `_has_local_diff`, `_output_mtime` and `_latest_data_input_commit`, then calls `check_source`. But `check_source` consults `_fingerprint_verdict` first, which reads the real `data/transformed/<source>/source_fingerprint.json`. On this checkout (16 FRESH markers) two of the four tests failed; in CI, with no output, they passed — the suite's result depended on whether the developer had run the pipeline.

## Change

- `setUp` stubs `_fingerprint_verdict` to return `None` ("no usable marker") for the class, so the timestamp path is the one under test regardless of what is on disk. The content path already has its own tests (`FreshnessSchemaTest`, driven by synthetic markers).
- One new test pins the behaviour that caused this: a verifying marker makes `check_source` report FRESH whatever the timestamps say.

Verified: the class passes on this checkout with the real markers present (was 2 failed / 2 passed), 5 passed; the whole file 28 passed.

### DIFF
--- a/tests/test_transform_dispatch_and_data_inputs.py
+++ b/tests/test_transform_dispatch_and_data_inputs.py
@@ -132,10 +132,19 @@ class DataInputsTest(TestCase):
 
 
 class FreshnessDataStalenessTest(TestCase):
-    """The freshness check must report data staleness, not just code staleness."""
+    """
+    The freshness check must report data staleness, not just code staleness.
+
+    These tests drive the *timestamp* path. ``check_source`` consults the
+    content fingerprint first, and on a checkout where the transform has run
+    the real ``data/transformed/<source>/source_fingerprint.json`` decided the
+    verdict and the mocks below never mattered -- two tests failed locally and
+    passed in CI, which has no output (#1001). The fingerprint verdict is
+    stubbed to "no marker" here; the content path has its own tests below.
+    """
 
     def setUp(self):
-        """Import the standalone freshness script."""
+        """Import the standalone freshness script and take the marker out of play."""
         spec = importlib.util.spec_from_file_location(
             "kgm_freshness_check",
             REPO_ROOT / ".claude" / "skills" / "kgm-freshness-check" / "kgm_freshness_check.py",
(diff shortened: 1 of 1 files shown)


realmarcin:
## Adversarial review

Read-only pass.

- **The stub is scoped to the class and cannot leak.** `mock.patch.object` started in `setUp` with `addCleanup(stop)`, on the module object each test re-imports via `spec_from_file_location` — so `FreshnessSchemaTest`, which drives `_fingerprint_verdict` for real, is untouched (its 4 tests still pass in the same file).
- **The stub does not hide a regression in the wiring.** If `check_source` stopped consulting the fingerprint at all, the new `test_a_content_verdict_wins_over_stale_timestamps` fails (it expects FRESH from a verifying marker with stale timestamps); if it stopped falling back to timestamps, the existing three fail. Both directions are covered.
- **Verified on the failing checkout** (16 real markers present): 5 passed; the PR body's whole-file figure was wrong on first write (30) and is corrected to the measured 28.

No new issues.

PR #1029, opened 2026-09-10
lpsn_api: publish nodes.tsv/edges.tsv atomically (#985)

### LINKED ISSUE #985: LPSN's paired atomic writes can still publish a torn pair if the second os.replace fails

realmarcin:
Closes #985.

`lpsn_api.run()` wrote `nodes.tsv` and `edges.tsv` with plain `open()`; a crash mid-run left a truncated pair the merge reads as complete. This is the same change #981 made in `lpsn.py`: both files go through `atomic_write`, so an exception inside the body unwinds both and neither is published.

The two-file window the issue's last comment describes (inner `os.replace` succeeds, outer fails) remains, as it does for `lpsn` — the repo has no two-file transaction and the issue records that as not worth building for these two transforms.

Test mirrors `test_a_run_that_dies_midway_leaves_no_output_file` from the lpsn suite: a crash after the first record leaves no output file and no `*.partial`. 21 passed.

Rerun: batched with the end-of-batch `lpsn_api` run (13 min); the code path that writes rows is unchanged, so the output is expected byte-identical.

### DIFF
--- a/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
+++ b/kg_microbe/transform_utils/lpsn_api/lpsn_api.py
@@ -75,6 +75,7 @@ from kg_microbe.transform_utils.constants import (
 )
 from kg_microbe.transform_utils.lpsn.lpsn import LPSN_KNOWLEDGE_SOURCE, LPSN_PREFIX
 from kg_microbe.transform_utils.transform import Transform
+from kg_microbe.utils.atomic_io import atomic_write
 
 # JSON keys returned by the LPSN API (documented at
 # https://lpsn.dsmz.de/text/lpsn-api). Kept as module constants so a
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only. Same shape as #981 one file over: both outputs go through `atomic_write`, the crash test asserts no file and no `*.partial`, and the row-writing code is untouched (the rerun at the end of this batch should be byte-identical — that is the acceptance check). The narrower two-file window from the issue's last comment is unchanged and stays recorded there. No new issues.

PR #1031, opened 2026-09-10
gtdb: give every genome one correct, versioned, resolvable assembly identifier (#882)

### LINKED ISSUE #882: Each assembly is indexed under only one of its two accession prefixes (GCF or GCA), so half of all lookups miss
## The defect

An NCBI assembly has **two** accessions for the same physical assembly: a GenBank one (`GCA_000005845.2`) and a RefSeq one (`GCF_000005845.2`). They share a numeric core and are paired 1:1.

KG-Microbe mints a node under whichever one GTDB happened to pick, and never the other. Measured on `data/transformed/gtdb/nodes.tsv`:

```
GCF nodes                                  447,137
GCA nodes                                  454,204
numeric cores present under BOTH prefixes        0
```

Zero. Every one of the 901,341 genome nodes is reachable by exactly one of its two accessions.

**Consequence:** a consumer holding `GCF_027889305` finds the node; the same consumer holding `GCA_027889305` for the same assembly finds nothing. Which half of the graph answers depends on whether GTDB classified that genome from RefSeq (`RS_GCF_`) or GenBank (`GB_GCA_`) — an upstream detail no downstream user knows or should have to.

This is not a dangling-edge or missing-node problem, so nothing currently detects it: both nodes exist, both are typed `biolink:Genome`, and the graph looks complete.

## Two adjacent defects in the same identifiers

Found while confirming the above; they share a fix site (`gtdb.py:306`).

**1. RefSeq accessions carry a `GenBank:` prefix.**

```
GenBank:GCF_027889305    biolink:Genome
GenBank:GCA_947493065    biolink:Genome
```

`gtdb.py:306` is unconditional:

```python
genome_id = f"{GENBANK_PREFIX}{base_accession}"
```

`GCF_*` is a **RefSeq** accession. 447,137 nodes assert it is a GenBank one. The description string compounds it — `description=f"GenBank genome {accession}"` for every row.

Note also that `GenBank` does not appear in `transform_utils/custom_curies.yaml`, so the prefix is unregistered as well as wrong.

**2. The version suffix is dropped.**

`extract_accession_type()` splits `GCF_000005845.2` into `("GCF_000005845", "2")` and only the base reaches the node id. Measured: **0 of 901,341** ids retain a version.

`GCA_000005845.1` and `.2` are different assemblies of the same genome, so the identifier is now ambiguous, and it is not resolvable — no assembly resolver accepts a version-less accession. `metatraits_gtdb.py:191,229` independently does the same thing (`.split(".")[0]`), so the two transforms agree, which is why nothing has surfaced it.

## Why this matters for the merged graph

GTDB contributes 901,341 genome nodes — the largest node block after NCBITaxon. Anything joining external genome data (bakta output, IMG/GOLD ids, an assembly-keyed dataset) hits a ~50% miss rate that presents as "we have no data for that genome" rather than as an error.

## Possible fixes, in rough order of cost

1. **Emit `same_as` / `close_match` between the paired accessions.** GTDB's metadata files carry both — `bac120_metadata.tsv.gz` has the GenBank and RefSeq accession for each genome — so the pairing is available without a new download. Keeps one canonical node, makes the other resolvable.
2. **Register the correct prefix per accession** — `refseq:` for `GCF_*`, `genbank:` for `GCA_*` — and add both to `custom_curies.yaml`. Fixes defect 1 and makes the CURIEs meaningful.
3. **Keep the version suffix**, or state deliberately why it is dropped. If it is dropped for join stability against sources that lack it, that reasoning belongs in a comment; right now it reads as an oversight and is duplicated in two transforms.

(1) and (2) are independent and can land separately. (3) changes every genome node id, so it wants its own PR and a re-merge.

## Verification notes

All counts above are from `data/transformed/gtdb/nodes.tsv`, which `kgm-freshness-check` currently reports as `STALE_VS_CODE`. The staleness is against #875's formatting sweep, not a behaviour change (see #879), and the accession-minting code at `gtdb.py:295-312` is unchanged since that output was built — so the counts stand. Worth re-confirming after the next `kg transform -s gtdb`.

realmarcin:
Closes #882.

All three defects the issue names share one line (`gtdb.py:306`) and one blast radius — `GenBank:` ids appear **only** in gtdb's own output (901,341 nodes and edges; nothing else in the repo joins on them), so fixing them separately would mean two id-changing PRs and two re-merges for the same set of rows.

## Measured before the change (FRESH 2026-09-10 output)

| | |
|---|---:|
| RefSeq accessions asserted as `GenBank:` | 447,137 |
| ids retaining a version | 0 of 901,341 |
| numeric cores reachable under both accessions | **0** of 901,341 |
| GTDB rows whose paired accessions differ in version | 3,016 |

## Change

- **id**: `ncbi.assembly:<accession>`, version intact. One bioregistry-registered prefix covers both archives, resolves to the assembly datasets page, and its pattern accepts `GCA_*` and `GCF_*` — unlike `refseq`, whose pattern (`NC_`, `NP_`, `NZ_`…) does not match assembly accessions at all, so the obvious genbank:/refseq: split would have produced invalid CURIEs.
- **pairing**: GTDB ships `ncbi_genbank_assembly_accession` for every row (0 missing). A RefSeq-primary genome records its GenBank twin in the standard `same_as` node column. Rejected alternative, recorded in the docstring: declaring a second node per twin — one physical assembly is one thing, and it would double the largest node block after NCBITaxon to say nothing new.
- **name / description**: the NCBI accession, not GTDB's `RS_`/`GB_` key. `"RefSeq assembly RS_GCF_000005845.2"` named a string RefSeq does not issue.
- `ncbi.assembly` registered in `custom_curies.yaml` and `kg_model_review.STANDARD_PREFIXES`; `gtdb-phylo-diagram` prose updated (it filters on the predicate, not the prefix — no behaviour change).

## Canary — real `poetry run kg transform -s gtdb` (35 s)

| check | before | after |
|---|---:|---:|
| nodes / edges | 1,148,709 / 1,471,034 | **identical** |
| assembly nodes | 901,341 | 901,341 |
| …with a version | 0 | 901,341 |
| …GCF / GCA | 447,137 / 454,204 | unchanged split |
| …with `same_as` | — | 447,137 (exactly the RefSeq-primary ones) |
| `GenBank:` occurrences | 901,341 | **0** |
| cores addressable under both accessions | 0 | **447,137** |

Downstream check: `kg transform -s lpsn` (21 s) — it reads gtdb's `nodes.tsv` with `csv.DictReader` and skips every row whose id is not `GTDB:`, so the new column and ids are invisible to it. Output **byte-identical** (sorted md5 match), which is the acceptance test for the header change.

## Rerun

`gtdb` is done (above). `lpsn` → `lpsn_api` → `microbedecoder` need a fingerprint refresh only (~1 min total, verified no content change at the first hop); batched with the canonical merge at the end of this series.

### DIFF
--- a/.claude/skills/gtdb-phylo-diagram/SKILL.md
+++ b/.claude/skills/gtdb-phylo-diagram/SKILL.md
@@ -18,7 +18,7 @@ It is intentionally a *tree* view, not a network view — for network views use 
 Every edge in the merged KG is classified by looking at its two endpoints and predicate:
 
 - If **both** endpoints are taxon-prefixed (`NCBITaxon:`, `GTDB:`, `kgmicrobe.strain:`, `kgmicrobe.genus:`) the edge is *structural* — it carries the GTDB hierarchy, GTDB↔NCBITaxon `close_match` links, or strain→species `subclass_of` links. **Excluded** from node sizing.
-- If the predicate is `biolink:subclass_of` (regardless of the other endpoint) the edge is *classification* — most importantly the ~732K `GenBank:<genome> --subclass_of--> GTDB:<species>` edges (one per GTDB genome). **Excluded** from node sizing, otherwise the circles would measure "how many genomes GTDB has for this clade" rather than metadata richness.
+- If the predicate is `biolink:subclass_of` (regardless of the other endpoint) the edge is *classification* — most importantly the ~901K `ncbi.assembly:<accession> --subclass_of--> GTDB:<species>` edges (`GenBank:<genome>` before #882) (one per GTDB genome). **Excluded** from node sizing, otherwise the circles would measure "how many genomes GTDB has for this clade" rather than metadata richness.
 - Otherwise the edge contributes +1 to the count for whichever endpoint is a taxon. These are the biologically interesting edges: `has_phenotype`, `location_of` (isolation source), growth-media (`METPO:2000517`), and the METPO trait predicates.
 
 Counts on NCBITaxon are folded onto their GTDB equivalent via the mapping built in step 1; strain counts are folded onto the NCBITaxon parent and then onto GTDB. Cumulative counts propagate up the tree.
--- a/.claude/skills/gtdb-phylo-diagram/gtdb_phylo_diagram.py
+++ b/.claude/skills/gtdb-phylo-diagram/gtdb_phylo_diagram.py
@@ -189,7 +189,7 @@ def stream_edges(
                 continue
 
             # subclass_of with one taxon endpoint is a classification edge,
-            # not metadata -- e.g. GenBank:<genome> --subclass_of--> GTDB:<species>
+            # not metadata -- e.g. ncbi.assembly:<accession> --subclass_of--> GTDB:<species>
             # (~732K of these, one per GTDB genome). Excluding them keeps the
             # circle size meaning "biologically interesting edges" (phenotype,
             # isolation source, growth media, METPO traits) rather than
--- a/.claude/skills/kg-model-review/kg_model_review.py
+++ b/.claude/skills/kg-model-review/kg_model_review.py
@@ -275,6 +275,9 @@ STANDARD_PREFIXES = {
     "PO",        # Plant Ontology (OBO; reachable via ENVO/FOODON closure)
     "TAXRANK",   # Taxonomic Rank vocabulary (OBO; from NCBITaxon)
     "GenBank",   # GenBank sequence accessions (from NCBITaxon xrefs)
+    # NCBI Assembly accessions (GCA_*/GCF_*) minted by the gtdb transform.
+    # One prefix for both archives; the accession says which one (#882).
+    "ncbi.assembly",
     "chemrof",   # chemical role framework (CHEBI-adjacent)
     "debio",     # domain entity for biology (Rhea-adjacent)
     "kgmicrobe", # KG-Microbe native prefix (bare form; dotted variants also registered)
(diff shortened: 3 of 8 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus five probes.

- **The asymmetry in `same_as` is correct, not partial coverage.** 447,137 of 901,341 nodes carry a twin, and that is exactly the RefSeq-primary set. GTDB's metadata has `accession` and `ncbi_genbank_assembly_accession` but **no RefSeq assembly accession column** — because a `GB_GCA_*` genome is one RefSeq never mirrored, so it genuinely has only a GenBank accession. Nothing is missing for the other 454,204; there is no second identifier to record.
- **No merge collision on the new column.** gtdb is the only source with a non-empty `same_as` (13 other transforms: 0), and the merge cleanup already dedupes `same_as` as a known node column, so the added column cannot change any other source's row.
- **No new invented endpoints.** `same_as` is a node column, not an edge endpoint, and `find_stub_nodes` scans endpoints — so the twin accessions cannot appear in `merged_stub_nodes.tsv`. `EXPECTED_STUB_PREFIXES` never listed `GenBank` either, confirming nothing treated these as foreign cross-references.
- **The prefix choice is checked against the registry, not assumed.** `ncbi.assembly`'s bioregistry pattern matches both `GCF_000005845.2` and `GCA_000005845.2`; `refseq`'s pattern (`AC_|NC_|NP_|NZ_…`) matches **neither**, so the genbank:/refseq: split the issue suggests would have minted invalid CURIEs for 447,137 nodes. Worth recording on the issue as a correction to its option (2).
- **Two `GenBank:GC…` strings remain in the tree**, both in comments explaining the fix (`constants.py:230`, a test docstring). Intentional.
- **Not exercised:** the merged graph. `merged_stub_nodes.tsv`, the model reviewer's prefix check, and the `same_as` dedupe all get their first real test at the canonical merge at the end of this series; that is the acceptance check for the merge-side claims above.

No new issues.

PR #1032, opened 2026-09-10
gtdb: a shared NCBI taxon is a broad_match, and every one of them is reported (#883)

### LINKED ISSUE #883: Join taxa on GTDB, not NCBI: 334 NCBI placeholder taxa absorb 42.7% of the GTDB mapping and pool unrelated species
## The finding

NCBI carries coarse placeholder taxa — `bacterium`, `uncultured bacterium`, `Pseudomonadota bacterium` — that exist to park unclassified sequence. GTDB does not: it assigns every genome a named species. So the `close_match` bridge from GTDB to NCBI is many-to-one, and severely so at the top.

Measured on `data/transformed/gtdb/edges.tsv`:

```
GTDB -> NCBITaxon close_match edges     322,327
distinct GTDB subjects                  199,923
distinct NCBI objects                   118,521
```

The worst offenders, with names resolved from the trimmed NCBITaxon extract:

| GTDB taxa collapsed | NCBI taxon | name |
|---:|---|---|
| 3,695 | NCBITaxon:244328 | uncultured Clostridia bacterium |
| 3,492 | NCBITaxon:1869227 | **bacterium** |
| 3,450 | NCBITaxon:77133 | uncultured bacterium |
| 3,271 | NCBITaxon:1977087 | Pseudomonadota bacterium |
| 2,876 | NCBITaxon:179883 | uncultured Candidatus Saccharibacteria bacterium |

## Why this is worse than "NCBI is coarser"

The pathology is **extremely concentrated**, which is what makes it a trap rather than a known limitation:

```
NCBI taxa receiving exactly 1 GTDB taxon : 109,895  (92.7%)
NCBI taxa receiving >=100 GTDB taxa      :     334  ( 0.3%)
   ...which absorb                       : 137,497 of 322,327 links (42.7%)
```

**0.3% of NCBI taxa absorb 42.7% of the mapping.** So the join looks well-behaved on inspection — nine of ten NCBI taxa are clean 1:1 — while nearly half the edges route through a few hundred grab-bags. Spot-checking will not find this; only the aggregate does.

## Load-bearing consequence for downstream work

**When joining phenotypes or traits to organisms, key on the GTDB id, not `ncbi_taxon_id`.**

Joining through NCBI silently pools unrelated species. A trait asserted for one genome under `NCBITaxon:1869227` becomes, after a join, a trait shared with 3,491 other unrelated organisms. The result is a many-to-many blow-up that looks like signal — this is the mechanism behind the ~212k strain fan-out observed earlier, most of which routes through these joints.

It also corrupts any "top N" summary computed in NCBI space. A ranking of heterotroph-rich taxa showing `Pseudomonadota bacterium` (801 rows) and `Bacteroidota bacterium` (666) at the top is not a biological result — it is ~800 distinct GTDB species piled into one placeholder. In GTDB space those same rows are 800 separately named species, and the apparent pooling disappears.

## Corrected reading of the heterotroph assembly table

With GTDB as the key, the sampling is deliberate curation rather than anything random:

- one representative assembly per species, spanning **28.7%** of the ~143k known bacterial species
- **100%** of the mega-sequenced species (>10k assemblies each) covered by a single pick — presumably the reference genome
- **~78%** of the species it samples are singletons in KGM, i.e. the only known assembly for that species

## Suggested actions

1. **Document the join key.** A line in `CLAUDE.md` next to the existing taxon-convention section: cross-source organism joins use the GTDB identifier; `ncbi_taxon_id` is for display and for reaching NCBITaxon's hierarchy, not for grouping.
2. **Make the grab-bags visible in-graph.** These placeholder taxa are identifiable — `bacterium`, `uncultured *`, `* bacterium` with no species epithet. Flagging them (a node property, or a curated deny-list in the shape of `mappings/gold_ecosystem_mesh_sites.tsv`) lets a query exclude them deliberately rather than discovering the problem in a result set.
3. **Consider whether `close_match` is the right predicate** for a 3,695:1 relation. It reads as near-equivalence, and to a 3,695-member grab-bag it is not; `broad_match` would be closer to the truth for the concentrated tail. This is the same family-mismatch question as #783/#790/#823 — the predicate asserts more than the data supports.

## Verification note

Counts are from `data/transformed/gtdb/`, currently `STALE_VS_CODE`. The staleness is against #875's formatting sweep rather than a behaviour change (#879), and the mapping code is unchanged since that output was built — so the figures stand. Worth re-confirming after the next `kg transform -s gtdb`. Related: #882, on the same transform's accession handling.

realmarcin:
Closes #883. Implements all three suggested actions.

## Re-measured on the freshly rebuilt output

The issue asked for re-confirmation after the next `kg transform -s gtdb` (which #1031 just did). Every figure reproduces exactly:

| | issue | now |
|---|---:|---:|
| GTDB→NCBITaxon `close_match` edges | 322,327 | 322,327 |
| distinct GTDB subjects / NCBI objects | 199,923 / 118,521 | 199,923 / 118,521 |
| NCBI taxa receiving exactly 1 | 109,895 (92.7%) | 109,895 (92.7%) |
| NCBI taxa receiving ≥100 | 334 (0.3%) absorbing 137,497 (42.7%) | identical |

## 3 — the predicate

Biolink defines `close_match` as *"semantically similar but not strictly equivalent, **broader**, or narrower"* — so it is wrong the moment two GTDB taxa share an NCBI taxon, because that taxon *is* broader. 1:1 keeps `close_match`; many-to-one becomes `broad_match` / `skos:broadMatch` (the object is the broader term, per SKOS). **No threshold is invented** — the line is exactly "is this NCBI taxon shared", which needs no magic number and is defensible per-edge.

Because the fan-in is not knowable row by row, the mapping pairs are now held and emitted in one pass after the last metadata row. A test pins that: the first of two rows onto one taxon looks 1:1 until the second arrives.

## 2 — the grab-bags, in a file

The same pass writes `data/transformed/gtdb/gtdb_ncbi_pooling_report.tsv` (gitignored, worst-first, `ncbi_taxon / gtdb_taxa / predicate / examples`), on every run whether or not anything pooled. 8,626 rows on real data; top three are `NCBITaxon:244328` (3,695), `1869227` (3,492), `77133` (3,450) — the issue's table.

Chose a report over a node property or a curated deny-list: the set is derived, changes with every GTDB release, and a curated file would drift. It follows `merged_strain_parent_violations.tsv`'s convention.

## 1 — the join key

`CLAUDE.md` states it next to the taxon-convention section, with the concentration figure and a pointer to the report.

## Canary — real `kg transform -s gtdb` (41 s)

| check | before | after |
|---|---:|---:|
| nodes / edges | 1,148,709 / 1,471,034 | **identical** |
| GTDB→NCBI `close_match` | 322,327 | 109,895 |
| GTDB→NCBI `broad_match` | 0 | 212,432 |
| shared NCBI taxa reported | — | 8,626 |

A pure predicate reassignment: same edges, same endpoints, 322,327 either way.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,6 +98,17 @@ traversable; it is not a claim that the isolate is an ontology class. Biolink
 validators will report it; do not silently change the convention. See issue
 #834 and [the 4.4.2 revalidation](docs/BIOLINK_4_4_2_REVALIDATION.md).
 
+Cross-source organism joins use the **GTDB identifier**, not `ncbi_taxon_id`.
+NCBI carries coarse placeholder taxa (`bacterium`, `uncultured bacterium`,
+`Pseudomonadota bacterium`) that park unclassified sequence; GTDB names every
+genome, so the bridge is many-to-one and pathologically concentrated — 0.3% of
+NCBI taxa absorb 42.7% of the links, one of them 3,695 GTDB taxa. Joining
+through NCBI pools unrelated species and produces fan-out that looks like
+signal. A mapping edge onto a shared NCBI taxon is `biolink:broad_match`, not
+`close_match`, and every shared taxon is listed worst-first in
+`data/transformed/gtdb/gtdb_ncbi_pooling_report.tsv` — the deny-list to exclude
+them deliberately rather than discovering them in a result set. See #883.
+
 A `kgmicrobe.strain:<code>` node minted from a culture-collection deposit number
 is shared: several source records can cite the same deposit, and they do not
 always agree on the taxon. Such a node gets the one claimed parent that every
--- a/kg_microbe/transform_utils/constants.py
+++ b/kg_microbe/transform_utils/constants.py
@@ -445,6 +445,9 @@ ASSESSED_ACTIVITY_RELATIONSHIP = "NCIT:C153110"
 CLOSE_MATCH = "skos:closeMatch"
 CLOSE_MATCH_PREDICATE = "biolink:close_match"
 CLOSE_MATCH_RELATION = "skos:closeMatch"
+BROAD_MATCH = "skos:broadMatch"
+BROAD_MATCH_PREDICATE = "biolink:broad_match"
+BROAD_MATCH_RELATION = "skos:broadMatch"
 EXACT_MATCH = "skos:exactMatch"
 EXACT_MATCH_PREDICATE = "biolink:exact_match"
 ASSOCIATED_WITH = "PATO:0001668"
@@ -713,6 +716,9 @@ GTDB_RAW_DIR = RAW_DATA_DIR / GTDB
 GTDB_BAC120_TAXONOMY = "bac120_taxonomy.tsv"
 GTDB_AR53_TAXONOMY = "ar53_taxonomy.tsv"
 GTDB_BAC120_METADATA = "bac120_metadata.tsv.gz"
+#: Report of NCBI taxa that several GTDB taxa map onto (#883). Written on
+#: every gtdb run, empty or not.
+GTDB_NCBI_POOLING_REPORT = "gtdb_ncbi_pooling_report.tsv"
 GTDB_AR53_METADATA = "ar53_metadata.tsv.gz"
 
 # Metatraits-specific paths
--- a/kg_microbe/transform_utils/gtdb/gtdb.py
+++ b/kg_microbe/transform_utils/gtdb/gtdb.py
@@ -2,10 +2,13 @@
 
 import csv
 import gzip
+from collections import Counter
 from pathlib import Path
-from typing import List, Tuple
+from typing import Dict, List, Tuple
 
 from kg_microbe.transform_utils.constants import (
+    BROAD_MATCH_PREDICATE,
+    BROAD_MATCH_RELATION,
     CATEGORY_COLUMN,
     CLOSE_MATCH_PREDICATE,
     CLOSE_MATCH_RELATION,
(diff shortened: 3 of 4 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus five probes.

- **The symmetry argument is stronger than the PR body says.** Biolink marks `close_match` **`symmetric: True`** and `broad_match` not symmetric. So the old edge did not merely overstate similarity onto a 3,695-member grab-bag — it asserted it *in both directions*, i.e. that `NCBITaxon:1869227` ("bacterium") is a close match for each of 3,492 named species. `broad_match` is directional and points the right way (`skos:broadMatch`: the object is broader). Same family as the `close_match`-across-a-collapse rule already in CLAUDE.md for deposits.
- **Deterministic across runs.** Two consecutive `kg transform -s gtdb` runs give identical sorted-edge md5 (`7496553590…`) and identical report md5 (`9a9fcfde2d…`). The deferred emission iterates a list in insertion order and the report sorts on `(-count, ncbi_id)`, so neither depends on dict or set ordering — the `#1006` failure mode is not reintroduced by holding the pairs.
- **No other source is touched.** lpsn also emits 19,362 `biolink:close_match` edges at `GTDB:` objects, but those are LPSN→GTDB name matches in the other direction and go through a different code path; unchanged.
- **No validator regression.** `predicate_mapping.yaml` contains **zero** `skos:` entries — `skos:closeMatch` was never in it either — and nothing validates the `relation` column against it, so introducing `skos:broadMatch` cannot fail a check that `skos:closeMatch` was passing.
- **Deferred emission changes edge order in the file** (mapping edges now all land after the genome edges instead of interleaved). Row order in a KGX TSV is not meaningful and the merge sorts nothing on it; the sorted-md5 check above is the one that matters. Noted so it is not a surprise in a diff.
- **Not exercised:** the merged graph. The model reviewer has no `broad_match` entry in its house-allowance or domain/range tables — Biolink gives `broad_match` no domain or range, so there is nothing to violate, but the first canonical merge at the end of this series is the acceptance check.

No new issues.

PR #1034, opened 2026-09-10
ontologies: shape every UPA node row to the header written above it (#1033)

### LINKED ISSUE #1033: ontologies: the upa branch writes a 9-column header over 12-column rows, so kg transform -s ontologies dies on upa
## What happens

`poetry run kg transform -s upa` (or the full `-s ontologies`) fails, reproducibly, in 12 s:

```
[transform] upa: FAILED — ParserError: Error tokenizing data. C error: Expected 9 fields in line 6, saw 12
  File ".../ontologies_transform.py", line 969, in _remove_iri_column
    df = pd.read_csv(nodes_file, sep="\t", low_memory=False)
```

`upa` is 6th of 14 in `ONTOLOGIES_MAP`, so before #685 this aborted the whole ontologies transform and the eight ontologies behind it (mondo, hp, metpo, uberon, foodon, pato, ro, taxrank) never ran. With #685 it is isolated, and `gold`/`prego` are skipped because they declare `ontologies` in `TRANSFORM_INPUTS` — but the output directory is left holding **two generations** of ontology files, which must not be merged.

## The file it produces

`data/transformed/ontologies/upa_nodes.tsv`, 2,050 rows:

| shape | rows | what they are |
|---|---:|---|
| 9 fields (matches the header) | 964 | `RHEA:*` stub nodes |
| **12 fields** | **1,086** | `UPA:UCR*` reaction nodes, 3 trailing KGX columns (`…\t\t\thttp://purl.obolibrary.org/obo/UPa_UCR00014`) |

Pandas tolerates rows *shorter* than the header and raises on rows *longer*, which is why only the UPA rows kill it.

## Root cause

`ontologies_transform.py:764` writes the canonical 9-column header:

```python
with open(nodes_file, "w") as new_nf:
    new_nf.write("\t".join(self.node_header) + "\n")
```

but the rows come back from `unipathways_utils.replace_category_for_unipathways`, whose last act is:

```python
complete_parts = parts + ([""] * (len(node_header) - len(parts)))
```

That **pads but never truncates**. When KGX's row is wider than the canonical header — it leaks `subsets`, `meta` and `iri` for nodes, which is exactly what `_normalize_schema` exists to strip — the multiplier goes negative, `[""] * -3` evaluates to `[]`, and all 12 fields survive under a 9-field header. `replace_id_with_xref` builds its stub rows at exactly `len(node_header)`, which is why the RHEA half is well-formed.

Two more instances of the same pad-without-truncate pattern sit in `kg_microbe/utils/uniprot_utils.py:563` and `:572`; those transforms are `MISSING_OUTPUT` today, so the defect is latent there.

`replace_id_with_xref` also indexes the row it is building with `category_index`, an index derived from the *source* header, which is only safe while the two headers happen to agree in position.

## Fix

Project each row onto `self.node_header` **by column name** rather than padding positionally, so an extra, missing or reordered KGX column cannot shift a value into the wrong field. Positional truncation would work today only because the canonical header happens to be a prefix of KGX's column order.

## Acceptance test

`poetry run kg transform -s upa` exits 0; every row of `upa_nodes.tsv` has exactly 9 fields; the 1,086 `UPA:UCR*` nodes and 964 `RHEA:*` stubs are all still present; two runs give identical sorted-row md5s.

realmarcin:
Closes #1033. Found by the full rebuild: `kg transform -s ontologies` dies on `upa`, 6th of 14.

## The bug

`ontologies_transform.py:764` writes the canonical 9-column header; the rows come back from `replace_category_for_unipathways`, ending in

```python
complete_parts = parts + ([""] * (len(node_header) - len(parts)))
```

which **pads and never truncates**. KGX leaks `subsets`, `meta` and `iri` onto node rows — precisely the columns `_normalize_schema` exists to strip — so on a 12-field row the multiplier goes negative, `[""] * -3` is `[]`, and all 12 fields survive under a 9-field header:

| shape | rows | |
|---|---:|---|
| 9 fields | 963 | `RHEA:*` stubs — built at `len(node_header)`, always fine |
| **12 fields** | **1,086** | `UPA:UCR*` rows, trailing `…\t\t\thttp://purl.obolibrary.org/obo/UPa_UCR00014` |

Pandas tolerates short rows and raises on long ones, which is why only the UPA half killed `_remove_iri_column`.

## The fix

Project each row onto the canonical header **by column name** (`project_onto_header`), passing KGX's own header down from the caller. Positional truncation would work *today* only because the canonical header happens to be a prefix of KGX's column order; matching by name means an extra, missing or reordered upstream column cannot shift a value into the wrong field.

The same change fixes the RHEA stub builder, which was writing `l_parts[category_index]` — `category_index` is an index into the **source** header, correct only while the two headers agreed in position.

## Canary — real `kg transform -s upa`

| | before | after |
|---|---:|---:|
| exit | **1** (`ParserError`) | **0** (10.9 s) |
| malformed rows | 1,086 of 2,049 | **0** |
| UPA / RHEA nodes | 1,086 / 963 | 1,086 / 963 — unchanged |
| `purl.obolibrary.org` in the file | present | 0 |

Two consecutive runs give identical sorted-row md5s for nodes (`b0087c06…`) and edges (`8990047a…`).

7 new tests: wide row truncated, short row padded, columns matched by name under a reordered header, RHEA stub width + category by name, and the no-source-header fallback.

## Note

Two more instances of the same pad-without-truncate pattern live in `kg_microbe/utils/uniprot_utils.py:563,572`. Those transforms are `MISSING_OUTPUT`, so the defect is latent; left in #1033's description rather than changed blind here.

### DIFF
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -725,12 +725,19 @@ class OntologiesTransform(Transform):
                 add_lines = []
                 for line in nf:
                     if line.startswith("id"):
+                        # KGX's own header for this file. Kept so each row can be
+                        # projected onto self.node_header by column name: KGX
+                        # leaks subsets/meta/iri onto node rows, and padding
+                        # positionally left those extra fields under a canonical
+                        # 9-column header, producing a file pandas could not read
+                        # (#1033).
+                        source_header = line.strip().split("\t")
                         # get the index for the term 'id'
-                        id_index = line.strip().split("\t").index(ID_COLUMN)
+                        id_index = source_header.index(ID_COLUMN)
                         # get the index for the term 'xref'
-                        xref_index = line.strip().split("\t").index(XREF_COLUMN)
+                        xref_index = source_header.index(XREF_COLUMN)
                         # get the index for the term 'category'
-                        category_index = line.strip().split("\t").index(CATEGORY_COLUMN)
+                        category_index = source_header.index(CATEGORY_COLUMN)
                     else:
                         line = remove_unwanted_prefixes_from_node_xrefs(line, xref_index)
                         # For Reactions only
@@ -744,6 +751,7 @@ class OntologiesTransform(Transform):
                                 category_index,
                                 nodes_dictionary,
                                 self.node_header,
+                                source_header,
                             )
                             for new_line in new_lines:
                                 if len(new_line) > 0:
@@ -754,7 +762,7 @@ class OntologiesTransform(Transform):
                             substring in line for substring in [UNIPATHWAYS_PATHWAY_PREFIX]
                         ):  # ,UNIPATHWAYS_LINEAR_SUB_PATHWAY_PREFIX]):
                             new_line = replace_category_for_unipathways(
-                                line, id_index, category_index, self.node_header
+                                line, id_index, category_index, self.node_header, source_header
                             )
                             if len(new_line) > 0:
                                 add_lines.append(new_line + "\n")
--- a/kg_microbe/utils/unipathways_utils.py
+++ b/kg_microbe/utils/unipathways_utils.py
@@ -3,7 +3,9 @@
 import re
 
 from kg_microbe.transform_utils.constants import (
+    CATEGORY_COLUMN,
     GO_PREFIX,
+    ID_COLUMN,
     OBJECT_COLUMN,
     RHEA_CATEGORY,
     RHEA_NEW_PREFIX,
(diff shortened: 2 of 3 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus a field-level before/after comparison of the real output.

**The fix repairs a second, silent defect the issue did not mention.** Diffing non-empty field counts per node showed every UPA *pathway* node losing exactly one value, which looked like data loss. It is not — it is the `deprecated` column being emptied of something that never belonged there:

| | before | after |
|---|---:|---:|
| UPA rows with a synonym | 453 | 453 |
| …of those, `deprecated` holding an **exact copy** of the synonym list | **453 / 453** | 0 |
| rows with a non-empty `deprecated` | 453 | **0** of 2,049 |
| synonyms still present | — | 453, unchanged |

So positional padding had been writing the synonym list into `deprecated` for every UPA pathway node with synonyms — a column KGX consumers read as a flag. Projecting by name puts the value in `synonym` only. This is upa-specific: `pato`, `ec`, `chebi`, `go` and `envo` all have 0 non-empty `deprecated`.

It also confirms the fix had to match **by name**: had I truncated positionally (the obvious one-line fix), those 453 rows would still carry the synonym list in `deprecated`, because the source column at index 8 is not `deprecated`.

Other checks:
- **No node gained or lost.** `comm -3` over the id sets before/after is empty — 1,086 UPA + 963 RHEA either way.
- **Blast radius is the upa branch only.** `replace_category_for_unipathways` and `replace_id_with_xref` have exactly two call sites, both inside `if name == "upa"`; no other ontology reaches them.
- **Determinism**: two runs, identical sorted-row md5s for nodes and edges.
- **The `strict=False` on `zip` is load-bearing**, not lint appeasement: rows are legitimately shorter than the header (KGX omits trailing empties) and were legitimately longer (the bug). Both are handled by the name lookup; `strict=True` would raise on the very inputs this fixes.

No new issues — the `uniprot_utils.py:563,572` siblings are already recorded in #1033's description and those transforms are `MISSING_OUTPUT`.

Issue #845, opened 2026-08-20
Nothing detects staleness against another transform's output (gold depends on ontologies)


realmarcin:
Found while canarying the fingerprint work in the #797/#836 PR.

`DATA_INPUTS` covers **curation files under `mappings/`**. It does not cover a transform's dependency on *another transform's output*, and several exist:

- `gold` reads `data/transformed/ontologies/ncbitaxon_nodes.tsv` to apply the NCBITaxon trim, and refuses to run without it
- `gold` also reads `data/transformed/ontologies/envo_nodes.tsv` for the ENVO crosswalk availability check
- `microbedecoder` emits only against `lpsn:` nodes the `lpsn` transform supplies (#811)
- `ontologies_stubs` is ordered after `ontologies` so the SemSQL DBs exist

The new content fingerprint records `{"code": ..., "data": ...}` where `data` is the digest of `DATA_INPUTS`. gold declares none, so its `data` digest is the digest of the empty set — correct under the current contract, and silent about a real dependency.

**Consequence:** re-run `ontologies` and the NCBITaxon trim changes; `gold` is genuinely stale and every check reports it fresh. That is the #812 shape again, across transforms rather than within one.

`DATA_SOURCES` in `transform.py` already encodes the ordering in comments ("Run gold after ontologies…", "Run ontologies_stubs after ontologies…"), so the knowledge exists — it just isn't machine-readable.

## Fix

Declare upstream transforms the way `DATA_INPUTS` declares curation files — e.g. `TRANSFORM_INPUTS = ("ontologies",)` — and fold the upstream's recorded fingerprint into the downstream's. A transform is then stale when its own code changed, its curation data changed, **or** anything it reads was rebuilt differently.

Worth doing before the next full rebuild, since that is exactly when the ordering matters and when a wrong "fresh" is most expensive.

PR #1037, opened 2026-09-10
Declare the three ontologies dependencies the guard could not see (#1035)

### LINKED ISSUE #1035: Three transforms read ontologies output through a shared constant, so the cross-transform guard never sees the dependency
Found during the 2026-09-10 full rebuild, while working out why `mediadive` and `metatraits` were being run against a half-updated `data/transformed/ontologies/`.

## The gap

`Transform.TRANSFORM_INPUTS`' docstring says **five** cross-transform dependencies exist (gold, lpsn, lpsn_api, microbedecoder, prego). There are at least **eight**:

| transform | reads | declares |
|---|---|---|
| `bacdive` | `ontologies/ncbitaxon_nodes.tsv`, `ontologies/chebi_nodes.tsv` | `()` |
| `mediadive` | `ontologies/chebi_nodes.tsv` (ChEBI roles + categories) | `()` |
| `metatraits` | `ontologies/ncbitaxon_nodes.tsv` (`_load_ncbitaxon_labels`) | `()` |

`metatraits_gtdb` inherits metatraits' declaration, so it is affected too.

## Why the guard misses them

`tests/test_cross_transform_inputs.py` derives the real map by regexing three path idioms

```
transformed/<name>/ | output_base_dir / "<name>" | output_dir.parent / "<name>"
```

but only over `kg_microbe/transform_utils/<source>/**/*.py`. These three transforms never write such a path: they use constants defined in `constants.py`, which is outside the scanned tree.

```
CHEBI_NODES_FILE     = ONTOLOGIES_TRANSFORMED_DIR / "chebi_nodes.tsv"   → bacdive, mediadive
NCBITAXON_NODES_FILE = ONTOLOGIES_TRANSFORMED_DIR / "ncbitaxon_nodes.tsv" → bacdive, metatraits
```

`poetry run pytest tests/test_cross_transform_inputs.py` → **5 passed**. The test's own docstring says "Three prior versions of this same contract were opt-in and each was forgotten… This one is caught by the suite." It is not: expressing the dependency through a shared constant makes it invisible.

## Why it matters now

Two ways, both live:

1. **Freshness lies.** Re-running `ontologies` leaves bacdive, mediadive and metatraits genuinely stale while all three report `FRESH` — the #812 shape that #845 was meant to close.
2. **#685's skip logic cannot protect them.** When `ontologies` failed in this rebuild, `gold` and `prego` were correctly skipped because they declare it; `bacdive`, `mediadive` and `metatraits` ran anyway, against a directory holding five ontologies from the current run and eight from 2026-09-04. That is mixed-provenance output that looks clean.

## Fix

Resolve the constants when scanning: follow `*_NODES_FILE`-style names to their definition in `constants.py` and match `ONTOLOGIES_TRANSFORMED_DIR`-rooted paths, or invert the scan to start from the constants and find their users. Then declare the three dependencies.

Acceptance: the guard fails on today's code before the declarations are added, and `kgm-freshness-check` reports bacdive/mediadive/metatraits `STALE_VS_UPSTREAM` after an `ontologies` re-run.

Related: #812, #839, #845, #876 — the same contract, forgotten a fourth time.

realmarcin:
Closes #1035. Found during the full rebuild, working out why `mediadive` and `metatraits` were being run against a half-updated `data/transformed/ontologies/`.

## The hole

`tests/test_cross_transform_inputs.py` exists to make an undeclared cross-transform dependency impossible — its docstring says three prior versions of this contract were opt-in and each was forgotten, "This one is caught by the suite." It wasn't:

| transform | reads | declared |
|---|---|---|
| `bacdive` | `ontologies/ncbitaxon_nodes.tsv`, `ontologies/chebi_nodes.tsv` | `()` |
| `mediadive` | `ontologies/chebi_nodes.tsv` | `()` |
| `metatraits` | `ontologies/ncbitaxon_nodes.tsv` | `()` |

The scan regexes path idioms **only within `kg_microbe/transform_utils/<source>/`**. These three never write a path — they import `NCBITAXON_NODES_FILE` / `CHEBI_NODES_FILE`, defined in `constants.py`, outside the scanned tree.

## Fix

The scan resolves `constants.py` first — directory constants rooted at `TRANSFORMED_DATA_DIR` (via literal or via the source-name constant), then file constants built on those — and looks for the constant *names* in each transform. Then the three declarations.

**The new guard fails on the code as it stood before this commit**, which is precisely what the old one could not do:

```
FAILED CrossTransformDeclarationTest::test_every_observed_dependency_is_declared
  'ontologies' : bacdive reads ['ontologies'] but does not declare it in TRANSFORM_INPUTS
```

Two new pins so a refactor cannot silently empty the scan: the three constant-mediated dependencies are asserted present, and `_constants_naming_a_transform_output()` is asserted to resolve `NCBITAXON_NODES_FILE`/`CHEBI_NODES_FILE`/`ONTOLOGIES_TRANSFORMED_DIR` → `ontologies`.

## What this closes, both observed today

1. **Freshness lied.** Re-running `ontologies` left these three genuinely stale while every signal reported `FRESH` — the #812 shape #845 was meant to close.
2. **#685's skip logic could not protect them.** When `ontologies` failed mid-rebuild, `gold` and `prego` were correctly skipped; `bacdive`, `mediadive` and `metatraits` ran against a directory holding five ontologies from that run and eight from 2026-09-04.

Declarations go from five to nine (`metatraits_gtdb` inherits metatraits'). The `TRANSFORM_INPUTS` docstring, which asserted "five exist", is updated.

Rerun: these three are now correctly `STALE_VS_UPSTREAM` after the `ontologies` rerun that follows — which is the point.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -237,6 +237,11 @@ logger = logging.getLogger(__name__)
 class BacDiveTransform(Transform):
     """Template for how the transform class would be designed."""
 
+    #: Reads ``ontologies/ncbitaxon_nodes.tsv`` and ``ontologies/chebi_nodes.tsv``
+    #: via NCBITAXON_NODES_FILE / CHEBI_NODES_FILE. Undeclared until #1035: the
+    #: path lives in constants.py, so the cross-transform guard never saw it.
+    TRANSFORM_INPUTS = ("ontologies",)
+
     DATA_INPUTS = (
         "mappings/isolation_source_to_ontology.tsv",
         "mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz",
--- a/kg_microbe/transform_utils/mediadive/mediadive.py
+++ b/kg_microbe/transform_utils/mediadive/mediadive.py
@@ -136,6 +136,11 @@ LEGACY_HTTP_CACHE_FILENAME = "mediadive_cache.sqlite"
 class MediaDiveTransform(Transform):
     """Template for how the transform class would be designed."""
 
+    #: Reads ``ontologies/chebi_nodes.tsv`` (ChEBI roles and categories) via
+    #: CHEBI_NODES_FILE; the path lives in constants.py, which is why this went
+    #: undeclared (#1035).
+    TRANSFORM_INPUTS = ("ontologies",)
+
     DATA_INPUTS = ("mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz",)
 
     def __init__(self, input_dir: Optional[Path] = None, output_dir: Optional[Path] = None):
--- a/kg_microbe/transform_utils/metatraits/metatraits.py
+++ b/kg_microbe/transform_utils/metatraits/metatraits.py
@@ -253,6 +253,11 @@ def _process_file_worker(args: Tuple[Path, Path, Dict[str, Any], bool]) -> Dict[
 class MetaTraitsTransform(Transform):
     """Transform metatraits summary JSONL files into KGX nodes and edges."""
 
+    #: Reads ``ontologies/ncbitaxon_nodes.tsv`` via NCBITAXON_NODES_FILE in
+    #: _load_ncbitaxon_labels. metatraits_gtdb inherits this. Undeclared until
+    #: #1035, because the path lives in constants.py rather than here.
+    TRANSFORM_INPUTS = ("ontologies",)
+
     DATA_INPUTS = ("mappings/kgmicrobe_unified_entity_mappings.sssom.tsv.gz",)
 
     # Measurement traits that should be excluded from unmapped_traits.tsv
--- a/kg_microbe/transform_utils/transform.py
+++ b/kg_microbe/transform_utils/transform.py
@@ -58,11 +58,16 @@ class Transform:
     #: Registered source names whose **output** this transform reads.
     #:
     #: `DATA_INPUTS` covers curation files under ``mappings/``. It does not
-    #: cover a dependency on another transform's output, and five exist: gold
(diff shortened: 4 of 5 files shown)


realmarcin:
## Adversarial review

Read-only pass plus three probes.

- **The scan does not over-declare.** Running `_observed_dependencies()` over the tree yields exactly eight sources, each matching its declaration and nothing more: bacdive/mediadive/metatraits/prego → ontologies, gold → ontologies + ontologies_stubs, lpsn → gtdb, lpsn_api/microbedecoder → lpsn. Five constants resolve to a source; no constant resolves to something that isn't a registered transform.
- **The original three idioms still match.** `test_the_known_dependencies_are_present` (gold, lpsn, microbedecoder — one per idiom, including the `output_dir.parent` case an earlier scan missed) still passes, so the constants path is additive rather than a replacement.
- **`metatraits_gtdb` is covered by inheritance, not by the scan.** Its own directory never names the constant; it picks the declaration up from `MetaTraitsTransform`. Verified against the live registry: `metatraits_gtdb <- ('ontologies',)`. That is the same mechanism `tests/…::test_metatraits_gtdb_inherits_rather_than_redeclaring` already pins.
- **Honest limit on the acceptance test.** The issue's acceptance criterion — these three reporting `STALE_VS_UPSTREAM` after an `ontologies` re-run — is **not observable yet**: all three currently report `STALE_VS_CODE` because their own code changed in this PR, and code staleness takes precedence. It becomes checkable after the rebuild, when their code is settled and only the upstream moves. Stating it rather than claiming a verification I did not do.
- **Scope note:** the fix makes the dependency visible; it does not change what any transform reads. The only behavioural consequence is that these three now correctly go stale when ontologies is re-run — which is the point, and which is why they are in the rebuild that follows.

No new issues.

PR #1039, opened 2026-09-10
Stop stamping every merge -dirty: porcelain output must not be stripped (#1038)

### LINKED ISSUE #1038: merged_graph_stats provenance always says -dirty: _git() strips the leading space off git status --porcelain
Found on the first production run of the #1013/#993 stats hook (2026-09-10 merge). The block written was

```yaml
provenance:
  commit: acff35880-dirty
```

but the only tracked modification in the tree was `merged_graph_stats.yaml` itself — the one file `git_commit(repo_root, ignore=(stats_file,))` exists to exclude, because KGX rewrites it moments before the hook runs.

## Cause

`stats_provenance._git()` ends with

```python
return result.stdout.strip() if result.returncode == 0 else ""
```

`git status --porcelain` emits `XY PATH`, two status columns then a space, so an unstaged modification is `" M merged_graph_stats.yaml"`. `.strip()` removes the leading space, and `git_commit`'s parser then does `line[3:]` on the shifted string:

```
raw stdout       : ' M merged_graph_stats.yaml\n'
_git() returns   : 'M merged_graph_stats.yaml'
line[3:] on that : ['erged_graph_stats.yaml']
```

`'erged_graph_stats.yaml'` never matches the ignore set, so `dirty` is always True whenever anything is modified at all. The `ignore` parameter has never had any effect, and the unit test did not catch it because it asserts the commit is non-`unknown` and that ignoring "can only make the verdict cleaner" — both true of a permanently dirty answer.

## Consequences

Low blast radius, but it is precisely the class of defect #1013 was filed about: the provenance block makes a claim about the checkout that is not true. Every merge from now on would stamp `-dirty` regardless of the tree's real state, so the flag carries no information.

## Fix

Do not strip the status output before parsing — strip per-field, or parse the porcelain path as `line[2:].lstrip()` after `_git` returns it unmodified. `rev-parse` still wants stripping, so the strip belongs at the call site, not inside `_git`.

## Acceptance

With only `merged_graph_stats.yaml` modified, `git_commit(root, ignore=(stats_file,))` returns the bare short SHA and `git_commit(root)` returns `<sha>-dirty`. A test must construct that exact state — one modified tracked file that is also the ignored one — rather than asserting an inequality that a constant answer satisfies.

realmarcin:
Closes #1038. Found on the **first production run** of the #1013/#993 stats hook.

## What it wrote

```yaml
provenance:
  commit: acff35880-dirty
```

The only tracked modification in the tree was `merged_graph_stats.yaml` — the one file `git_commit(repo_root, ignore=(stats_file,))` exists to exclude, since KGX rewrites it moments before the hook runs.

## Cause

`_git()` ended with `result.stdout.strip()`. Porcelain is column-oriented (`XY PATH`), so an unstaged change is `" M merged_graph_stats.yaml"` and stripping the leading space shifts every column:

```
raw stdout       : ' M merged_graph_stats.yaml\n'
_git() returned  : 'M merged_graph_stats.yaml'
line[3:] on that : ['erged_graph_stats.yaml']
```

That never matches the ignore set. **`ignore` has never had any effect**, and `-dirty` was a constant — carrying no information about the checkout, which is the same class of false provenance claim #1013 was filed about.

## Fix

`_git()` returns stdout verbatim (`rev-parse` strips at its call site, where a bare value is wanted); `_porcelain_paths()` parses the column format and handles renames (`R  old -> new` → `new`) and quoted paths.

## The old tests passed against the bug

They asserted the commit was not `"unknown"` and that ignoring "can only make the verdict cleaner" — both true of a permanently dirty answer. The new tests construct the exact state: a real git checkout (`git init`, one commit) whose only modified tracked file is the ignored one.

**Falsified against the pre-fix code** — restoring `HEAD`'s `stats_provenance.py` and re-running:

```
FAILED test_the_ignored_stats_file_alone_does_not_make_the_tree_dirty
FAILED test_porcelain_paths_parses_the_column_format
2 failed, 1 passed
```

and 11 passed on the fix. A third test pins that ignoring the stats file does not blind the flag to a *second* modified file.

## Follow-through

The stats file produced by today's merge carries the wrong `-dirty`. Rather than re-run the 2½-hour merge, I re-run the annotation step alone against the same archive — it is idempotent by design and tested as such — so the committed block states the checkout truthfully.

### DIFF
--- a/kg_microbe/merge_utils/stats_provenance.py
+++ b/kg_microbe/merge_utils/stats_provenance.py
@@ -84,24 +84,56 @@ def git_commit(repo_root: Path, ignore: Iterable[Path] = ()) -> str:
     if git is None:
         return "unknown"
     try:
-        head = _git(git, repo_root, "rev-parse", "--short", "HEAD")
+        head = _git(git, repo_root, "rev-parse", "--short", "HEAD").strip()
+        # Not stripped: `git status --porcelain` emits "XY PATH", two status
+        # columns then a space, so an unstaged change reads " M path". Stripping
+        # the leading space shifted every column and the path parsed one
+        # character short, which made `ignore` a no-op and stamped -dirty on
+        # every merge (#1038).
         status = _git(git, repo_root, "status", "--porcelain", "--untracked-files=no")
     except (OSError, subprocess.SubprocessError):
         return "unknown"
     if not head:
         return "unknown"
     ignored = {_relative(Path(path), repo_root) for path in ignore}
-    changed = [line[3:] for line in status.splitlines() if line.strip()]
-    dirty = any(path not in ignored for path in changed)
+    dirty = any(path not in ignored for path in _porcelain_paths(status))
     return f"{head}-dirty" if dirty else head
 
 
 def _git(git: str, repo_root: Path, *args: str) -> str:
-    """Run one git command in the checkout and return stripped stdout."""
+    """
+    Run one git command in the checkout and return stdout verbatim.
+
+    Deliberately unstripped: porcelain output is column-oriented and a caller
+    that strips it before slicing loses a character off every path (#1038).
+    Callers wanting a bare value strip it themselves.
+    """
     result = subprocess.run(  # noqa: S603 - fixed argv, no shell
         [git, *args], cwd=repo_root, capture_output=True, text=True, check=False, timeout=30
     )
-    return result.stdout.strip() if result.returncode == 0 else ""
+    return result.stdout if result.returncode == 0 else ""
+
+
+def _porcelain_paths(status: str) -> list:
+    """
+    Return the paths named by ``git status --porcelain`` output.
+
+    Each line is ``XY PATH``: two status columns, a space, then the path. A
+    rename reads ``R  old -> new``; the destination is what changed, so that is
+    what is returned.
+
+    :param status: Raw stdout of ``git status --porcelain``, unmodified.
+    :return: One path per changed file.
+    """
+    paths = []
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff.

- **`_git` has exactly two callers** and the strip moved to the one that wants it. `rev-parse --short HEAD` now strips at the call site; `status --porcelain` is parsed unstripped. No third caller exists, so nothing else silently changes behaviour.
- **`len(line) < 4` guard is right, not off-by-one.** The shortest real porcelain line is `XY P` — two columns, a space, a one-character path — which is exactly 4. A blank trailing line (which the previous `if line.strip()` was there to skip) is length 0 and still skipped.
- **Renames were broken before too.** `R  old -> new` would previously have yielded `"old -> new"` as one path, never matching an ignore entry; now it yields the destination. Not the reported bug, but the same parser, so fixed and tested together.
- **The fix cannot mask a genuinely dirty tree** — pinned by `test_a_second_modified_file_still_reports_dirty`, which modifies `other.txt` alongside the ignored stats file and asserts `-dirty` survives.
- **The test creates a real repo rather than mocking `git`.** Slower (three subprocess calls) but it is the only way to exercise the actual porcelain bytes, which is where the defect lived — a mocked string would have encoded my own misreading of the format.
- **Honest limit:** the third new test asserts `_porcelain_paths` on a synthetic status string including `??`. Untracked entries never reach it in production (`--untracked-files=no`), so that case is documentation of the parser's contract rather than a live path.

No new issues.

PR #1021, opened 2026-09-10
merged_graph_stats.yaml from the 2026-09-10 canonical merge

realmarcin:
Stats from `poetry run kg merge -y merge.yaml` on master `4fcea53b1` (35 min), the first graph carrying #1017 (#1008) and #1019 (#1004). Every registered source with output was `FRESH` under fingerprint scheme 3 before the merge.

The committed baseline is the #1012 merge (2026-09-08, **before** #1010), so this diff also carries #1010's UBERON/FOODON/PATO inclusion: 328 keys added (category pairs such as `AnatomicalEntity-related_to-*`) and 20 removed (the `…-unknown` pairs the undeclared endpoints produced). Against the post-#1010 stats (measured, never committed) the key set is identical and only counts move.

| | committed (#1012) | post-#1010 (uncommitted) | this merge |
|---|---|---|---|
| nodes | 3,384,379 | 3,384,379 | 3,384,379 |
| edges | 15,408,449 | 15,408,449 | 15,392,517 |
| invented endpoints (`merged_stub_nodes.tsv`) | 818 | 635 / 7 prefixes | 635 / 7 prefixes, none `lpsn` |
| `merged_strain_parent_violations.tsv` | empty | empty | empty |

Node total unchanged because every endpoint the two changes touched was already declared (NCBITaxon by the ontology, the 665 basonyms as stubs since #1003 — now named; `lpsn:4387 Bacterium sonnei` is in the archive). Edge total down 15,932: `metatraits_gtdb` itself grew (+31,656), but edges that moved onto species-level NCBITaxon ids now dedupe against identical edges from other sources.

Same convention as #1012: the file records the graph on disk; nothing else in the tree changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1012, opened 2026-09-09
Refresh merged_graph_stats.yaml from the 2026-09-08 canonical merge (#689)

realmarcin:
Closes #689.

`merged_graph_stats.yaml` regenerated by `kg merge -y merge.yaml` on master `e7bb7cc3b` — the first canonical merge whose `metatraits_gtdb` input is reproducible (#1007): **3,330,088 nodes / 15,299,158 edges**, all 15 sources FRESH, merge FRESH.

- `grep -c biolink:in_taxon` → **0** (#689's acceptance test; #680 replaced the predicate).
- `biolink:has_attribute` 45,208 and `biolink:capable_of` 300,594 reflect #999/#1003.
- The previous committed copy came from the 2026-08-02 merge (#647).

Known limitation, unchanged by this file: the stats cannot name any METPO-predicate edge (#993).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SjJ3SqbfGvwsiezu4TNS4E

PR #1040, opened 2026-09-10
merged_graph_stats.yaml from the 2026-09-10 full rebuild, with its first provenance block

realmarcin:
The graph behind this file is the first built entirely from **one generation of code**: all 14 registered sources re-run after `constants.py` moved (4h04m, no failures, no skips), `kgm-freshness-check` reporting **14 FRESH / 0 stale**, then one canonical `merge.yaml`.

| | previous (2026-09-10 01:24) | this |
|---|---:|---:|
| nodes | 3,384,379 | 3,384,379 |
| edges | 15,392,517 | 15,285,903 (−106,614) |
| invented endpoints | 635 / 7 prefixes | 635 / 7 prefixes |
| `merged_strain_parent_violations.tsv` | empty | empty |

Neither `lpsn` nor the new `ncbi.assembly` prefix appears among the invented endpoints — #882 declares all 901,341 genome nodes.

## What is new in the file

**`provenance`** — when, commit, merge config, edges file, python, kgx, and each of the 24 merge sources' `source_fingerprint.json` digests. A stale copy can now be told from a current one without diffing counts against a graph you already trust, which is the failure #1013 was filed for.

**`edge_stats.count_by_raw_predicate`** — 113 predicates summing to **exactly `total_edges`**. KGX's own `count_by_predicates` still records **0 METPO entries**, so it cannot see the 67 METPO predicates carrying **7,656,831 edges**, including `METPO:2000511`'s 706,765 (#909). Two thirds of the graph was invisible in the artifact the README points readers at.

Also newly visible: **`biolink:broad_match`, 212,432 edges**, from #883's rule that a GTDB→NCBI mapping onto a shared NCBI taxon is not a symmetric `close_match`.

## Disclosure about the commit field

The archive was merged under `acff35880`. The `provenance` block was written by re-running the **annotation step alone** under `4f7cd3f9b`, because the merge-time run stamped a false `-dirty` (#1038, fixed in #1039 — which touches nothing that builds the graph). Re-annotating is idempotent by design and tested as such, and it counted the same archive: its raw-predicate total matches KGX's `total_edges`. The alternative was a 2½-hour re-merge to change one string.

Same convention as #1012 and #1021: the file records the graph on disk; nothing else in the tree changes.

### DIFF
--- a/merged_graph_stats.yaml
+++ b/merged_graph_stats.yaml
@@ -1,5 +1,21 @@
 edge_stats:
   count_by_predicates:
+    biolink:actively_involved_in:
+      count: 7
+      provided_by:
+        unknown:
+          count: 7
+      source:
+        unknown:
+          count: 7
+    biolink:affects:
+      count: 174
+      provided_by:
+        unknown:
+          count: 174
+      source:
+        unknown:
+          count: 174
     biolink:associated_with_resistance_to:
       count: 10585
       provided_by:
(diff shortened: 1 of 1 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus five probes against the archive itself, not the stats file's own claims.

- **The diff is stats-only** — one file, no other tracked change.
- **The archive matches the claimed counts**: 3,384,380 node lines (3,384,379 + header).
- **The three headline changes are actually in the graph**, checked by reading it: `FOODON:03301710 → biolink:Food` (#1015); `ncbi.assembly:GCF_027889305.1` with `same_as=ncbi.assembly:GCA_027889305.1` (#882); **212,432** `biolink:broad_match` edges (#883), matching `count_by_raw_predicate` exactly.
- **`same_as` survives the merge for exactly the right rows**: 447,137 of 901,341 assembly nodes, all of them GCF-primary — the RefSeq-primary set, which is the correct asymmetry (a `GB_GCA_*` genome is one RefSeq never mirrored, so it has no twin to record).
- **Found a pre-existing defect while doing this, filed as #1041.** My first count said "901,341 of 901,341 have a `same_as`", which was wrong: the merged graph is **CRLF on every one of its 3,384,380 node and 15,285,904 edge lines**, so an "empty" last field is actually `"\r"` and a truthiness test on it inverts. The real `same_as` values carry the CR too (`…GCA_027889305.1\r`). Not a regression — the 2026-09-07 archive is CRLF as well — and it does not affect this stats file, which is YAML written by `yaml.safe_dump`. But it means open issue #541 ("duckdb_loader silently corrupts the last column on CRLF input") is not hypothetical: the shipped graph triggers it.

**Correction to my own earlier claim in this thread:** I reported "901,341 with `same_as`" before checking what was in the field. The correct number is 447,137, and finding the discrepancy is what surfaced #1041.

No blockers for this PR — the stats accurately describe the archive that exists.

Issue #541, opened 2026-04-17
duckdb_loader.py silently corrupts the last column on CRLF input files


turbomam:
## Context

Surfaced in Copilot's 2026-04-16 review of #531.

`kg_microbe/query_utils/duckdb_loader.py` sets `lineterminator="\n"` on the pandas reader and strips `\r` from the header line — but not from the rest of the file. On a file with CRLF (`\r\n`) line endings this leaves a trailing `\r` in the last field of every data row, because `\r` is no longer treated as part of the newline.

Effects:

- string values are silently altered (`"foo"` becomes `"foo\r"`)
- equality filtering and indexing break in ways that are hard to spot — most viewers render the `\r` invisibly
- downstream joins on those columns silently drop rows

## Suggested fix

Either:

- drop the custom `lineterminator` and let pandas handle CRLF normally, or
- normalize the file contents on load (strip `\r` from every line, not only the header).

Option 1 is the simpler fix unless there's a concrete reason the custom terminator was introduced.

## File involved

- `kg_microbe/query_utils/duckdb_loader.py`

## References

- [PR #531](https://github.com/Knowledge-Graph-Hub/kg-microbe/pull/531)
- Copilot review at commit `1de973d`, 2026-04-16T23:15Z

PR #1042, opened 2026-09-10
Write LF, not CRLF: csv.writer's default terminator reached the shipped graph (#1041)

### LINKED ISSUE #1041: The shipped merged graph is CRLF, so every last-column value carries a trailing carriage return — and #541 says that corrupts the DuckDB loader
Found while verifying #882's `same_as` pairing reached the 2026-09-10 merged graph. It did, but every value is `…\r`.

## Measured on `data/merged/merged-kg.tar.gz` (2026-09-10 rebuild)

```
merged-kg_nodes.tsv : 3,384,380 of 3,384,380 lines end with CR
merged-kg_edges.tsv : 15,285,904 of 15,285,904 lines end with CR
```

Every line. Splitting on `\t`, the **last column of every row** carries a trailing carriage return:

```
$ ... | awk -F'\t' '$1=="ncbi.assembly:GCF_027889305.1"{print $9}' | od -c
   n c b i . a s s e m b l y : G C A _ 0 2 7 8 8 9 3 0 5 . 1  \r  \n
```

For nodes the last column is `same_as`; for edges it is `agent_type`. An empty last field is not empty either — it is `"\r"`, so a truthiness test on it is inverted. That is how this was found: a count of "assembly nodes with a `same_as`" returned 901,341 instead of the correct 447,137, because the 454,204 empties each held a CR.

## Not a regression

The previous `merged-kg-no-prego.tar.gz` (2026-09-07) is CRLF too, so the shipped graph has been this way for as long as the archives on disk go back.

## Cause, in two independent places

1. **Transforms.** `gtdb`, `lpsn`, `lpsn_api` and `gold` write via `csv.writer` / `csv.DictWriter` on a file opened **without `newline=""`** (e.g. `gtdb.py:480`). Python's csv default is `lineterminator="\r\n"`. `bacdive`, `mediadive` and `metatraits` write LF.
2. **The merge normalizers.** `merge_kg.py:331` and `:360` build `csv.writer(tmp, delimiter="\t")` the same way, so the normalization pass re-emits CRLF for *every* row regardless of which source it came from — which is why the merged file is 100% CRLF while only half the transforms are.

## Why it matters

**#541 is open and is exactly this failure**: "duckdb_loader.py silently corrupts the last column on CRLF input files." That issue reads as a hypothetical about badly-formed inputs. It is not — the graph this repo ships triggers it, so `kg query-organism` and the `kg-query` skill load a graph whose last column is corrupt on every row.

`tests/test_duckdb_query_cli.py::test_direct_loader_handles_crlf_and_records_fingerprints` covers the loader tolerating CRLF, which is the mitigation; nothing asserts the producer emits LF.

## Fix

`newline=""` at every `csv.writer`/`DictWriter` call site — the six in the transforms and the two in `merge_kg.py`. Cheap in code, but correcting the shipped artifact needs a re-merge (~2.5 h at current memory pressure), so it is worth batching with the next rebuild rather than run on its own.

## Acceptance

`tar -xzOf data/merged/merged-kg.tar.gz merged-kg_nodes.tsv | awk '/\r$/{c++} END{print c+0}'` is 0, likewise for edges; a test asserts each transform's output has no CR line endings, so a writer added later cannot reintroduce it.

realmarcin:
Closes #1041. Found while verifying #882's `same_as` pairing survived the merge — it did, as `…GCA_027889305.1\r`.

## What was wrong

Every line of `merged-kg.tar.gz` ended with CR — **3,384,380 node lines and 15,285,904 edge lines**. Splitting a row on tabs leaves the last column carrying it as data (`same_as` for nodes, `agent_type` for edges). An "empty" last field is `"\r"`, so a truthiness test inverts: a count of gtdb nodes with a `same_as` read **901,341** instead of the correct **447,137**.

## Corrected cause

My first analysis on the issue blamed a missing `newline=""` — wrong, and the evidence contradicted it: `gold`, `lpsn` and `lpsn_api` all pass `newline=""` and still emitted CRLF.

```
open(w) + csv default             -> b'a\tb\r\n'
open(w, newline="") + csv default -> b'a\tb\r\n'     <-- newline="" changes nothing
open(w, newline="") + LF          -> b'a\tb\n'
```

`csv.writer`'s default `lineterminator` is `"\r\n"` on every platform and is written literally; `newline=""` only stops the text layer translating it.

## Fix

`kg_microbe/utils/tsv_io.py` owns the decision (`tsv_writer` / `tsv_dict_writer`); the six graph-writing modules go through it.

Four transforms emitted CRLF (`gtdb`, `lpsn`, `lpsn_api`, `gold`) — but the merged file was **100% CRLF regardless of source**, because `merge_kg`'s two normalizers rebuild every row through a default writer, converting the LF majority on the way. Those two call sites are most of the fix.

## Guard

An AST test fails on any of the six modules constructing `csv.writer`/`DictWriter` without an explicit `lineterminator`, **and** pins the premise (`csv.writer` emits `\r\n`) so it cannot become vacuous if the stdlib ever changes. Falsified: it fails on the pre-fix `merge_kg.py`.

## Canaries

| | rows | CR lines |
|---|---:|---:|
| `kg transform -s gtdb` (34 s) — nodes | 1,148,710 | **0** |
| — edges | 1,471,035 | **0** |
| — pooling report | 8,627 | **0** |
| `kg transform -s lpsn` — nodes / edges | 86,667 / 156,205 | **0** |

Row counts unchanged, and the `same_as` count now reads 447,137 directly with no special handling. 124 tests pass across the touched suites.

## Not a regression, and its consequence

The 2026-09-07 archive is CRLF too. But it means open issue **#541** ("duckdb_loader.py silently corrupts the last column on CRLF input files") describes the graph this repo actually ships, not a hypothetical input — so `kg query-organism` and the `kg-query` skill have been loading a corrupt last column.

Correcting the merged artifact needs a re-merge; batching that with the next rebuild rather than spending 2½ hours to change line endings alone.

### DIFF
--- a/kg_microbe/merge_utils/invariants.py
+++ b/kg_microbe/merge_utils/invariants.py
@@ -26,6 +26,7 @@ from kg_microbe.transform_utils.constants import (
     SUBJECT_COLUMN,
 )
 from kg_microbe.utils.atomic_io import atomic_write
+from kg_microbe.utils.tsv_io import tsv_writer
 
 logger = logging.getLogger(__name__)
 
@@ -124,7 +125,7 @@ def check_merged_invariants(
     violations = find_multi_parent_strains(edges_file)
     destination = Path(output_dir or edges_file.parent) / STRAIN_PARENT_REPORT
     with atomic_write(destination, newline="") as handle:
-        writer = csv.writer(handle, delimiter="\t")
+        writer = tsv_writer(handle)
         writer.writerow(STRAIN_PARENT_REPORT_HEADER)
         writer.writerows(strain_parent_rows(violations))
     # Absent is normal: the strain-parent check needs no nodes file, and callers
@@ -139,7 +140,7 @@ def check_merged_invariants(
         stubs = find_stub_nodes(nodes_file)
         stub_destination = Path(output_dir or edges_file.parent) / STUB_NODE_REPORT
         with atomic_write(stub_destination, newline="") as handle:
-            writer = csv.writer(handle, delimiter="\t")
+            writer = tsv_writer(handle)
             writer.writerow(STUB_NODE_REPORT_HEADER)
             writer.writerows(stub_node_rows(stubs))
         unexpected = {p: c for p, c in stubs.items() if p not in EXPECTED_STUB_PREFIXES}
--- a/kg_microbe/merge_utils/merge_kg.py
+++ b/kg_microbe/merge_utils/merge_kg.py
@@ -31,6 +31,7 @@ from kg_microbe.transform_utils.constants import (
     SYNONYM_COLUMN,
     XREF_COLUMN,
 )
+from kg_microbe.utils.tsv_io import tsv_writer
 
 CANONICAL_NODE_HEADER = [
     ID_COLUMN,
@@ -328,7 +329,7 @@ def _normalize_nodes_tsv(path: Path) -> None:
             header, CANONICAL_NODE_HEADER, NODE_COLUMNS_TO_DROP, extension_columns=set()
         )
         _log_schema_diff("nodes", path, header, out_header)
-        writer = csv.writer(tmp, delimiter="\t")
+        writer = tsv_writer(tmp)
         writer.writerow(out_header)
         for row in reader:
             writer.writerow(_project_row(row, keep_indices))
@@ -357,7 +358,7 @@ def _normalize_edges_tsv(path: Path) -> None:
             extension_columns=EDGE_EXTENSION_COLUMNS,
         )
         _log_schema_diff("edges", path, header, out_header)
-        writer = csv.writer(tmp, delimiter="\t")
+        writer = tsv_writer(tmp)
         writer.writerow(out_header)
         for row in reader:
             if ks_idx is not None and pks_idx is not None and pks_idx < len(row):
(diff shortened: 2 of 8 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff plus four probes.

- **No behaviour change beyond the terminator.** The rewrite is mechanical (`csv.writer(` → `tsv_writer(`, dropping the now-redundant `delimiter="\t"`), and both canaries produced identical row counts — gtdb 1,148,709/1,471,034, lpsn 86,666/156,204 — so no row was added, lost or reshaped.
- **The scope is exactly the six modules that write parseable TSVs.** Other `csv.writer` call sites exist (scripts/, `metatraits/io.py`, report writers in `microbedecoder`, `prego`, `bacdive`) and are deliberately left: they either already set a terminator or write files nothing joins on. Widening the guard to them is a judgement call I did not make unilaterally; the list is explicit in the test so adding one is a one-line change.
- **`setdefault`, not a hard override.** A caller that genuinely needs a different delimiter or terminator can still pass one; the helper only supplies the default that was wrong.
- **The guard cannot pass vacuously.** `test_the_helper_writes_lf_where_a_bare_writer_writes_crlf` asserts the stdlib still emits `\r\n`, so if that ever changed the test fails loudly rather than silently guarding nothing.
- **Honest limit:** the merged graph is still CRLF on disk. Only the producers are fixed; the artifact is corrected by the next merge, and #541's loader-side mitigation stays worth having regardless, since third-party TSVs can still arrive CRLF.

No new issues.

PR #1043, opened 2026-09-10
bacdive: publish nodes.tsv/edges.tsv atomically (#1036)

### LINKED ISSUE #1036: bacdive writes nodes.tsv/edges.tsv non-atomically, so an interrupted run leaves a truncated pair that reads as complete
Same defect as #820 (lpsn) and #985 (lpsn_api), one transform over. Demonstrated accidentally on 2026-09-10: a full rebuild was stopped with SIGTERM while bacdive was running, and it left

```
data/transformed/bacdive/nodes.tsv    71,005,945 bytes
data/transformed/bacdive/edges.tsv   101,895,861 bytes
```

against ~633 MB for a complete run — about 27%, with no `.partial` marker and nothing else to distinguish it from a finished build. The merge reads whatever pair is on disk.

`bacdive.py` already imports `atomic_write` and uses it for `bacdive_media_links.txt` (line 3591), but the two files that matter go through plain writes.

What saved this instance is unrelated to bacdive: `_record_fingerprint` runs *after* the outputs in `transform.py`, so the interrupted run wrote no marker and the stale one reports `STALE_VS_CODE`. That is luck about ordering, not a property of the output — a rerun that is interrupted after a previous successful run of the *same* code would leave a truncated pair under a marker that verifies.

Fix: publish both files through `atomic_write`, as lpsn does since #981 and lpsn_api since #1029. The test those PRs used transfers directly — kill the run partway, assert no output file and no `*.partial`.

Worth checking the other transforms for the same shape while in there; `lpsn`, `lpsn_api` and `gtdb` are already atomic, the rest have not been audited.

realmarcin:
Closes #1036.

## The failure, observed

A SIGTERM during the 2026-09-10 rebuild left

```
data/transformed/bacdive/nodes.tsv    71,005,945 bytes
data/transformed/bacdive/edges.tsv   101,895,861 bytes
```

against ~633 MB for a complete run — 27%, with no `.partial` marker and nothing else distinguishing it from a finished build. The merge reads whatever pair is on disk.

## Where the gap actually was

The final `drop_duplicates` rewrite is **already atomic** (`pandas_utils.py` documents it), so my first reading — "bacdive writes non-atomically" — was only half right. The gap is the big `csv.writer` block that runs long before it, opening both graph outputs with a truncating `open()`. That is where the SIGTERM landed.

Both now go through `atomic_write`, the rule lpsn took in #820 and lpsn_api in #985.

The stub pass that appends to `nodes.tsv` is unchanged and still correct — it runs after the context manager closes, so it never sees a partial file. A test pins that, so a later reader does not mistake it for the same defect.

## Also: bacdive joins the CRLF guard (#1041)

Its seven `csv.writer` call sites go through `tsv_writer`. The graph outputs already reached disk as LF, but only because pandas rewrote them at the end; the **intermediates did not**, and `bacdive.tsv` is read by the mediadive transform.

## Guards

- AST test: fails if `self.output_node_file` / `self.output_edge_file` is opened for a truncating write.
- Positive assertion that both go through `atomic_write`, so deleting the writes entirely would not pass vacuously.
- Both **fail on the pre-fix file** (2 failed / 1 passed).

`pytest -k bacdive`: 102 passed, 1 skipped. Full-tree `ruff check` clean.

No rerun needed for the PR — this changes how the file is published, not its contents; bacdive's output is byte-identical modulo the intermediates' line endings, and it is rebuilt with the next batch anyway.

### DIFF
--- a/kg_microbe/transform_utils/bacdive/bacdive.py
+++ b/kg_microbe/transform_utils/bacdive/bacdive.py
@@ -222,6 +222,7 @@ from kg_microbe.utils.ontology_utils import (
 # Note: get_label and search_by_label are imported lazily in fallback methods
 from kg_microbe.utils.pandas_utils import drop_duplicates
 from kg_microbe.utils.string_coding import remove_nextlines
+from kg_microbe.utils.tsv_io import tsv_writer
 
 # Anitibiotic resistance
 if Path(METABOLITE_MAPPING_FILE).is_file():
@@ -1006,7 +1007,7 @@ class BacDiveTransform(Transform):
         if not missing:
             return 0
         with open(self.output_node_file, "a", encoding="utf-8", newline="") as handle:
-            writer = csv.writer(handle, delimiter="\t")
+            writer = tsv_writer(handle)
             for curie in missing:
                 label = self._get_ncbitaxon_label(curie)
                 writer.writerow(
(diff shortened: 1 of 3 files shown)


realmarcin:
## Adversarial review

Read-only pass over the diff.

- **Corrected my own issue text.** #1036 said "bacdive writes nodes.tsv/edges.tsv non-atomically", which was half wrong: the *final* `drop_duplicates` rewrite has been atomic all along. The truncation came from the earlier `csv.writer` block. Fixing the wrong stage would have left the real gap open, so the distinction mattered.
- **`atomic_write` cannot break the append that follows.** It renames on context exit, so by the time `_emit_stubs_for_unresolvable_taxa` opens the file in `"a"` mode the published file exists. Order verified in `run()`: write block (1837) → stub append → `drop_duplicates` (3632).
- **Failure semantics improve in the right direction.** Before: crash → truncated file that reads as complete. After: crash → no file, which is the state `microbedecoder` and the freshness check already understand (an absent output is unambiguous; a short one is not).
- **The seven writer conversions are mechanical** and covered by `pytest -k bacdive` (102 passed). `bacdive.tsv`'s consumers read it with pandas, which tolerated CRLF, so this is hygiene rather than a live bug — stated as such rather than claimed as a fix.
- **Honest limit:** I did not run a full bacdive transform to prove the atomic path end-to-end — it is a ~30-minute run and the change is a context-manager swap with AST coverage on both sides. It gets exercised for real in the next rebuild; if `atomic_write` mishandled the 633 MB write, that run would fail loudly rather than silently, since the outputs would be absent rather than wrong.

No new issues.

PR #1044, opened 2026-09-10
ontologies: an ontology's annotation vocabulary is not one of its classes (#1023)

### LINKED ISSUE #1023: ontologies: ec_nodes.tsv emits dcterms:license or dct:license for the same IRI depending on the run
Found by the #690 canary (2026-09-10): two runs of the EC ontology transform with identical code and input produced `ec_edges.tsv` byte-identical and `ec_nodes.tsv` differing in exactly one row:

```
< dcterms:license	biolink:OntologyClass				ec.json
> dct:license	biolink:OntologyClass				ec.json
```

Same IRI (`http://purl.org/dc/terms/license`), two CURIE prefixes, chosen differently between runs — the prefix map the converter uses has both `dcterms` and `dct` for that namespace and the winner depends on iteration order. Two consequences:

1. The transform is not deterministic (the #1006 protocol — sorted-row md5 over two runs — catches it; the fingerprint does not).
2. A license annotation property is being emitted as a `biolink:OntologyClass` node at all; it is not a class. Likely present in other `<ontology>_nodes.tsv` files too — check with `grep -c "^dc\(terms\)\?:" data/transformed/ontologies/*_nodes.tsv`.

Cheapest fix: drop annotation-property IRIs outside the ontology's own namespace from the node output (and any `dct:`/`dcterms:` prefix from the converter's ambiguity), then two-run md5 check on `-s ec` (11 min each with the new single-ontology flag from #690).

realmarcin:
Closes #1023.

## Scope, measured

`rdfs:label` was a node in the knowledge graph, typed `biolink:OntologyClass`, connected to nothing. So were `owl:deprecated`, `dc:title`, `dcterms:license` and 51 others:

| | |
|---|---:|
| distinct annotation-property ids as nodes | **55** |
| rows across the ten ontology outputs | 172 |
| **edges touching any of them** | **0** |
| reaching the shipped merged graph | 36 |

They are the vocabulary an OWL file annotates *with*, which KGX's OBO-JSON loader emits a node for whenever it encounters one.

## Fix

`_drop_metamodel_nodes` mirrors the existing `_drop_metamodel_edges` — a pure DataFrame filter, called from `_normalize_schema` where the leaked `subsets`/`meta`/`iri` columns are already stripped, keyed on the id prefix.

**This subsumes the defect the issue was opened for.** `dcterms:license` vs `dct:license` alternated between runs because the converter's prefix map carries both for one namespace and the winner depended on iteration order. Dropping the row removes the nondeterminism without needing to pick a winner.

Deliberately narrow: it matches the `id` column of a **nodes** file only. `skos:closeMatch` as a *relation value* is legitimate, lives in a different column of a different file, and is untouched — a test pins that distinction.

## Canary — `kg transform -s pato` (12 s)

| | before | after |
|---|---:|---:|
| annotation-property nodes | 14 | **0** |
| rows | 7,978 | 7,964 (−14 exactly) |
| real `PATO:` terms | 1,887 | 1,887 |
| edges | 18,910 | 18,910 |

An `ec` run (the issue's original example) is in flight; result posted as a comment.

## One documented sharp edge

A bare token equal to a metamodel prefix (`owl` with no colon) is treated as one. No ontology mints such an id, and the alternative — requiring a colon — would let a malformed row through. Pinned by a test that says so explicitly rather than left as a surprise.

### DIFF
--- a/kg_microbe/transform_utils/ontologies/ontologies_transform.py
+++ b/kg_microbe/transform_utils/ontologies/ontologies_transform.py
@@ -118,6 +118,18 @@ ONTOLOGIES_MAP = {
 # RDF/OWL predicate CURIEs — not biolink entity relationships — so they clutter
 # the merged KG (kgxval flags them as non-biolink) without carrying queryable
 # entity data. Dropped from the edge output; nodes are left untouched.
+#: Prefixes of the annotation and metadata vocabulary an OWL file annotates
+#: *with* -- never anything it defines. KGX's OBO-JSON loader emits a node for
+#: each one it encounters, so `rdfs:label`, `owl:deprecated`, `dc:title` and
+#: `dcterms:license` arrived as biolink:OntologyClass nodes connected to
+#: nothing: 55 distinct ids, 172 rows across ten ontologies, 0 edges, 36 of
+#: them in the shipped graph (#1023). Matching on the id column of a *nodes*
+#: file only -- `skos:closeMatch` as a relation value is legitimate and lives
+#: in a different column of a different file.
+METAMODEL_NODE_PREFIXES = frozenset(
+    {"dc", "dct", "dcterms", "terms", "doap", "foaf", "oio", "owl", "pav", "rdf", "rdfs", "skos"}
+)
+
 METAMODEL_EDGE_PREDICATES = frozenset(
     {
         "rdfs:subPropertyOf",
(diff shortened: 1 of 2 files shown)


realmarcin:
## Adversarial review

Read-only pass plus the `ec` canary the issue was opened about.

**`kg transform -s ec` (11 min):** `[_normalize_schema] ec_nodes.tsv: dropped 5 annotation-property node(s)`, leaving 11,866 rows with all 11,861 real `EC:` terms intact and **no `license` row of either spelling**. The `dcterms:license` / `dct:license` alternation that opened #1023 cannot recur, because the row no longer exists — which is a stronger resolution than picking a winning prefix.

- **Mirrors the edge-side sibling exactly** — same signature shape, same `(df, count)` return, same "no column, no opinion" guard, called from the same normalizer. A reader who knows `_drop_metamodel_edges` knows this.
- **The narrowness is load-bearing and tested.** `skos:closeMatch` is a legitimate `relation` value on real edges; the filter only ever sees the `id` column of a nodes file. If it were applied to the relation column it would silently delete real edges — so a test asserts `skos:closeMatch` is dropped as an *id* while nothing touches relations.
- **0 edges were touching these nodes**, so removing them cannot dangle an endpoint or create a KGX-invented stub at merge time. That was checked before the fix, not assumed.
- **The bare-token edge case is documented rather than hidden.** `owl` with no colon matches. No ontology mints such an id; requiring a colon would let a malformed row through instead. The test states the choice.
- **Honest limit:** verified on `pato` (14 dropped) and `ec` (5 dropped). The other eight ontologies use the identical code path but were not individually run — the full set gets exercised in the next rebuild, where the expected total is the 172 rows measured.

No new issues.

realmarcin:
Correction to the review above: I wrote "all 11,861 real `EC:` terms intact", which I had not measured. The correct figures are **7,251 `EC:` terms** of 11,866 total rows — the remainder are imported terms from other ontologies, which the filter also leaves alone. 11,866 is the row count, not the EC-term count. The substantive claims stand: 5 annotation-property nodes dropped, 0 remaining, and no `license` row of either spelling.

